PR #10, opened 2014-06-04
MAHOUT-1572 blockify() to detect (naively) the data sparsity in the loaded data

dlyubimov:
naive initial  fix (?)


mariusmuja:
For me, this caused ```spark-rowsimilarity``` to always stop with an ```java.lang.OutOfMemoryError: Java heap space```. Reverting this allowed spark-rowsimilarity to successfully complete.

dlyubimov:
yes this is not the way to do it 
actually it seems vectors are not reporting density faithfully, i had to do a more thorough analysis of data density elsewhere before doing things. this probably needs to be patched with a more thorough technique. 
+1

PR #44, opened 2014-08-13
jBLAS - initial implementation

avati:
This is a prototype for jBLAS integration for in-Core operations in Mahout

### DIFF
--- a/jblas/pom.xml
+++ b/jblas/pom.xml
@@ -1,0 +1,154 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+ Licensed to the Apache Software Foundation (ASF) under one or more
+ contributor license agreements.  See the NOTICE file distributed with
+ this work for additional information regarding copyright ownership.
+ The ASF licenses this file to You under the Apache License, Version 2.0
+ (the "License"); you may not use this file except in compliance with
+ the License.  You may obtain a copy of the License at
+
+     http://www.apache.org/licenses/LICENSE-2.0
+
+ Unless required by applicable law or agreed to in writing, software
+ distributed under the License is distributed on an "AS IS" BASIS,
+ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ See the License for the specific language governing permissions and
+ limitations under the License.
+-->
+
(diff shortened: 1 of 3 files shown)


dlyubimov:
My idea on this is to simply build a quick polinomial regressoin model (which is possible to calibrate on each platform individually via command line if desired) which will just predict the total turn-around time of an operation (like gemm) using geometry and non-default elements as well as source format (to account to expected conversion cost). This is a great simplification of things compared to what @tdunning is saying but given that the main purpose it automatic adaptation of bolt-on third party hardware accelerators (rather than optimizing default java implementation) i think it might be ok in pracitce. Don't have practical benchmarks though yet.

Then every other back (mkl, jblas or stuff like magma) would just be pluggable into this calibration thing.

We will provide default set of calibrated models for a few typical platforms with redistribution.

I've made some initial api changes to that effect, can't publish it right now though. 

Note that some hardware optimized libraries also provide benefits not just for dense but also sparse algebra.

But that's the direction i've been thinking on.

dlyubimov:
@avati, 

Based on my work for optimizations in #135, i am starting to think perhaps we need to approach this issue from operation-centric point of view rather than Matrix-centric point of view. 

The reason is that i am thinking that majority of Matrix api will perform very poorly anyway -- penalty for serialing data to and from native storages is too great for bulk but small iterators.

Even for bulk operators like %*% there is something called "truncation point" -- i.e. minimum size of the matrix when operation starts making sense in comparison to default one because of higher non-recurring costs. This is even true for Strassen algorithm in native code (let alone the one that would involve additional memory trasfers).

Instead, maybe, we could integrate implementations of an operator (see MMul class in #135, for example, which is being scala-side in-core %*% operator there) and focus on implementation selection based on calibrated truncation point and availability of native implementation. 

Also, in more complicated cases like GPU, or even perhaps in case of libblas.so cases, we may track cached matrix representations native format -- so we perform data transfers to and from lazily, e.g. if matrix i an intermediate result then it never needs to even be loaded into main memory -- that's exactly how BidMat does it.

The more i think about it, the more operation-centric integration makes sense to me (and the less the one Matrix-centric does).

andrewmusselman:
Is this PR still relevant @dlyubimov and @avati or should we close it?

avati:
closing PR for now

PR #57, opened 2014-10-07
MAHOUT-1619 - HighDFWordsPruner overwrites cache files

bwebster:
Fix for MAHOUT-1619.


smarthi:
This patch is failing unit tests when applied against present master. Could u verify/update ur patch? Thanks.

bwebster:
Thanks for updating the patch and pulling this in.

smarthi:
@bwebster Np, thanks again for the patch and pardon the delay in committing the same to master.

PR #60, opened 2014-10-19
Minor change to EuclideanDistanceSimilarity.java file

deepcoder007:
This is just a minor change to the file . The end product is same for both the cases.


smarthi:
Is there related project JIRA for this?

smarthi:
This doesn't seem correct, there is a correction that needs to be done in the code comments which is to be addressed by MAHOUT-1602. Will close this PR along with that Jira.

PR #62, opened 2014-11-15
MAHOUT-1626 Support for required quasi-algebraic operations and starting with aggregating rows/blocks

gcapan:
As discussed in the dev-list, we are now in the progress of adding an additional set of distributed operations to be performed on checkpointed matrices.

An initial implementation for aggregating rows/blocks is included in this request.


dlyubimov:
what's the status?

Also, what happens to H20 support of this? it'll just be failing?

h2o/src/main/scala/org/apache/mahout/h2obindings/H2OEngine.scala
```diff
@@ -32,6 +32,7 @@ import org.apache.mahout.common.{Hadoop1HDFSUtil, HDFSUtil}
 object H2OEngine extends DistributedEngine {
   // By default, use Hadoop 1 utils
   var hdfsUtils: HDFSUtil = Hadoop1HDFSUtil
+  val operations: DistributedOperations = null
```
dlyubimov: I guess in case of H20 this will break with NPE... Perhaps we can make it an Option[...] and gracefully report capability is not supported by engine if it happens to be None.

gcapan: You're right, I'll fix it.

gcapan:
The status is that I need to revise the code based on reviews.

But I have some concerns, summarized below:

Here is the story.

I'm going to contribute my recent work on distributed implementation of stochastic optimization to some open source library, and for me, the only reason that accumulating blocks matters is that I require it for averaging-based distributed stochastic gradient descent (DSGD).

I was an advocate of having Mahout as the ML and Matrix Computations core for distributed processing engines, and was thinking that the Matrix DSL would be sufficient for implementing such algorithms (such as DSGD) in an engine-agnostic way. 

It seems that for implementing most optimization algorithms and ML models, one requires other-than-DSL operations. And those operations are highly engine-specific.

Repeating the aggregating operation in Mahout is duplicate work, just like MLlib's having some of Mahout's Matrix DSL capabilities duplicated in uglier ways. Plus, having an algorithm in Mahout but not in MLlib (or vice versa) really bothers me because other's users could not benefit.

Considering your recent codebase refactoring effort, @dlyubimov, I imagine the best way to use the DSL is by utilizing it inside MLlib (or whatever your favorite ML library is). That is, MLlib depends on Mahout Matrix-DSL implementation, Matrix I/O and computations are handled in Mahout, ML algorithms are handled in MLlib and/or other libraries.

Can we just slow this down and think about what should be contributed to where, and reconsider the ideal Mahout-Spark integration?

dlyubimov:
What's your method on DSGD? i know there's DSGD paper but it deals with SGD specifically for matrix factorizations, not a general SGD computational scheme. Could you please adduce a reference.

In general, computing batch (non-stochastic) gradient is demonstrably algebraic task -- but that's likely not what you are doing at all here of course.

gcapan:
* The first and simplest will be Zinkevich et al.'s Parallelized Stochastic Gradient Descent [1]:  The algorithm is basically running multiple local SGD's in parallel, then averaging them. For implementation, I was thinking of running SGD's locally in blocks of rows and averaging them.

* Further, I hope to implement distributed stratified SGD for matrix factorization (Gemulla et al.)[2]: The algorithm is forming strata (where each stratum consists of a set of blocks that do not share any rows or columns), then for each stratum, performing SGD updates in parallel.

I am not yet sure if the latter would require additional non-DSL stuff. I will raise my concerns once I get to it.

[1] http://martin.zinkevich.org/publications/nips2010.pdf
[2] http://dl.acm.org/citation.cfm?id=2020426

avati:
Is this still being worked on? I can at least help with the H2O implementation of aggregate(), if there is interest in this PR..

gcapan:
Under some conditions, which are satisfied in the case of linear and logistic regression, a statistical optimization problem (the parameter estimation) over i.i.d. data, distributed as:

1. average of the local estimates
2. a combination of the average of the local estimates and the average of the estimates on the subsamples of the local sample sets

converges in mean to the optimal risk minimizer, as it is described in [1]. Given that, these methods are not only a way to distribute machine learning, they also provide a _justification for machine learning on Big Data_ (that is, these algorithms converge to the true risk minimizer as the whole data were processed on a single computer).

With this motivation, I propose to add the two distributing schemes for machine learning: averaging and bootstrap-averaging. These would be abstracted away from the actual loss minimization algorithms, and the backend engines would only provide these two simple functions. The users can throw their favourite (in-core) optimization algorithm, and of course we would want to provide some of them out-of-box.

Very soon, I am hoping to submit a patch for that. The current patch would be obsolete then, so there is no need to replicate this. Once I submit it, I'll close the current PR.

[1] http://arxiv.org/abs/1209.4129 
(The short version in NIPS: http://stanford.edu/~jduchi/projects/ZhangDuWa12_nips.pdf)

dlyubimov:
let me route for that.

dlyubimov:
ok i introduced allreduceBlock() in #135 but that, like spark's reduce, ends up in the driver collection, and supports only tensor types which is what i think we really should be constraining ourselves with (otherwise it would create with serialization requirements to the engines). Everything non-tensor and non-key i guess will have to be native engine code. 

now, very speculatively, allreduceBlock (think spark's map+reduce on matrix blocks) should cover what you are talking about.

gcapan:
It seems it does, let me try.

Thanks, Dmitriy

Sent from my iPhone

On Jun 8, 2015, at 23:31, Dmitriy Lyubimov <notifications@github.com> wrote:

ok i introduced allreduceBlock() in #135
<https://github.com/apache/mahout/pull/135> but that, like spark's reduce,
ends up in the driver collection, and supports only tensor types which is
what i think we really should be constraining ourselves with (otherwise it
would create with serialization requirements to the engines). Everything
non-tensor and non-key i guess will have to be native engine code.

now, very speculatively, allreduceBlock (think spark's map+reduce on matrix
blocks) should cover what you are talking about.

—
Reply to this email directly or view it on GitHub
<https://github.com/apache/mahout/pull/62#issuecomment-110128740>.

andrewmusselman:
I see this is old; where did we land with this @gcapan and @dlyubimov?

dlyubimov:
i think this became stale. the api part has been superceded by allreduce() operatora and the method part still is  unfinished up till now so unless @gcapan wants to finish this as he said he would try, i'd close it

sscdotopen:
I guess its fine closing this after a year of inactivity. @gcapan can reopen it if he continues the work.

gcapan:
Yeah, the required primitives are now included. It is fine to close this.

PR #68, opened 2014-12-28
H2O dependency upgraded

mmalohlava:
- h2o-core upgraded to version 0.1.17
    - API changes
    - more stable infrastructure
  - code using internal H2O API was refactored


andrewpalumbo:
thanks again @mmalohlava!

PR #69, opened 2014-12-29
MAHOUT-1636

pferrel:
Started out simplifying driver code and making changes to all drivers to support that. Then ran into the fat job.jar issue of MAHOUT-1636 so created a slimmed down version of the old job.jar by adding excludes to job.xml and changing the name to "dependencies.jar"

The new jar works for spark-itemsimilarity and spark-row-similarity but needs to be tested for the naive bayes drivers. 

The dependencies.jar still contains a lot of stuff from mrlegacy, some is in external projects, like jackson that can be excluded with this mechanism but also a lot of mahout code that is unneeded in this jar. This later case would require some other mechanism than a simple <exclude> clause in the assembly xml file.

I believe the new dependencies.jar is the only thing that needs to be on the classpath when running spark drivers or the spark-shell. I haven't changed this but it is a further refinement we can try.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/drivers/MahoutDriver.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/drivers/MahoutDriver.scala
@@ -25,7 +25,7 @@ abstract class MahoutDriver {
 
 
   implicit protected var mc: DistributedContext = _
-  protected var parser: MahoutOptionParser = _
+  implicit protected var parser: MahoutOptionParser = _
 
   var _useExistingContext: Boolean = false // used in the test suite to reuse one context per suite
 
--- a/spark/pom.xml
+++ b/spark/pom.xml
@@ -157,8 +157,8 @@
         </executions>
       </plugin>
 
-      <!-- create job jar to include CLI driver deps-->
-      <!-- leave this in even though there are no hadoop mapreduce jobs in this module -->
+      <!-- create an all dependencies job.jar -->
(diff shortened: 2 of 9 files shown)


pferrel:
Good.

Actually I wish you would comment on how we should structure the build. I'll post an email to the dev list, this one has become a TL;DR candidate.

andrewpalumbo:
as well, #68 should fix the broken h2o build that many people are experiencing.  Do you mind if I push that today- it's only one line change in the root pom and the rest is in h2o.

pferrel:
Go for it

dlyubimov:
What is the status? last update 25 days ago, was any action taken?

pferrel:
nothing on the assembly, I'll wait to see your refactoring.

Did lots of comment and scaladoc cleanup to take care of a fair bit of untidiness. Didn't remove scopt, which is just too small to be worth the effort. 

Replaced use of o.a.m.Pair with Scala tuples There may be some unneeded imports since I can't trust my IDE to clean those up. I'll have to take a longer look to remove any that are unnecessary. 

Still testing against the latest master but hope to push soon.

pferrel:
Wrong branch this PR will be dropped or suspended until @dlyubimov  refactoring is in. Another PR will have Pair removal and cleanup

pferrel:
Made a first cut of including a "dependency-reduced" jar with only Guava and Scopt included with <include> directive, see PR#76 https://github.com/apache/mahout/pull/76

PR #70, opened 2015-01-10
Non-negative Matrix Factorization and Probabilistic Matrix Factorization

3110101637:
Non-negative Matrix Factorization, using classical multiplicative update rule.
Probabilistic Matrix Factorization, using stochastic gradient descent

Use Movielens dataset  to test, get reasonable result.


math/src/main/java/org/apache/mahout/math/mf/PMF_SGD.java
```diff
@@ -0,0 +1,226 @@
+package org.apache.mahout.math.mf;
+
+import org.apache.mahout.math.DenseMatrix;
+import org.apache.mahout.math.Matrix;
+import org.apache.mahout.math.function.Functions;
+
+<<<<<<< HEAD
```
dlyubimov: merge artifact

dlyubimov:
so this is in-core stuff mostly?

I think Scala based contributions are more preferable at this point... 

needs cleanup a bit although looking quite neat

dlyubimov:
Also needs a jira # to discuss.

andrewmusselman:
@ghost is this still in flight?

agnisys:
How to start work on it..

PR #71, opened 2015-01-22
Upgrade spark to 1.1.1, make ready for 1.2.0 in test

hamnis:


spark/src/test/scala/org/apache/mahout/sparkbindings/test/DistributedSparkSuite.scala
```diff
@@ -39,6 +39,7 @@ trait DistributedSparkSuite extends DistributedMahoutSuite with LoggerConfigurat
           .set("spark.kryoserializer.buffer.mb", "15")
           .set("spark.akka.frameSize", "30")
           .set("spark.default.parallelism", "10")
+          .set("spark.driver.allowMultipleContexts", "true")
```
dlyubimov: Why is this property important for tests? they won't run otherwise?

hamnis: for 1.2.0 they wont run, so this was to future proof upgrade to 1.2.0. Maybe could have bumped to 1.2.0 directly?

dlyubimov: @hamnis, could you please explain what is significance of this parameter -- "allowMultipleContexts" in 1.2.0 so that tests won't run?

As far as i remember, current head is using only one context (per suite).so next suite should run another context, but by that time the previous one should already be gone.

hamnis: This was done since the tests failed with an error message similar to:
"Please enable this property to make tests work".

Currently using my phone, so can't check exact error message.

To reproduce: set spark version to 1.2.0 and run tests.

dlyubimov: Hm... I think i want to understand this more.

I suspect the right thing to do is to understand why how our test setup is
flawed for 1.2  and fix context instantiation rather than put this property
in.

Could we please delay this until we move on to 1.2 at which point.

Otherwise i suppose i don't see a reason not to bump to 1.1.1 at this point.

pferrel: This would seem to imply some tests are nesting contexts. If so it's a Mahout bug. 

Spark 1.2 is probably the next thing to target since it's been out for awhile and is in the distros. It also has the fix for missing anon functions in the artifacts so it should remove the need to build Spark from source.

PR #73, opened 2015-02-05
Sparse-vector implementation based on fastutil (MAHOUT-1640)

vigna:
The collections currently used by Mahout to implement sparse vectors are extremely slow. The proposed patch (localized to RandomAccessSparseVector) uses fastutil's maps and the speed improvements in vector benchmarks are very significant. It would be interesting to see whether these improvements percolate to high-level classes using sparse vectors.

I had to patch two unit tests (an off-by-one bug and an overfitting bug; both were exposed by the different order in which key/values were returned by iterators).

Some more speed might be gained by using everywhere the standard java.util.Map.Entry interface instead of Element.

### DIFF
--- a/math/pom.xml
+++ b/math/pom.xml
@@ -148,6 +148,12 @@
     </dependency>
 
     <dependency>
+       <groupId>it.unimi.dsi</groupId>
+       <artifactId>fastutil</artifactId>
+       <version>6.6.2</version>
+    </dependency>
+
+    <dependency>
       <groupId>org.slf4j</groupId>
       <artifactId>slf4j-api</artifactId>
     </dependency>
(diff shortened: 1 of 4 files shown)


dlyubimov:
is there an issue for this? Could you please include it like this : MAHOUT-XXX in the header. then close and re-open.

vigna:
> On 25 Mar 2015, at 00:10, Dmitriy Lyubimov <notifications@github.com> wrote:
> 
> is there an issue for this? Could you please include it like this : MAHOUT-XXX in the header. then close and re-open.

Issue:

https://issues.apache.org/jira/browse/MAHOUT-1640

Pull:

https://github.com/apache/mahout/pull/81

Ciao,

					seba

PR #75, opened 2015-02-13
Basic TF-IDF classes

andrewpalumbo:
This is a very basic port of the `TF` and `TFIDF` classes from `o.a.m.vectorizer`.  There are two commits here.  The first uses the Lucene's `DefaultSimilarity` to calculate the weights.  The second removes the Lucene dependency and calculates the weights directly.  As noted in the comments,  the Lucene weights are calculated differently than weights using Spark's `IDF` methods.  The weights are consistent with TF-IDF weights generated using `seq2sparse`.

### DIFF
--- a/math-scala/src/test/scala/org/apache/mahout/nlp/tfidf/TFIDFtestBase.scala
+++ b/math-scala/src/test/scala/org/apache/mahout/nlp/tfidf/TFIDFtestBase.scala
@@ -1,0 +1,153 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
+ * The ASF licenses this file to You under the Apache License, Version 2.0
+ * (the "License"); you may not use this file except in compliance with
+ * the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


andrewpalumbo:
I've added a `MLlibTFIDF` weight for interoperability with MLlib.  I'll push this in the next couple of days if there are no objections.

math-scala/src/test/scala/org/apache/mahout/nlp/tfidf/TFIDFtestBase.scala
```diff
@@ -0,0 +1,184 @@
...
+
+    // dictonary:
+    //  (this -> 0, 4 -> 1, three -> 2, document -> 3, two -> 4, term -> 5, 5 -> 6, contain -> 7,
+    //   each -> 8, first -> 9, terms -> 10, contains -> 11, including -> 12, the -> 13)
+
+    // dfMap:
+    //  (0 -> 1, 5 -> 1, 10 -> 3, 1 -> 1, 6 -> 1, 9 -> 1, 13 -> 2, 2 -> 1, 12 -> 1, 7 -> 1, 3 -> 4,
+    //   11 -> 2, 8 -> 1, 4 -> 1)
+
+    math.abs(vectorizedDocuments(0, 0) -  0.0) should be < epsilon
```
dlyubimov: i think maybe it is worth to import scala.math._ and use abs() directly..

andrewpalumbo: thx.

PR #76, opened 2015-02-18
General cleanup of Scala Similarity and Drivers

pferrel:
Many small cleanup changes:

 * simplified drivers
 * removed any reference to cross-indicator and most references to indicator
 * shortened long lines
 * cleaned up comments and scaladoc annotations
 * replaced old use of o.a.m.math.Pair with Scala tuples

Tested ```spark-itemsimilarity``` on cluster but not naive Bayes stuff

Decided not to remove scopt. Removing it would be more trouble than it's worth given how small the lib is. It still may be worth using case classes for options to get rid of verbose casts but doesn't seem pressing.

This PR doesn't touch the job.jar assembly in the spark module. I have a pare down of that waiting for other refactoring  commits from @dlyubimov.

No other work planned on this PR


spark/pom.xml
```diff
@@ -171,13 +171,14 @@
             </goals>
             <configuration>
               <descriptors>
-                <descriptor>src/main/assembly/job.xml</descriptor>
+                <descriptor>../spark/src/main/assembly/job.xml</descriptor>
```
dlyubimov: use of canonical relative paths is IMO preferred.

spark/src/main/scala/org/apache/mahout/drivers/ItemSimilarityDriver.scala
```diff
@@ -164,12 +154,13 @@ object ItemSimilarityDriver extends MahoutSparkDriver {
         datasetA.dfsWrite(parser.opts("output").asInstanceOf[String] + "../input-datasets/primary-interactions",
           schema = writeSchema)
 
-      // The case of reading B can be a bit tricky when the exact same row IDs don't exist for A and B
-      // Here we assume there is one row ID space for all interactions. To do this we calculate the
-      // row cardinality only after reading in A and B (or potentially C...) We then adjust the
-      // cardinality so all match, which is required for the math to work.
-      // Note: this may leave blank rows with no representation in any DRM. Blank rows need to
-      // be supported (and are at least on Spark) or the row cardinality adjustment will not work.
+      /** The case of reading B can be a bit tricky when the exact same row IDs don't exist for A and B
```
dlyubimov: i thought this is scaladoc style comment. I thought inline comment is just // or /* */I may be wrong.

spark/src/main/scala/org/apache/mahout/drivers/MahoutSparkDriver.scala
```diff
@@ -46,7 +45,7 @@ import org.apache.mahout.sparkbindings._
   *       opts = opts ++ SomeOptions
   *       note("\nAlgorithm control options:")
   *       opt[Int]("maxThings") abbr ("mt") action { (x, options) =>
-  *         options + ("maxThings" -> x) ...
+  *         options + ("maxThings" ->  x) ...
```
dlyubimov: i believe this change introduces wrong style. the prior version was the correct style.

spark/src/main/scala/org/apache/mahout/drivers/RowSimilarityDriver.scala
```diff
@@ -135,7 +133,7 @@ object RowSimilarityDriver extends MahoutSparkDriver {
   }
 
   override def process: Unit = {
-    start()
+    start
```
dlyubimov: no-parenthesis is a proprety style in Scala. E.g. all java getters by default are styled without (). I am guessing start() is really a routine here, not a property. The convention is that a thing without () must not change object state or have any other side effects.  The setters (optionally) then are encoded using `def start_=(value)` form.

spark/src/main/scala/org/apache/mahout/sparkbindings/SparkEngine.scala
```diff
@@ -263,15 +263,15 @@ object SparkEngine extends DistributedEngine {
...
     IndexedDatasetSpark = {
     val reader = new TextDelimitedIndexedDatasetReader(schema)(sc)
-    val ids = reader.readDRMFrom(src, existingRowIDs)
+    val ids = reader.readRowsFrom(src, existingRowIDs)
     ids
   }
 
   /**
-   * reads an IndexedDatasetSpark from default text delimited files
+   * Returns an [[org.apache.mahout.sparkbindings.indexeddataset.IndexedDatasetSpark]] from default text
```
dlyubimov: we probably need to choose between scala-style scaladoc multiline 

      /** Multiline Comment
       */

and Spark (java) -style multiline 

      /**
       * multiline comment
       */

I generally follow spark convention (i.e. scaladoc style for single line and the latter variation for multilines) but in this patch i see mix of both styles even in the same file.

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
```diff
@@ -73,40 +75,44 @@ class MahoutOptionParser(programName: String) extends OptionParser[Map[String, A
...
-      if (x >= 0) success else failure("Option --rowIDColNum must be >= 0")
+    } text ("Column number (0 based Int) containing the row ID string (optional). Default: 0") validate {
+      x => if (x >= 0) success else failure("Option --rowIDColNum must be >= 0")
     }
 
     opt[Int]("itemIDColumn") abbr ("ic") action { (x, options) =>
       options + ("itemIDColumn" -> x)
-    } text ("Column number (0 based Int) containing the item ID string (optional). Default: 1") validate { x =>
-      if (x >= 0) success else failure("Option --itemIDColNum must be >= 0")
+    } text ("Column number (0 based Int) containing the item ID string (optional). Default: 1") validate {
```
dlyubimov: for multiline closures i generally prefer to see a line feed after argument line 

    ... { x => 
        do something 
    }

style as opposed to 

    ... {  
        x => do something 
    }

I guess either is fine but in this patch i see both styles.

andrewpalumbo:
Tested  `TrainNBDriver` and `TestNBDriver` from the command line with no problems.  I need to do some cleanup of the rest of the Naive Bayes classes as well.

mrlegacy/src/main/java/org/apache/mahout/classifier/ConfusionMatrix.java
```diff
@@ -36,7 +36,7 @@
 /**
  * The ConfusionMatrix Class stores the result of Classification of a Test Dataset.
  * 
- * The fact of whether there is a default is not stored. A row of zeros is the only indicator that there is no default.
+ * The fact of whether there is a default is not stored. A row of zeros is the only similarity that there is no default.
```
andrewpalumbo: An accidental replacement of "indicator" with "similarity" here?

math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
```diff
@@ -127,12 +129,11 @@ object SimilarityAnalysis extends Serializable {
     retIDSs.toList
   }
 
-  /** Calculates row-wise similarity using the log-likelihood ratio on AA' and returns a drm of rows and similar rows
+  /** Calculates row-wise similarity using the log-likelihood ratio on AA' and returns a DRM of rows and similar rows
     * @param drmARaw Primary interaction matrix
     * @param randomSeed when kept to a constant will make repeatable downsampling
     * @param maxInterestingSimilaritiesPerRow number of similar items to return per item, default: 50
     * @param maxNumInteractions max number of interactions after downsampling, default: 500
-    * @return
```
andrewpalumbo: Was this `@return` meant to be removed?

spark/src/main/scala/org/apache/mahout/drivers/ItemSimilarityDriver.scala
pferrel: yes, the only comment guidelines are in the scaladoc section of the style guide so I follow them everywhere except a few one liners.

spark/src/main/scala/org/apache/mahout/drivers/MahoutSparkDriver.scala
pferrel: an extra space didn't seem worth the commit history. The original substantive change was a missing ">" typo if you care.

spark/src/main/scala/org/apache/mahout/drivers/ItemSimilarityDriver.scala
dlyubimov: well the point is this is not scaladoc comment at all. As in no scaladoc is generated here. This is an inline comment. therefore it should not be marked with /**.

spark/src/main/scala/org/apache/mahout/drivers/RowSimilarityDriver.scala
pferrel: The function is an Arity-0 method. Perhaps I misread: http://docs.scala-lang.org/style/method-invocation.html

Arity-0
Scala allows the omission of parentheses on methods of arity-0 (no arguments):

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
pferrel: yep, trying to shorten lines.

pferrel: The larger issue is probably that infix is used and that is because the referenced examples for this code use infix as Scala style dictates. Unless we want to make some decision about that I have no problem with mixing infix and dot styling. These are more of a DSL than a dot method call if that make any difference.

spark/src/main/scala/org/apache/mahout/drivers/RowSimilarityDriver.scala
dlyubimov: quoting from your reference:
`However, this syntax should *only* be used when the method in question has
no side-effects (purely-functional). In other words, it would be acceptable
to omit parentheses when calling queue.size, but not when calling println().
`

Kind of what i am saying. No side effects rule is violated (and, frankly,
it is violated by anything but a computed property).

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
dlyubimov: if we don't have enough space we still can do a new line after closure params, right?  i.e. 

    blah blah blah .... {
      x =>
      do something else 
    }


it's just looks inconsistent with that other closure style pattern you use elsewhere.

math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
pferrel: Yeah, they (Scala style) say to not put it in if the description says what is returned. Debatable here.

mrlegacy/src/main/java/org/apache/mahout/classifier/ConfusionMatrix.java
pferrel: thought I got those, thanks

spark/src/main/scala/org/apache/mahout/drivers/RowSimilarityDriver.scala
pferrel: You should have been a lawyer, I'll fix

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
pferrel: Wouldn't that be 

    blah blah blah ... {
      x =>
        do something
    }

At least that's what my editor wants to do.

math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
dlyubimov: this is Mahout's position on this with javadoc style as well afaik. Not to leave empty @@tags if they are already obvious.

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
dlyubimov: i think either is fine. (fwiw idea scala plugin, which is the only scala plugin out there that actually works, does not add extra indentation inside curly brackets regardless).

spark/src/main/scala/org/apache/mahout/drivers/RowSimilarityDriver.scala
dlyubimov: imo this is no small convention. at least i found it incredibly useful since they expand syntax of an attribute here that significant way. Note also attribute-esque nature of the convention since they also augmented 0-arity call with an assignemt "magic" syntax def name_=(value) which in code will just be an attribute assignment semantics such as `name = value`.

So this is not random.

pferrel:
As soon as this looks ok, I'd like to work on the getting 1.2.1 in next so let me know

spark/pom.xml
dlyubimov: i still don't understand why canonical path is replaced with non-canonical path here.

pferrel: No idea. I don't remember doing anything with that, I'll look.

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
pferrel: Must be a setting because that was from Idea

math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
andrewpalumbo: I see... makes sense.

andrewpalumbo:
The only thing that stands out to me is the move to scaladoc style doc comments (away from the spark/java style doc comments).  I'm not sure if it matters, but as Dmitriy mentioned above, we should really pick one style to go with.  Spark style seems to be the de facto convention.  It would make it easier to point people to the Spark style guide as discussed on the dev list.

pferrel:
I don't recall a decision on that or I wouldn't have set my line width to 120. If we are going to be that exacting let's make sure everyone agrees and I'll change this later if necessary. I need to get Spark 1.2.1 in this week if possible

andrewpalumbo:
I don't think we really made any hard decision on it- just discussed it a bit..  I don't see any reason not to commit and revise later after we decide.

pferrel:
Let me see what I can do. It should be easy to change the scaladoc comments, it's just the first line. 100 chars will be harder. Would be nice to get it over with.

dlyubimov:
i'd suggest to be ok with 120 characters. at least our java style is 120-line-long compliant.

Spark people were also ok with 120 characters in my contributions (in fact, initially 100 character constraint applied only to comment style, not the code there. For reasons nobody truly knows).

pferrel:
That will make it much easier, ok Andrew?

andrewpalumbo:
I was only referring to the comment style ... not the 100 chars limit.

andrewpalumbo:
Basically what we were talking about on this thread: 

http://mail-archives.apache.org/mod_mbox/mahout-dev/201501.mbox/%3CBLU436-SMTP226CD60AFFD7C2AE452278694320%40phx.gbl%3E

pferrel:
So something like this?

 * javadoc style documentation comments like Spark
 * For Scala follow: http://docs.scala-lang.org/style/
 * We are relaxing the required use of infix for *all* appropriate method calls and encouraging
“dot” notation. Spark goes further to say you must not use infix unless it’s an operator
method.
 * Line lengths not to exceed 120 chars—Spark limits to 100
  

I'm all for using infix as little as possible.  It really does make 
scala unnecessarily difficult to read for those  coming from other 
languages.

  I think we're really only using using infix (aside from operators) in 
tests where it seems useful and relatively straightforward. eg:

     InCore(dslCat2, 3) - aggInCore(dslCat2, 3) should be < epsilon

dlyubimov:
+1. General notion is that custom dsls are big consumers of infix. Indeed.
but OOA should not be one.

andrewpalumbo:
Yes- that's what I was thinking.  On the list there is also a link to the Spark Code Style  giude:

https://cwiki.apache.org/confluence/display/SPARK/Spark+Code+Style+Guide

and I was suggesting that we use the (javadoc) comment style from that rather than the scaladoc style.  Since it seems to be what a lot of projects are adopting and what we presently use in our codebase.

pferrel:
OOA?

andrewpalumbo:
oh sorry i didn't see that you had "javadoc style documentation comments like Spark" in that list.

andrewpalumbo:
+1 on doc commit.

pferrel:
OK, well thanks. Hopefully this cleans my desk up a bit.

andrewpalumbo:
>BTW I like being able to use infix for little DSLs even though they don't use operators they are sometimes easier to read--like the test ones.

agreed

pferrel:
Since 1.2.1 was pushed I had to merge those changes but they break execution of spark-itemsimilarity due to HashBiMap class missing at deserialization.

This is a known bug in Spark 1.2.x with at work around that I am trying.

pferrel:
Seems like we should avoid 1.2 and 1.2.1 though we could tell someone how to make it work and only item and rowsimilairty are broken anyway.

So I'm testing this with Spark 1.1.1 and will push that for now. All that is required to upgrade to 1.2 or 1.2.1 would be the root pom Spark version change and build from source

PR #77, opened 2015-02-20
MAHOUT-1642 fix missing first element in SimilarItemsIterator

OlegZotov:


fape:
Could you please someone tell me why this was closed without merged?

OlegZotov:
After I opened it and some time passed, I asked in mailing list why nobody merge it and received the following response:
"Part of the problem is that this is a fix to taste recommender.
There are two problems with this: first, it is mrlegacy component, and nobody these days cares so much maintaining these. Second, the author of this code, who would be qualified to assess the validity of the fix, has not come forward, which kind of goes back to the first point."
So ibid in mailing list it was decided to close the pull request.

PR #78, opened 2015-03-07
MAHOUT-1493 driver cleanup

andrewpalumbo:
This removes parseIndexedDatasetFormatOptions() and some other option parsers not used by `spark-trainnb` and `spark-testnb`.  There are still a couple of options that I need to add but first want to figure out how to handle the  required `-o` option.

Currently `spark-testnb` doesn't have any actual output.  I was thinking about writing a confusion matrix as Text out to a file.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
@@ -50,7 +50,7 @@ class MahoutOptionParser(programName: String) extends OptionParser[Map[String, A
         "(optional). Default: empty.")
     }
 
-    opt[String]('o', "output") required() action { (x, options) =>
+    opt[String]('o', "output") action { (x, options) =>
       if (x.endsWith("/")) {
         options + ("output" -> x)
       } else {
--- a/spark/src/main/scala/org/apache/mahout/drivers/TestNBDriver.scala
+++ b/spark/src/main/scala/org/apache/mahout/drivers/TestNBDriver.scala
@@ -18,7 +18,6 @@
 package org.apache.mahout.drivers
 
 import org.apache.mahout.classifier.naivebayes.{NBModel, NaiveBayes}
-import org.apache.mahout.classifier.stats.ConfusionMatrix
 import org.apache.mahout.math.drm
 import org.apache.mahout.math.drm.DrmLike
 import scala.collection.immutable.HashMap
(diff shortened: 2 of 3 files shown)


dlyubimov:
Seems clean to me.

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
```diff
@@ -50,7 +50,7 @@ class MahoutOptionParser(programName: String) extends OptionParser[Map[String, A
         "(optional). Default: empty.")
     }
 
-    opt[String]('o', "output") required() action { (x, options) =>
+    opt[String]('o', "output") action { (x, options) =>
```
andrewpalumbo: @pferrel would changing this from a required option break anything for you?

andrewpalumbo:
cool- thanks for looking at it @dlyubimov.

math-scala/src/main/scala/org/apache/mahout/drivers/MahoutOptionParser.scala
pferrel: You could add a default param to the function, like:

def parseIOOptions(numInputs: Int = 1, noOutput: Boolean = false)

Then a conditional around the requied(). This will work fine for everyone else with no change.

pferrel: or just don't call that helper and put your own output option in your driver. I made these helpers to keep some options standardized but no hard rule.

pferrel:
@andrewpalumbo see comments about the "required" change. I wouldn't push this as-is. It will allow illegal params for other drivers.

andrewpalumbo:
@pferrel thanks- I'll change it back and probably just add in a -i option and remove `parseIOOptions()` from `testNBDriver` as you suggested.

pferrel:
Sounds good

dlyubimov:
BTW @andrewpalumbo could you please change the header to show the issue like this: MAHOUT-1493,  then close and repopen it. Once re-opened, @asfgit bot will start synching jira to reviews and backwards.

andrewpalumbo:
thx - was gonna do this under a new issue, but since the branch is for M-1493 will finish this up there.

dlyubimov:
well if you decide to get another issue, just edit the header again and re-open again. since current issue seems to be closed. or just re-open the current jira issue.

andrewpalumbo:
fixing in a different branch

PR #81, opened 2015-03-25
Sparse-vector implementation based on fastutil (MAHOUT-1640)

vigna:
The collections currently used by Mahout to implement sparse vectors are extremely slow. The proposed patch (localized to RandomAccessSparseVector) uses fastutil's maps and the speed improvements in vector benchmarks are very significant. It would be interesting to see whether these improvements percolate to high-level classes using sparse vectors.

I had to patch two unit tests (an off-by-one bug and an overfitting bug; both were exposed by the different order in which key/values were returned by iterators).

Some more speed might be gained by using everywhere the standard java.util.Map.Entry interface instead of Element.


dlyubimov:
still need to look at this.

dlyubimov:
Ok here are the quick sparse multiplication performance results in our unit tests. In short, looks like 30% improvement for sparse opeartions on average. Which is decent. I will test the rest of the tests and report back. If everything is ok, looks like a very valuable addition for the time being. Thank you very much @vigna !

Fastutil PR over current master (the second numbers are pragmatically important): 

Asr %*% Bsr: (26.0,14.333333333333334)
Asr' %*% Bsr: (1214.0,16.0)
Asr %*% Bsr': (96.0,19.0)
Asr' %*% Bsr': (1002.6666666666666,13.666666666666666)
Asr'' %*% Bsr'': (14.333333333333334,14.333333333333334)

Asm %*% Bsm: (1523.3333333333333,18.0)
Asm' %*% Bsm: (1929.0,21.0)
Asm %*% Bsm': (1428.6666666666667,20.666666666666668)
Asm' %*% Bsm': (1669.6666666666667,17.333333333333332)
Asm'' %*% Bsm'': (1541.3333333333333,14.333333333333334)

Asm %*% Bsr: (1160.0,14.0)
Asm' %*% Bsr: (1609.3333333333333,20.333333333333332)
Asm %*% Bsr': (1142.3333333333333,17.666666666666668)
Asm' %*% Bsr': (1408.6666666666667,14.333333333333334)
Asm'' %*% Bsr'': (1154.6666666666667,14.0)

Current master:

Asr %*% Bsr: (30.0,22.0)
Asr' %*% Bsr: (1918.6666666666667,26.0)
Asr %*% Bsr': (118.66666666666667,27.0)
Asr' %*% Bsr': (1551.6666666666667,21.666666666666668)
Asr'' %*% Bsr'': (22.333333333333332,21.666666666666668)

Asm %*% Bsm: (2191.0,26.333333333333332)
Asm' %*% Bsm: (2543.6666666666665,30.333333333333332)
Asm %*% Bsm': (2067.0,30.333333333333332)
Asm' %*% Bsm': (2274.0,22.666666666666668)
Asm'' %*% Bsm'': (2167.0,22.666666666666668)

Asm %*% Bsr: (1717.3333333333333,21.666666666666668)
Asm' %*% Bsr: (2254.0,32.0)
Asm %*% Bsr': (1762.3333333333333,26.666666666666668)
Asm' %*% Bsr': (2004.6666666666667,21.666666666666668)
Asm'' %*% Bsr'': (1711.0,24.333333333333332)

vigna:
If you merge this, I promise I'll look into the other improvements (e.g., not using Element). :)

dlyubimov:
@vigna i apologize it took so long. it completely went off the radar.

dlyubimov:
@vigna is 0.7.2 fastutil is still the best version to use? I can't immediately find the license on it?
@smarthi et. al. : need a few votes on inclusion of fastutil as a dependency

dlyubimov:
ok all tests are passing for me.

vigna:
@dlyubimov no, actually there are newer versions. I am doing a release in few days with a lot of small glitches fixed thanks to the Guava test suite. Current version is 7.0.10.

dlyubimov:
so it's Apache 2.0 now, great.

smarthi:
Seems like it's ASL 2.0 -
https://github.com/vigna/fastutil/blob/master/LICENSE-2.0

+1 from me, good to go.

andrewpalumbo:
+1 from me too

smarthi:
@vigna Please also update the LICENSE.Txt file for fastutil.

https://github.com/apache/mahout/blob/master/LICENSE.txt

Thanks again for this.

vigna:
@smarthi Do you want me to modify that file and add "fastutil" to the list of projects to which the license applies?

vigna:
BTW, I noticed that SparseMatrix can undergo the same improvement with near-to-zero effort. Do you want me to pack everything into this pull request or do you prefer to merge this part and then I'll work on SparseMatrix?

smarthi:
Yes

smarthi:
Let's merge this first, please create another PR for sparsematrix. Thanks again for this.

vigna:
It's all done.

smarthi:
Thanks @vigna.  We'll merge this in the next day or two.

pferrel:
Not sure it's still needed but +1 from me. I've used fastutils in other projects quite successfully. This might also be the basis for faster BiMap implementation. Will look into this for a future release.

r-tock:
Its great to see this getting merged!

smarthi:
@vigna https://issues.apache.org/jira/browse/MAHOUT-1801

dlyubimov:
@vigna if i may suggest a topic: we need a fast way to approximately estimate sparsity inside a matrix and reconstitute it into a dense matrix (with a given row- or column-major order) and vice versa. It is right now a big performance issue with distributed (and in-memory) algebra as many distributed operators tend to always form sparse structures even if result doesn't warrant it (completely dense).

so it would be nice to have a method on a matrix, something like "sparsityAnalysis" that accepts sparsity threshold (double) and returns a matrix: self if sparsity is within threshold for the current data structure, or rewrite itself as a dense/sparse counterpart using requested row or column major order (this is known as part of matrix "flavor").

dlyubimov:
@vigna so for example: 

    def MatrixOps$sparsityAnalysis(sparsityThreshold:Double=.25):Matrix 

i have SparseRowMatrix `mxA` which is 95% dense. I then call 

    mxA = mxA.sparsityAnalysis()

and now mxA is a DenseMatrix.

This internally will of course rely on some implementation-specific api that can tell us sparsity estimate (or that it doesn't know how to estimate sparsity).

smarthi:
@dlyubimov created M-1801 today, its best to capture the above comments in the Jira there.

vigna:
Er... this says that there are unmerged commits. Why?

smarthi:
where u see that @vigna ?

vigna:
Oh well. It just vanished. I'm new to this stuff. Probably something happened in the last few minutes.

vigna:
I created a new pull request https://github.com/apache/mahout/pull/186 for the sparse matrix implementation.

PR #83, opened 2015-03-26
Initial split of mrlegacy into hdfs and mr

andrewmusselman:
Moving mrlegacy directory to hdfs, starting minimal mr directory, repointing references to mrlegacy everywhere.


pferrel:
There should be far less in mahout-hdfs than mahout-mr according to Dmitriy so I'd have started with nothing in mahout-hdfs and depend only on that in math-scala and spark. Start by migrating the missing classes. The debugger (IntelliJ IDEA) will flag missing definitions so much of it can be done with copy and pasting of files between mahout-mr and mahout-hdfs

pferrel:
Obviously too many changes for diffing, sorry.

andrewmusselman:
Moved the contents of hdfs over to mr since there are more things staying in mr apparently.

dlyubimov:
i can't seem to see mods to spark/pom.

only the following files need to go to the shared module are really necessary dependency for /spark and spark-shell (unless something new happened):

/MatrixWritable.java
/MatrixWritableTest.java
/Varint.java
/VarintTest.java
/VarIntWritable.java
/VarLongWritable.java
/VectorWritable.java
/VectorWritableTest.java

There also was that Pair thing but it just needs to be replaced by regular scala's Tuple2 IMO in the code that uses it

pferrel:
Pair was remove awhile back

dlyubimov:
also def findMahoutJars needs to be adjusted to reflect new backend deps.

andrewmusselman:
Yeah I don't see the spark/pom changes too but they're just like the others plus removal of a large comment for profiles.

andrewmusselman:
I don't find "findMahoutJars" anywhere; where is that?

dlyubimov:
@andrewmusselman you need to fix filters line ~203 of ...sparkbindings/package.scala in the spark module.

mrlegacy out, hdfs in.

dlyubimov:
@andrewmusselman also: this code relies on reading output of `mahout -spark classpath`. Please double check running this is producing correct mahout jars. please double check the hdfs jar is reported there correctly.

pferrel:
findMahoutContextJars

dlyubimov:
Also: there's test that tests the classpath correct detection in SparkBindginsSuite. It is "ignored" because it requires correct MAHOUT_HOME and mahout compiled (which is not the case during CI build). But you can change "ignore" to "test" and try to run it. It should print the mahout spark context path and also asserts that it fished out exactly 4 jars there.

pferrel:
will fail then because the new dependency-reduced jar should also be included making it 5

dlyubimov:
why? we are replacing, not adding

dlyubimov:
anyhow, nothing fails on my branch.

Granted, i don't run stuff like co-occurrence though.

pferrel:
Because Guava and Scopt are required for Drivers and they should be all that is included in that assembly jar

pferrel:
Actually IndexedDataset requires guava, drivers require scopt, and cooccurrence requires neither.

dlyubimov:
No.

Guava is included with hadoop. Assuming spark backend is set up correctly
with Mahout, it should see Guava 11 classes.

Scopt (cli option parser) must not be needed in the workers.

pferrel:
Right, scopt is not needed for workers.

Guava is shaded in Spark (as of 1.2) and not sure what is in Hadoop. I ran into missing classes from Guava until I provided the jar and anyhow it makes the jar brew rather fragile if we rely on external things accidentally provided in Mahout or Spark as the move to Spark 1.2 shows.

All this makes me think that explicit jars should be provided in the future, not discovered. Spark stores optional jars in an "External" directory. Leave it for 0.10.1 though where we may want to rethink how we launch things.

dlyubimov:
Guava is shaded in spark exactly for the reason that hadoop already have
guava 11.  naming clashes caused conflict. there is no other reason.

So if anything, there's both guava 11 and 14 in the backend (although 14 is
shaded, and wouldn't be picked up). shipping yet another one is not only
unnecessary, but actually may mess it all up again.

dlyubimov:
As far as "reliance", i don't see how we can change spark java namespaces.
afaik spark doesn't isolate worker namespace from its own or hadoop's
dependencies (nor it probably should, since everything is built for
compatibility on class serialization level).

dlyubimov:
Although i suppose we could shade guava once again independently of spark
and hadoop, but i won't do it (for lack of "shading" expertise), and at any
rate, it would just inflate my job submission once again even though none
of my jobs are using guava. Guava use is exclusive to the co-occurrence
recommender. Therefore, if it needs to be listed for a particular method,
it needs to be listed in "customJars" attribute to the
"createMahoutContext" call. That's why it exists. To add jars not everyone
needs.

Aside from that, dependency conflicts are always a big issue, down to the
point of nightmares (e.g. try to embed spark context into a play
application just for the fun of it), and actually mahout is having it
super-easy so far. Believe me. Even spark having these nightmares with
hadoop, it looks like. We are nothing new really in that.

pferrel:
Fair point about custom jars feel free to create a Jira for 0.10.1 BTW I was you who asked for an assembly jar with "included" dependencies only when we discussed this exact issue before. Also feel free to change the code.

dlyubimov:
not sure what you are referring to. We never discussed backend classpaths.

we (well, I) discussed front end classpath, yes, and only for the issue of
scopt inclusion into it, which is 180 degrees different issue. But i never
suggested to change a bit in the backend classpath. Actually, if anything,
by doing the refactoring, i effectively reduced it to the minimum necessary
by throwing away mrlegacy behemoth.

pferrel:
actually custom jars won't work here since it is in cooccurrence, not the driver.

I'll give this a spin tomorrow.

pferrel:
OK so no classes have been moved, just the structure created?

I'll start by moving the classes Dmitriy lists. The job jar should be part of mahout-mr so that will be moved.

andrewmusselman:
That's right, just the structure; let me know what I can do next.

pferrel:
All other pending PRs will be messed up with this so starting now.

pferrel:
intermediate branch that compiles here: https://github.com/pferrel/mahout/tree/MAHOUT-1655

PR #88, opened 2015-03-29
MAHOUT-1659

magsol:
Removed the dependency on the Lanczos solver from spectral clustering. Now exclusively uses SSVD for dimensionality reduction.


smarthi:
Looks good Shannon, a few other changes that would be good to have in this PR

1.  Replace all Guava API calls in DisplaySpectralKMeans.java with the appropriate Java 7 api.
2.  think we should now purge Lanczos for good. Correct? If so please either create a new Jira for that or update this Jira with deprecated Lanczos.   Regardless Lanczos needs to be marked as deprecated in the code.

PR #89, opened 2015-03-30
MAHOUT-1662

magsol:
Changed how the qualified Path is determined so as not to append itself repeatedly. Requires verification.


magsol:
New PR forthcoming; same edits.

PR #91, opened 2015-03-31
MAHOUT-1564 Naive Bayes Classifier for New Text Documents

andrewpalumbo:
I've decided to add this as a spark-shell script example. It can serve as an example of running mahout spark-shell scripts and integrating spark-naive bayes into scala code and  can be easily adapted to an application.


andrewpalumbo:
This script creates a classifier for the Wikipedia dataset and depends on MapReduce seqwiki and seq2sparse.

PR #95, opened 2015-04-02
Bump spark-shell spark.executor.memory up to 1g (default is 512m).

andrewpalumbo:
also a couple of cosmetic fixes for driver version numbers.

Any reason not to move the  spark.executor.memory up ?

### DIFF
--- a/spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
+++ b/spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
@@ -45,6 +45,8 @@ class MahoutSparkILoop extends SparkILoop {
       conf.set("spark.executor.uri", execUri)
     }
 
+    conf.set("spark.executor.memory", "1g")
+
     sparkContext = mahoutSparkContext(
       masterUrl = master,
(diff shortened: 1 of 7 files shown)


dlyubimov:
I forget, did we fix MAHOUT_OPTS env  for spark shell in public version? Generally we should be able to run it with MAHOUT_OPT="-Dspark.executor.memory=whatever", right?

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
```diff
@@ -45,6 +45,8 @@ class MahoutSparkILoop extends SparkILoop {
       conf.set("spark.executor.uri", execUri)
     }
 
+    conf.set("spark.executor.memory", "1g")
```
dlyubimov: Defaults should be overwritten from MAHOUT_OPTS, right? but it doesn't look like it checks if it is already set?

andrewpalumbo:
No- we pushed that fix for 0.10.1

dlyubimov:
ok

andrewpalumbo:
thx for looking at it. gonna push now.

PR #96, opened 2015-04-02
MAHOUT-1662

magsol:
Fix for a path bug that would append the path of the data to itself and crash. Same as previous, but done on a branch of my fork so I can simultaneously submit other fixes.


smarthi:
There r other places in the code that have this. Its a good idea to fix all of them in a single commit. Its outcome of a previous commit from me that went horribly wrong.

magsol:
Are there more details on this? i.e. why this change was originally made,
what it was supposed to fix, etc?

dlyubimov:
(noted)

smarthi:
@magsol - search for makeQualified() in the legacy and anything that has an extra path in the args needs to be reverted back to original shape and form. 

Why was this change done?  Was originally intended to be upgraded all deprecated MR apis to Hadoop 2, but the attempt went all Whack!! . So much for trying to push code at 4:30AM.

smarthi:
@magsol verified ur patch and Looks Good to Merge, will commit this to master

PR #72, opened 2015-01-27
MAHOUT-1493-h2o: Enable Naive Bayes Tests in h2o Module

andrewpalumbo:

PR #99, opened 2015-04-03
MAHOUT-1638: implement String equivalent of Vec.makeCons()

avati:
Necessary for creating an empty Vec to hold String labels in the
future.

Replace calls to Vec.makeCons with H2OHelper.makeEmptyStrVec()
where the created Vec is used to store label strings.

Also, fix an off-by-one bug in MapBlockHelper


andrewpalumbo:
@avati thanks alot for the fix.  Looks good!  I'll push this. along with #72

dlyubimov:
This is superawesome. Thank you @avati !

PR #107, opened 2015-04-05
MAHOUT-1605: fixes in VisualizerTest

avati:
Support decimal separator of various locales


FRosner:
Thanks for upgrading to Java 7!

PR #111, opened 2015-04-05
MAHOUT-1493:  Add CLI options for --overwrite and --alphaI  to NB Drivers 

andrewpalumbo:
Presently `mahout spark-trainnb` will not complete if a model already exists in the output directory.  These last options add in an `--overwrite` option to overwrite a model in the given output directory.  

as well:
  - adds `.par(auto = true)` to the input Drm
  - adds a `delete(...)` method to `Hadppo1HDFSUtils` which does not handle any IO exceptions
  - adds an almost trivial  `--alphaI` option to set the Laplace smoothing factor from the CLI

This patch will complete the full port of the old MapReduce Naive Bayes to the `math-scala` and `spark` modules.

### DIFF
--- a/spark/src/main/scala/org/apache/mahout/drivers/TestNBDriver.scala
+++ b/spark/src/main/scala/org/apache/mahout/drivers/TestNBDriver.scala
@@ -77,7 +77,7 @@ object TestNBDriver extends MahoutSparkDriver {
   /** Read the test set from inputPath/part-x-00000 sequence file of form <Text,VectorWritable> */
   private def readTestSet: DrmLike[_] = {
     val inputPath = parser.opts("input").asInstanceOf[String]
-    val trainingSet = drm.drmDfsRead(inputPath).par(auto = true)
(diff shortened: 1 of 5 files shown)


spark/src/main/scala/org/apache/mahout/drivers/TestNBDriver.scala
```diff
@@ -76,7 +77,7 @@ object TestNBDriver extends MahoutSparkDriver {
   /** Read the test set from inputPath/part-x-00000 sequence file of form <Text,VectorWritable> */
   private def readTestSet: DrmLike[_] = {
     val inputPath = parser.opts("input").asInstanceOf[String]
-    val trainingSet = drm.drmDfsRead(inputPath)
+    val trainingSet = drm.drmDfsRead(inputPath).par(auto = true)
```
andrewpalumbo: @dlyubimov, @pferrel  -does adding `.par(auto = true)` here make sense?  Otherwise there are no explicit partitioning instructions for the input data.

PR #116, opened 2015-04-10
move Hadoop1HDFSUtils into a new h2o package to not coflict with spark

andrewpalumbo:


dlyubimov:
could they perhaps be using the same thing? 
(Question for Anand probably)

andrewpalumbo:
Oops.. sorry forgot to close this.. this has been committed. We needed to make this change because packages were conflicting in the binary release. Stevo has a branch in which he's factored both into a new `scala-hdfs` module so we can do that for 0.10.1.

PR #117, opened 2015-04-12
MAHOUT-1653 Spark-1.3 for 0.10.0

andrewpalumbo:
Uses a  deprecated method in `CheckpointedDrmSpark.scala` and has the shell build disabled.

### DIFF
--- a/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
+++ b/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
@@ -30,6 +30,8 @@ import org.apache.hadoop.io.{LongWritable, Text, IntWritable, Writable}
 import org.apache.mahout.math.drm._
 import org.apache.mahout.sparkbindings._
 import org.apache.spark.SparkContext._
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.RDD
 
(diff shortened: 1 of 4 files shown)


spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -165,7 +167,8 @@ class CheckpointedDrmSpark[K: ClassTag](
       else if (classOf[Writable].isAssignableFrom(ktag.runtimeClass)) (x: K) => x.asInstanceOf[Writable]
       else throw new IllegalArgumentException("Do not know how to convert class tag %s to Writable.".format(ktag))
 
-    rdd.saveAsSequenceFile(path)
+    // this is a (working) deprecated method used as a stop-gap while we investigate the shell issues
```
dlyubimov: ok stop-gap but perhaps we can do it on top of #135 ..

andrewpalumbo: yeah- this definitely can wait..  I'll re-synch it after #135 is pushed and we can keep it here as a patch for people that want to use 1.3 on the 0.10.x branch.

pferrel:
This should be closed in favor of #136 so it goes into 0.11.0

andrewpalumbo:
see #136

PR #120, opened 2015-04-21
MAHOUT-1693: FunctionalMatrixView materializes row vectors in scala shell

andrewpalumbo:
Override .toString() to simply return the class name in  FunctionalMatrixView.java solves this problem.


smarthi:
Also add a test :
val inCoreB = Matrices.symmetricUniformView(50000, 50000, 1)

that demonstrates this problem to not occur after this fix

PR #121, opened 2015-04-21
MAHOUT-1693: Override .toString() in AbstractMatrix using VectorView

andrewpalumbo:
I still have to test this out a bit, but this is a fix for the memory issues caused by the shell triggering .toString() after the instantiation of large  DenseMatrix, SparseMatrix and FunctionalMatrixViews (MAHOUT-1693).  This is set to display an (arbitrarily sized)10x20 upper left block of the matrix.  For sparse matrices this does affect the printout, so it may not be acceptable.

### DIFF
--- a/math/src/main/java/org/apache/mahout/math/AbstractMatrix.java
+++ b/math/src/main/java/org/apache/mahout/math/AbstractMatrix.java
@@ -780,15 +780,53 @@ public abstract class AbstractMatrix implements Matrix {
     }
   }
 
+//  @Override
+//  public String toString() {
+//    int row = 0;
+//    int col = 0;
+//    int maxRowToDisplay = 10;
+//    int maxColToDisplay = 50;
+//    StringBuilder s = new StringBuilder("{\n");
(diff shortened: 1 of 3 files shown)


math/src/main/java/org/apache/mahout/math/AbstractMatrix.java
```diff
@@ -782,13 +782,34 @@ public boolean isAddConstantTime() {
 
   @Override
   public String toString() {
+    int row = 0;
+    int maxRowsToDisplay = 10;
+    int maxColsToDisplay = 20;
```
andrewpalumbo: If there are no objections to a max 10x20 matrix to display, or anything else, i will commit this to master and the 0.10.1 branch.

andrewpalumbo:
I'm going to push this to master tonight since it's a better fix than we have in place already, and I'm not sure how much time I'll have tomorrow.  We can always adjust the size of the matrix block to be displayed.

smarthi:
10 * 20 to display is fine IMO. +1 to push this to master.

PR #125, opened 2015-04-29
Spark 1.2

pferrel:
Created a Scala based BiMap and BiDictionary and removed usage of Guava HashBiMap (actually no Guava is used now), I hope this gets past the Javaserializer bug in https://issues.apache.org/jira/browse/SPARK-6069

Also found an error that did an unneeded collect, which will drastically reduce memory requirements in reading in data for cooccurrence.

Passing around ```Option[BiDictionary] = None``` instead of an empty dictionary, which seems a better idiom when using an immutable BiDictionary.

Questions:
* Passes local build tests but haven't tried on a cluster to verify it works in spite of SPARK-6069
* Need to try it on epinions data before pushing
* Where should this go, it's in 0.11.0-SNAPSHOT now
* Still need to try putting Scala libs in mahout shell module only. This is to check to see if they are needed in non-shell Scala modules

I'm having trouble getting HDFS+Spark running, even standalone, so if anyone wants to try this I can provide instructions.


dlyubimov:
On Wed, Apr 29, 2015 at 11:53 AM, Pat Ferrel <notifications@github.com>
wrote:

> Created a Scala based BiMap and BiDictionary and removed usage of Guava
> HashBiMap (actually no Guava is used now),
>

hurray to that!

math-scala/src/main/scala/org/apache/mahout/math/indexeddataset/BiMap.scala
```diff
@@ -0,0 +1,129 @@
...
+/** BiDictionary is a specialized BiMap that has non-negative Ints as values for use as DRM keys.
+  * The companion object provides modification methods specific to maintaining contiguous Int values
+  * and unique String keys */
+object BiDictionary {
+
+  /**
+   * Append new keys to an existing BiDictionary and return the result. The values will start
+   * at m.size and increase to create a continuous non-zero value set from 0 to size - 1
+   * @param keys new keys to append
+   * @return a BiDictionary with added mappings
```
tedyu: Add @param for biDi

math-scala/src/main/scala/org/apache/mahout/math/indexeddataset/BiMap.scala
```diff
@@ -0,0 +1,129 @@
...
+}
+
+/** BiDictionary is a specialized BiMap that has non-negative Ints as values for use as DRM keys.
+  * The companion object provides modification methods specific to maintaining contiguous Int values
+  * and unique String keys */
+object BiDictionary {
+
+  /**
+   * Append new keys to an existing BiDictionary and return the result. The values will start
+   * at m.size and increase to create a continuous non-zero value set from 0 to size - 1
```
tedyu: m.size -> biDi.size

spark/src/main/scala/org/apache/mahout/drivers/TextDelimitedReaderWriter.scala
```diff
@@ -279,17 +282,20 @@ trait TDIndexedDatasetWriter extends Writer[IndexedDatasetSpark]{
 
         // first get the external rowID token
         if (!vector.isEmpty){
-          var line: String = rowIDDictionary.inverse.get(rowID) + rowKeyDelim
+          //var line: String = rowIDDictionary.inverse.get(rowID) + rowKeyDelim
```
tedyu: This line can be dropped.

pferrel:
@dlyubimov first Scala collection class design so any thoughts on BiMap or BiDictionary are appreciated.   The append and create do not use operators because they seem pretty non-standard though I guess the create could just be another constructor in the companion object.

andrewpalumbo:
It looks like only the spark-shell module needs the added scala dependencies.  I moved them there and everything built and tested fine.

pferrel:
Great, shall I  merge with this branch?

math-scala/src/main/scala/org/apache/mahout/math/indexeddataset/BiMap.scala
```diff
@@ -0,0 +1,122 @@
...
+   * @param keys new keys to append
+   * @param biDi append keys to this BiDictionary and create new values buy incremeting from the highest Int value
+   * @return a BiDictionary with added mappings
+   */
+  def append(keys: List[String], biDi: BiDictionary): BiDictionary = {
+    val hm = HashMap(keys.view.zip (Stream from biDi.size): _*)
```
pferrel: Seems like a Seq will do for most collections of String so only one append method, I'll change and test. Not sure if there are significant performance issues. The use of .view is very subtle, not sure if it has any effect.

pferrel:
Still would like an eye on BiMap and BiDictionary.

The remaining question here is if it solves https://issues.apache.org/jira/browse/SPARK-6069 on a cluster.

I'll concentrate on finding a way to test this. A user tried this on a cluster and got an infinite recursion so wouldn't want to push yet.

pferrel:
OK, fixed the infinite recursion during deserialization. Very subtle fix but works. Asked the user to test again but since I reproduced I think it's fine.

pferrel:
Yikes, @dlyubimov can you look at the serialization control for BiMap here: https://github.com/apache/mahout/pull/125/files#diff-a2665c3f542bcd58a8ae174eb99676e0R36

The idea is to have the inverse not serialized (@transient lazy) so deserialization doesn't recurse, which was a problem before this change. Then I have ```@transient val size_ = inverse.size```, which is never used but forces the inverse to be instantiated inside the  deserialization during construction instead of at the first use of inverse in a task. Not sure this last part is needed but might make a single task per node slow.

I also register the class being serialized (BiDictionary) with Kryo so its serialized with kryo. 

Does this sound ok?

pferrel:
Guava cannot be removed from the dependency-reduced assembly because Preconditions are used quite a bit in mahout-math and mahout-hdfs and therefore may be executed by math-scala and spark. 

Removing Preconditions should be fairly easy but not in the scope of this PR.

pferrel:
If anyone can test on a pseudo-cluster.

    cd mahout/
    hadoop fs -put examples/src/main/resources/cf-* ."
    mahout spark-itemsimilarity -i cf-data-purchase.txt -i2 cf-data-view.txt -o item-sim-out -ma spark://your-spark-master:7077

Then check hdfs ```item-sim-out```

spark/src/main/assembly/dependency-reduced.xml
```diff
@@ -36,7 +36,11 @@
       <outputDirectory>/</outputDirectory>
       <useTransitiveFiltering>true</useTransitiveFiltering>
       <includes>
+<<<<<<< HEAD
```
andrewpalumbo: there's a conflict here.

andrewpalumbo:
hey @pferrel, i ran the test on HDFS in pseudo cluster mode and did get the null pointer  exception:
 ```
{...}
$mahout spark-itemsimilarity -i /cf-data-purchase.txt -i2 /cf-data-view.txt -o /item-sim-out -ma $MASTER -sem 4g

15/05/03 21:15:34 WARN TaskSetManager: Lost task 0.0 in stage 34.0 (TID 46, 192.168.200.135): java.lang.NullPointerException
        at org.apache.mahout.math.indexeddataset.BiMap.inverse$lzycompute(BiMap.scala:37)
        at org.apache.mahout.math.indexeddataset.BiMap.inverse(BiMap.scala:37)
        at org.apache.mahout.drivers.TDIndexedDatasetWriter$$anonfun$writer$3.apply(TextDelimitedReaderWriter.scala:293)
        at org.apache.mahout.drivers.TDIndexedDatasetWriter$$anonfun$writer$3.apply(TextDelimitedReaderWriter.scala:271)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
        at scala.collection.Iterator$$anon$11.next(Iterator.scala:328)
        at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1064)
        at org.apache.spark.rdd.PairRDDFunctions$$anonfun$13.apply(PairRDDFunctions.scala:1051)
        at org.apache.spark.scheduler.ResultTask.runTask(ResultTask.scala:61)
        at org.apache.spark.scheduler.Task.run(Task.scala:56)
        at org.apache.spark.executor.Executor$TaskRunner.run(Executor.scala:200)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1145)
        at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:615)
{...}
```

File: spark/src/main/scala/org/apache/mahout/sparkbindings/io/MahoutKryoRegistrator.scala [hunk 6]
```diff
@@ -35,7 +35,5 @@ class MahoutKryoRegistrator extends KryoRegistrator {
     kryo.addDefaultSerializer(classOf[Vector], new WritableKryoSerializer[Vector, VectorWritable])
     kryo.addDefaultSerializer(classOf[DenseVector], new WritableKryoSerializer[Vector, VectorWritable])
     kryo.addDefaultSerializer(classOf[Matrix], new WritableKryoSerializer[Matrix, MatrixWritable])
-    kryo.register(classOf[com.google.common.collect.HashBiMap[String, Int]], new JavaSerializer())
-
   }
```
pferrel: Took out BiDictionary from Kryo per @dlyubimov Not completely sure why this is bad except *any* mods to Spark/Kryo setup can bite you later. This means that serialization of the dictionaries is not as fast as it could be but make Mahout easier to integrate with applications as a library, though not much because you still have to make sure this method get called before a context is created.

pferrel:
This branch seems to work on a pseudo-cluster and even works on Spark 1.3 though it's not in the POM yet.

Note that the Spark version in the POM is 1.2.2, the latest in that line.

pferrel:
pushing to apache/master as soon as build an simple smoke test pass to speak now.

andrewpalumbo:
@pferrel Is there anything to be done here?

PR #126, opened 2015-05-01
Spark 1.2

pferrel:
Makes cooccurrence work on Spark 1.2 (and 1.3) by removing the use of Javaserializer on Guava. Implements a new BiDictionary based on a BiMap, based on two Scala Maps. Only one map is serialized in broadcast.


pferrel:
moved to mahout-0.10.x branch

pferrel:
not sure what branch to target for this but it should go into the next release if it smoke tests.

spark/src/main/scala/org/apache/mahout/sparkbindings/io/MahoutKryoRegistrator.scala
```diff
@@ -35,7 +35,6 @@ class MahoutKryoRegistrator extends KryoRegistrator {
     kryo.addDefaultSerializer(classOf[Vector], new WritableKryoSerializer[Vector, VectorWritable])
     kryo.addDefaultSerializer(classOf[DenseVector], new WritableKryoSerializer[Vector, VectorWritable])
     kryo.addDefaultSerializer(classOf[Matrix], new WritableKryoSerializer[Matrix, MatrixWritable])
-    kryo.register(classOf[com.google.common.collect.HashBiMap[String, Int]], new JavaSerializer())
-
+    kryo.register(classOf[BiDictionary])
```
dlyubimov: -1 So i told before i don't need to see a need for this. These structures only exist in the memory. HashMap, TreeMap, (Bi)Dictionary etc.etc. on the wire they all serialize as Seq[(K,V)]. 

So by that logic if i start using tree types in my mappers, i need to add a specific serialization support for it. No, i don't. Just because i want to build a treeSet in memory, doesn't mean i need to add extra support for it in serialization. 

I have a pretty strong opinion that we must not have any specific type in kryo serialization but matrices and vectors (because they use less than trivial logic there, they are not sequences of stuff). But any collection type of standard stuff should be happily serializable as a standard wire-friendly collection (i.e. sequence of standard stuff). 

The only tricky thing might be that we don't want to materialize sequences of AnyVal stuff before rebuilding them into indexed collections; so this latter issue may need a bit of experimenting, but this is a technical issue, not architectural.

pferrel: Ok, this works without any kryo.register. I've done it both ways. 

Note that there is no separate serializer class, the class extends Serializable but doesn't implement anything special and so should just use the Seq[(K,V)] for one map. The inverse is calculated after the broadcast deserialization so it is never on the wire.

If the rest looks ok I'll remove this line but still not sure why it is bad. Spark recommends using kryo by registering classes. Doesn't removing this mean using default serialization instead of Kryo?

pferrel:
pushing to master so closing this, see PR #125

PR #127, opened 2015-05-01
Spark 1.2

pferrel:
Makes cooccurrence work on Spark 1.2 (and 1.3) by removing the use of Javaserializer on Guava. Implements a new BiDictionary based on a BiMap, based on two Scala Maps. Only one map is serialized in broadcast.

moved from pushing to master to pushing to mahout-0.10.x branch


pferrel:
this has too much history, going to remerge

PR #128, opened 2015-05-02
MAHOUT-1704: Pare down dependency jar for h2o

andrewpalumbo:
This takes the dependency jar for h2o down to ~17M from~68M. Need to verify if this contains all needed deps. I have successfully tested launching {{$mahout h2o-node}} and run $mvn test in parallel on 2 nodes.

### DIFF
--- a/bin/mahout
+++ b/bin/mahout
@@ -186,9 +186,10 @@ then
        CLASSPATH=${CLASSPATH}:$f;
     done
 
-    for f in $MAHOUT_HOME/h2o/target/mahout-h2o-*.jar; do
+    for f in $MAHOUT_HOME/h2o/target/mahout-h2o*.jar; do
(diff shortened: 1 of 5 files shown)


andrewpalumbo:
@avati, could you take a look at this?  I think that we will have to supply other mahout classes to `h2o-node` in `bin\mahout` or elsewhere.

andrewpalumbo:
Added scala libraries into dependency jar.  This bumps the size up to 20M. H2O Tests are passing in parallel now.

avati:
Will look into this soon

andrewpalumbo:
thx.  Should be good. removed scala-reflect. this brings the dependency jar down to ~17M. Tests are passing in parallel.

andrewpalumbo:
this tests fine in parallel.  I'm going to clean it up and push it this afternoon if there are no objections.

PR #130, opened 2015-05-08
Readme update

aaronbenz:
Updating the README.txt to a .md. Also incorporating some other useful information from the Mahout website to make it a little more pretty

I'm not sure if the artifacts are correct however as I found two different uses from https://mahout.apache.org/users/basics/quickstart.html and https://mahout.apache.org/general/downloads.html


dlyubimov:
.md conversion is welcome. but the content  i think is a bit outdated

PR #131, opened 2015-05-13
Fix bug: Add commit after executeUpdate

sugaE:
It will not write database without commit().


andrewmusselman:
@sugaE this may be reasonable, so even though we're not maintaining the old map-reduce-based code would you be able to add a test to confirm this is an improvement, showing the behavior with and without the call to commit?

PR #133, opened 2015-05-29
MAHOUT-1706-0.10.x

andrewpalumbo:
The mahout distribution currently is shipping ~56 MB of dependency jars in the /lib directory of the distribution. These are only added to the classpath by /bin/mahout in the binary distribution. It seems that we can remove them from the distribution. (we need to get the size of the distribution down).

This branch still needs testing on an cluster.

### DIFF
--- a/bin/mahout
+++ b/bin/mahout
@@ -234,13 +234,6 @@ else
   CLASSPATH=${CLASSPATH}:$MAHOUT_HOME/h2o/target/classes
 fi
 
-# add development dependencies to CLASSPATH
-if [ "$SPARK" != "1" ]; then
-  for f in $MAHOUT_HOME/examples/target/dependency/*.jar; do
-    CLASSPATH=${CLASSPATH}:$f;
-  done
-fi
-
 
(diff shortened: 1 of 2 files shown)


dlyubimov:
noted... 
this does or does not affect shell classpath?

andrewpalumbo:
It should not (negatively) affect the shell classpath, since all of the necessary dependencies are either packaged with {{mahout-spark-*.jars}} or in the classpath returned by {{${SPARK_HOME}/bin/compute-classpath.sh}}.  

This patch actually does not affect the classpath of a source build at all.  It only prevents the release process from packaging dependency jars into the {{/lib}} dir of the binary distribution artifact.

However since it does add all of the jars in the {{/lib}} directory of the binary build for any mahout command line flavor (spark, MR, h2o), I have been hesitant to push it without more extensive testing (cluster and local).

I am going to close this for now and investigate a better approach.

andrewpalumbo:
closing without pushing while looking at different approaches.

PR #134, opened 2015-06-05
MAHOUT-1642 fix missing first element in SimilarItemsIterator

fape:
based on @OlegZotov's fix #77

### DIFF
--- a/mr/src/test/java/org/apache/mahout/cf/taste/similarity/precompute/SimilarItemsTest.java
+++ b/mr/src/test/java/org/apache/mahout/cf/taste/similarity/precompute/SimilarItemsTest.java
@@ -1,0 +1,51 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one or more
+ * contributor license agreements.  See the NOTICE file distributed with
+ * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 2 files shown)


dlyubimov:
ok, without being in context of this algorithmically, this looks like a valid fix to me. I think I'd rely on the opinion on contributors and commit it even though it is to a taste recommender. If @srowen happens to come and say yay or nay, it'd be great.

mr/src/test/java/org/apache/mahout/cf/taste/similarity/precompute/SimilarItemsTest.java
```diff
@@ -0,0 +1,51 @@
...
+import org.apache.mahout.cf.taste.recommender.RecommendedItem;
+import org.hamcrest.Matchers;
+import org.junit.Test;
+
+public class SimilarItemsTest extends TasteTestCase {
+
+  @Test
+  public void testIterator() {
+
+	List<RecommendedItem> recommendedItems = new ArrayList<RecommendedItem>();
```
smarthi: change this to be JDK 1.7 compliant, the project's on 1.7 since 0.10.0

PR #135, opened 2015-06-05
Mahout 0.10.x ora 0515 MAHOUT-1660 MAHOUT-1713 MAHOUT-1714 MAHOUT-1715 MAHOUT-1716 MAHOUT-1717 MAHOUT-1718 MAHOUT-1719 MAHOUT-1720 MAHOUT-1721 MAHOUT-1722 MAHOUT-1723 MAHOUT-1724 MAHOUT-1725 MAHOUT-1726 MAHOUT-1727 MAHOUT-1728 MAHOUT-1729 MAHOUT-1730  MAHOUT-1731 MAHOUT-1732 

dlyubimov:
Cumulative patch for the issues above.


h2o/src/main/scala/org/apache/mahout/h2obindings/H2OEngine.scala
```diff
@@ -26,9 +26,13 @@ import org.apache.mahout.math.drm.logical._
 import org.apache.mahout.h2obindings.ops._
 import org.apache.mahout.h2obindings.drm._
 import org.apache.mahout.h2o.common.{Hadoop1HDFSUtil, HDFSUtil}
+import org.apache.mahout.logging._
 
 /** H2O specific non-DRM operations */
 object H2OEngine extends DistributedEngine {
+
+  private implicit val log = getLog(H2OEngine.getClass)
```
smarthi: See an 'final implicit' further down in a different file, should this be 'final implicit' too ?

dlyubimov: yes it should, in fact.  thanks. although it is more about private than final; private values cannot be neighter changed nor overriden anyway. I am a bit surprised Scala allows "private final" in the first place.

dlyubimov: taken care of

dlyubimov:
@avati @Anand Avati: support for MAHOUT-1717 needed for H20 that comes with this PR #135 (allreduceBlock)

dlyubimov:
tracked as a part of this #135 : (for asfgit to leave a reference in these issues):
MAHOUT-1660 MAHOUT-1713 MAHOUT-1714 MAHOUT-1715 MAHOUT-1716 MAHOUT-1717 MAHOUT-1718 MAHOUT-1719 MAHOUT-1720 MAHOUT-1721 MAHOUT-1722 MAHOUT-1723 MAHOUT-1724 MAHOUT-1725 MAHOUT-1726 MAHOUT-1727 MAHOUT-1728 MAHOUT-1729 MAHOUT-1730 MAHOUT-1731 MAHOUT-1732

dlyubimov:
MAHOUT-1713 is part of #135

dlyubimov:
MAHOUT-1714 is part of #135

dlyubimov:
MAHOUT-1715 is part of #135

dlyubimov:
MAHOUT-1716 is part of #135

avati:
@dlyubimov Yes, I will have a look into it. How do you prefer:

1. You merge this by leaving ??? in H2O, and I raise a new PR after that replacing ??? with actual code

2. I work against this development branch, you merge my code and re-submit this PR that includes H2O changes

dlyubimov:
MAHOUT-1717 is part of #135

dlyubimov:
MAHOUT-1718 is part of #135

dlyubimov:
MAHOUT-1719 is part of #135

dlyubimov:
MAHOUT-1720 is part of #135

dlyubimov:
MAHOUT-1721 is part of #135

dlyubimov:
MAHOUT-1722 is part of #135

dlyubimov:
MAHOUT-1723 is part of #135

dlyubimov:
MAHOUT-1724 is part of #135

dlyubimov:
MAHOUT-1725 is part of #135

dlyubimov:
MAHOUT-1726 is part of #135

dlyubimov:
MAHOUT-1727 is part of #135

dlyubimov:
MAHOUT-1728 is part of #135

dlyubimov:
MAHOUT-1729 is part of #135

dlyubimov:
MAHOUT-1730 is part of #135

dlyubimov:
MAHOUT-1731 is part of #135

dlyubimov:
MAHOUT-1732 is part of #135

dlyubimov:
@avati -- you can start on this PR branch if you prefer so however, #135 will go in on Wednesday. at that point the history of this issue will be squashed; however, i don't imagine it should create issues for you to merge in the 0.10.x branch later, since squashed commit will be equivalent to the history prefix you'd have.

dlyubimov:
ok so this will break h20 tests.

for one, it replaces physical operator of AewScalar with AewUnaryFunc (elementwise application of a function).

avati:
Ah yes, that will break existing tests. I think we need a workflow strategy for such changes in the future, especially as more engines are added.

dlyubimov:
yeah.. although this might be tough, it doesn't look terribly likely that we want to introduce a lot. We had a conversation about it with @pferrel who added a bunch of ??? stuff to H20Engine which do not seem terribly relevant to algebraic stuff (to me). We agreed that if indexed format is needed, perhaps this is something to solve ouside of individual backend scope. Either way, something pending as ??? forever is not the effect we want -- and i suggest to apply a higher level of scrutiny on what is going on there. 


in my case, motivation of adding (replacing) dedicated physical operator computing elementwise scalar operation with elementwise function operator is pretty clear. it allows to detect and optimize and fuse more general elementwise cases on upper level of optimizations, such as for example 1 + drmP * dlog(drmP), into one physical map. Also it detects some very important and frequent enough  cases like X * X (self-hadamard) and sends it into this unary elementwise function instead of AewB.

But in case of indexed dictionary structure i am not sure i have heard  convincing arguments this needs to be there... at least in the algebraic engine itself... i am perfectly fine with quasi-algebraic support with what @pferrel  is doing. I can quite imagine this will not be an outlier case...

dlyubimov:
@avati ok i created two issues, MAHOUT-1736 to track allreaduceBlock() and MAHOUT-1737 to track AewUnaryFunc physical operator in H20. 

there are a couple of samplers there as well that would be nice to support at least with approximate methods similar to what was applied in spark but if they are not supported oh well. 

But the former two are much more essential for portability support.

dlyubimov:
closing manuall after commit. @asfgit seems to fail to auto-close requests that are not on master. bummer.

h2o/src/main/java/org/apache/mahout/h2obindings/drm/H2OBCast.java
```diff
@@ -118,4 +118,16 @@ private T deserialize(byte buf[]) {
...
   }
+
+  /**
+   * Stop broadcasting when called on driver side. Release any network resources.
+   *
+   */
+  @Override
+  public void close() throws IOException {
+
+    // TODO: review this. It looks like it is not really a broadcast mechanism but rather just a
```
andrewmusselman: Empty method?

dlyubimov: I am not sure how h20 does it but it looks like broadcast is just serialized to the backend with all requests, not really broadcasted. if so, there are no associated resources so method needs to do nothing. Anyhow it's something for Anand @avati  to look at.

math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
```diff
@@ -43,18 +46,22 @@ object DSSVD {
       case (keys, blockA) =>
         val blockY = blockA %*% Matrices.symmetricUniformView(n, r, omegaSeed)
         keys -> blockY
-    }
+    }.checkpoint()
```
andrewmusselman: This puts results into a cache?

math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
```diff
@@ -62,13 +69,17 @@ object DSSVD {
       if (i == q - 1) drmQ = drmQ.checkpoint()
 
       // This on the other hand should be inner-join-and-map A'B optimization since A and Q_i are not
-      // identically partitioned anymore.
+      // identically partitioned anymore.`
```
andrewmusselman: Trailing backtick

dlyubimov: Noted for round of style fixes.

math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
andrewmusselman: Oh I see this is just moved out of line 51

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
```diff
@@ -73,20 +75,39 @@ trait DistributedEngine {
...
+   * (Optional) Sampling operation. Consistent with Spark semantics of the same.
+   * @param drmX
+   * @param fraction
+   * @param replacement
+   * @tparam K
+   * @return
+   */
+  def drmSampleRows[K: ClassTag](drmX: DrmLike[K], fraction: Double, replacement: Boolean = false): DrmLike[K]
+
+  def drmSampleKRows[K: ClassTag](drmX: DrmLike[K], numSamples:Int, replacement:Boolean = false) : Matrix
```
andrewmusselman: Why does this return a Matrix whereas the previous one returns DrmLike[K], and is there a default number of samples in the previous one?

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
```diff
@@ -119,38 +140,49 @@ object DistributedEngine {
   private def pass1[K: ClassTag](action: DrmLike[K]): DrmLike[K] = {
 
     action match {
-      case OpAB(OpAt(a), b) if (a == b) => OpAtA(pass1(a))
-      case OpABAnyKey(OpAtAnyKey(a), b) if (a == b) => OpAtA(pass1(a))
+
+      // self element-wise rewrite
+      case OpAewB(a, b, op) if (a == b) => {
```
andrewmusselman: => instead of ⇒

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
```diff
@@ -160,17 +192,30 @@ object DistributedEngine {
   /** This would remove stuff like A.t.t that previous step may have created */
   private def pass2[K: ClassTag](action: DrmLike[K]): DrmLike[K] = {
     action match {
+
+      // Fusion of unary funcs into single, like 1 + x * x.
+      // Since we repeating the pass over self after rewrite, we dont' need to descend into arguments
```
andrewmusselman: Apostrophe here and line 203

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
```diff
@@ -182,29 +227,29 @@ object DistributedEngine {
     action match {
 
       // matrix products.
-      case OpAB(a, OpAt(b)) => OpABt(pass3(a), pass3(b))
+      case OpAB(a, OpAt(b)) ⇒  OpABt(pass3(a), pass3(b))
```
andrewmusselman: Extra whitespace after several => operators

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
dlyubimov: I am using the unicode chars now (change of style that happened to me a few months ago) but it has not gone over the entire codebase. We probably may want to standardize on this (later) but for now i just keep style consistent with the rest of the file. If i change the style, i change the style in the whole file.

math-scala/src/main/scala/org/apache/mahout/math/drm/DistributedEngine.scala
dlyubimov: I needed approximate uniform hypergeometric sampler quick. I have distributed version as well but it is not the code i am allowed to commit. Spark only supports k-sampler that can fit in memory, unfortunately. I thought of keeping it uniform but it wouldn't make much sense since implementation would have to re-distribute it again and the way i used it i collect it back again so it would be double-trip for nothing but just keeping the API uniform... we can discuss it in a separate issue if you want.

math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
```diff
@@ -115,6 +121,46 @@ package object drm {
...
+   * row matrices. So this method is to help.
+   *
+   * @param drmX input to be transcoded
+   * @param computeMap collect `old key -> int key` map to front-end?
+   * @tparam K key type
+   * @return Sequentially keyed matrix + (optionally) map from non-int key to [[Int]] key. If the
+   *         key type is actually Int, then we just return the argument with None for the map,
+   *         regardless of computeMap parameter.
+   */
+  def drm2IntKeyed[K: ClassTag](drmX: DrmLike[K], computeMap: Boolean = false): (DrmLike[Int], Option[DrmLike[K]]) =
```
andrewmusselman: This looks like a pass-through to DistributedEngine; am I reading it right or missing something?

math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
```diff
@@ -115,6 +121,46 @@ package object drm {
...
+
+  /**
+   * (Optional) Sampling operation. Consistent with Spark semantics of the same.
+   * @param drmX
+   * @param fraction
+   * @param replacement
+   * @tparam K
+   * @return samples
+   */
+  def drmSampleRows[K: ClassTag](drmX: DrmLike[K], fraction: Double, replacement: Boolean = false): DrmLike[K] =
```
andrewmusselman: Same with these; I don't understand the package.scala construct though.

h2o/src/main/java/org/apache/mahout/h2obindings/drm/H2OBCast.java
avati: Yes, h2o's broadcast is just serialize + multiple unicast (almost like spark's http based broadcast)

spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -78,8 +78,8 @@ class CheckpointedDrmSpark[K: ClassTag](
   }
 
   def cache() = {
-    if (!cached) {
-      rdd.persist(_cacheStorageLevel)
+    if (!cached && _cacheStorageLevel != StorageLevel.NONE) {
```
andrewmusselman: Will this be confusing if someone tries to call cache() with NONE set and it's not cached?

dlyubimov: NONE means yes, it's not cached.

The material effect of setting "NONE" cache level in spark, as far as i
understand, is exactly as not setting anything at all; the difference is
once you set it to anything, you cannot change it. So this means if i call
"NONE" it would lock me to that choice with spark (which did not make much
sense to me) otherwise it doesn't matter.

math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
dlyubimov: Any public api, if it is not a matrix method,  is package-level set of functions. This is to follow R conventions where most of the functions are not obviously connected to any object. I.e. we just write something like 

import o.a.m.math.drm._
....

val sample = drmSampleRows(...)

dlyubimov: also, i think engine variable is not visible to user, only to samsara functions. so there is no 2 ways for user to invoke that.

spark/src/main/scala/org/apache/mahout/sparkbindings/blas/Par.scala
```diff
@@ -1,50 +1,58 @@
...
-    if (src.isBlockified) {
-      val rdd = src.toBlockifiedDrmRdd()
-      new DrmRddInput[K](blockifiedSrc = Some(adjust(rdd)))
+    val srcBlockified = src.isBlockified
+
+    val srcRdd = if (srcBlockified) src.toBlockifiedDrmRdd(op.ncol) else src.toDrmRdd()
+    val srcNParts = srcRdd.partitions.size
+
+    // To what size?
+    val targetParts = if (op.minSplits > 0) srcNParts max op.minSplits
```
andrewmusselman: This is a bit tough to read but could be I'm just not accustomed to the idiom

dlyubimov: this is scala style yes. they decided to do away with inline ?: (bincond)

spark/src/main/scala/org/apache/mahout/sparkbindings/blas/AtB.scala
```diff
@@ -27,92 +32,330 @@ import org.apache.spark.SparkContext._
...
    * zip all the rows. Otherwise, we need to inner-join them first.
+   *
    */
-  def atb_nograph[A: ClassTag](
-      operator: OpAtB[A],
-      srcA: DrmRddInput[A],
-      srcB: DrmRddInput[A],
-      zippable:Boolean = false
-      ): DrmRddInput[Int] = {
+  @deprecated("slow, will remove", since = "0.10.2")
```
andrewmusselman: Do we have a plan to improve/remove?

math/src/main/java/org/apache/mahout/math/VectorView.java
```diff
@@ -122,7 +127,7 @@ protected Element computeNext() {
       while (it.hasNext()) {
         Element el = it.next();
         if (isInView(el.index()) && el.get() != 0) {
-          Element decorated = vector.getElement(el.index());
+          Element decorated = el; /* vector.getElement(el.index()); */
```
andrewmusselman: Remove comment

spark/src/main/scala/org/apache/mahout/sparkbindings/blas/AtB.scala
dlyubimov: it's replaced. remove is the plan. experiments show it's decisively losing
in current spark implementations. May change though with future spark
versions. one never can be sure.

andrewmusselman:
Looks great, amazing drop, and very readable even for a Scala novice like me.

PR #136, opened 2015-06-08
MAHOUT-1653 Spark-1.3 for 0.11.0

pferrel:
Added a few comments to Andy's Spark 1.3 branch need to replace the de…pricated sequence file write code still.

### DIFF
--- a/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
+++ b/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
@@ -28,7 +28,7 @@ import scala.collection.JavaConversions._
 import org.apache.spark.storage.StorageLevel
 import reflect._
 import scala.util.Random
-import org.apache.hadoop.io.{LongWritable, Text, IntWritable, Writable}
+import org.apache.hadoop.io._
 import org.apache.mahout.math.drm._
 import org.apache.mahout.sparkbindings._
 import org.apache.spark.SparkContext._
@@ -159,7 +159,7 @@ class CheckpointedDrmSpark[K: ClassTag](
    */
   def dfsWrite(path: String) = {
     val ktag = implicitly[ClassTag[K]]
-    val vtag = implicitly[ClassTag[Vector]]
+    //val vtag = implicitly[ClassTag[Vector]]
 
     implicit val k2wFunc: (K) => Writable =
       if (ktag.runtimeClass == classOf[Int]) (x: K) => new IntWritable(x.asInstanceOf[Int])
@@ -171,12 +171,11 @@ class CheckpointedDrmSpark[K: ClassTag](
     // rdd.saveAsSequenceFile(path)
     // this is a (working) deprecated method used as a stop-gap while we investigate the shell issues
     SparkContext.rddToSequenceFileRDDFunctions(rdd.asInstanceOf[RDD[(K, Vector)]]).saveAsSequenceFile(path)
-    // parameters to the non-depricated function are
-    // SequenceFileRDDFunctions(
-    //   self: RDD[(K, V)], // rdd.asInstanceOf[RDD[(K, Vector)] is ok here
-    //   _keyWritableClass: Class[_ <: Writable], // k2wFunc is ok here
-    //   _valueWritableClass: Class[_ <: Writable] // we need to have a factory for vector here, VectorWritable isn't
-    //                                             // acceptible as-is I think.
+    // discussion here: http://search-lucene.com/m/rcu7o1J1Z7mvnGmP1
+    // The correct function to use is here: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L1671
+    // The commit message for this change is here: https://github.com/apache/spark/tree/master/core/src/test/scala/org/apache/sparktest
+    // The key and value factories will be determined using reflection on the RDD if the right classes are set to null
+    // see this example: https://github.com/apache/spark/blob/branch-1.4/core/src/main/scala/org/apache/spark/SparkContext.scala#L2236
   }
 
   protected def computeNRow = {
--- a/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
+++ b/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
@@ -18,6 +18,8 @@
 package org.apache.mahout.sparkbindings.drm
 
 import org.apache.mahout.math._
+import org.apache.spark.SparkContext
+import org.apache.spark.rdd.{SequenceFileRDDFunctions, RDD}
 import math._
 import scalabindings._
 import RLikeOps._
@@ -157,6 +159,7 @@ class CheckpointedDrmSpark[K: ClassTag](
    */
   def dfsWrite(path: String) = {
     val ktag = implicitly[ClassTag[K]]
+    val vtag = implicitly[ClassTag[Vector]]
 
     implicit val k2wFunc: (K) => Writable =
       if (ktag.runtimeClass == classOf[Int]) (x: K) => new IntWritable(x.asInstanceOf[Int])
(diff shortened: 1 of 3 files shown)


spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -165,7 +168,15 @@ class CheckpointedDrmSpark[K: ClassTag](
...
+    // rdd.saveAsSequenceFile(path)
+    // this is a (working) deprecated method used as a stop-gap while we investigate the shell issues
+    SparkContext.rddToSequenceFileRDDFunctions(rdd.asInstanceOf[RDD[(K, Vector)]]).saveAsSequenceFile(path)
+    // parameters to the non-depricated function are
+    // SequenceFileRDDFunctions(
+    //   self: RDD[(K, V)], // rdd.asInstanceOf[RDD[(K, Vector)] is ok here
+    //   _keyWritableClass: Class[_ <: Writable], // k2wFunc is ok here
+    //   _valueWritableClass: Class[_ <: Writable] // we need to have a factory for vector here, VectorWritable isn't
+    //                                             // acceptible as-is I think.
   }
```
pferrel: minimally dfsWrite should probably use a non-deprecated write function. This requires a factory for the vector used in drms, is this VectorWritable?

spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -157,6 +159,7 @@ class CheckpointedDrmSpark[K: ClassTag](
    */
   def dfsWrite(path: String) = {
     val ktag = implicitly[ClassTag[K]]
+    val vtag = implicitly[ClassTag[Vector]]
```
andrewpalumbo: I think this line can come out for now.

spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -165,7 +168,14 @@ class CheckpointedDrmSpark[K: ClassTag](
...
-    rdd.saveAsSequenceFile(path)
+    // rdd.saveAsSequenceFile(path)
+    // this is a (working) deprecated method used as a stop-gap while we investigate the shell issues
+    SparkContext.rddToSequenceFileRDDFunctions(rdd.asInstanceOf[RDD[(K, Vector)]]).saveAsSequenceFile(path)
+    // discussion here: http://search-lucene.com/m/rcu7o1J1Z7mvnGmP1
+    // The correct function to use is here: https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/RDD.scala#L1671
+    // The commit message for this change is here: https://github.com/apache/spark/tree/master/core/src/test/scala/org/apache/sparktest
+    // The key and value factories will be determined using reflection on the RDD if the right classes are set to null
+    // see this example: https://github.com/apache/spark/blob/branch-1.4/core/src/main/scala/org/apache/spark/SparkContext.scala#L2236
   }
```
pferrel: There seems to be a way to use reflection to get the Writable factory from the RDD by setting it to null. not sure exactly how to do that here.

dlyubimov: I am sorry. but this is wrong.

andrewpalumbo: I think that it should be something like:
```
implicit val keyWritableFactory = spark.WritableFactory.writableWritableFactory[IntWritable]
implicit val valueWritableFactory = spark.WritableFactory.writableWritableFactory[VectorWritable] 
```
For an Int-keyed Drm. Similarly with String-keyed, and Long-keyed.

spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala
```diff
@@ -165,7 +168,14 @@ class CheckpointedDrmSpark[K: ClassTag](
       else if (classOf[Writable].isAssignableFrom(ktag.runtimeClass)) (x: K) => x.asInstanceOf[Writable]
       else throw new IllegalArgumentException("Do not know how to convert class tag %s to Writable.".format(ktag))
 
-    rdd.saveAsSequenceFile(path)
```
andrewpalumbo: It seems that the solution could be to map the DrmRdd keys and values to their respective Writables before calling `.saveAsSequenceFile(path)`.    We can use the k2wFunc for the Key, and the value will always be a Vector for Drms so simply map to a VectorWritable.     This is essentially what is done internally in `.saveAsSequenceFile` for standard Writables.
```
    // convert RDD to Writables
    rddInput.toDrmRdd().map(x => (k2wFunc(x._1), v2w(x._2)) )
                       .saveAsSequenceFile(path)
```

However the above gives a runtime exception since the Writables are not Java Serializable: ```org.apache.spark.SparkException: Task not serializable```.  Any thoughts on how to do this correctly?

andrewpalumbo: Ok, so we can eliminate the runtime error by mapping each key/value to the appropriate `Writable` (not from within a function):

```
if (ktag.runtimeClass == classOf[Int]) {
      rddInput.toDrmRdd()

        .map( x =>(new IntWritable(x._1.asInstanceOf[Int]), new VectorWritable(x._2))).saveAsSequenceFile(path)

    } else if (ktag.runtimeClass == classOf[String]){
      rddInput.toDrmRdd()

        .map( x =>(new Text(x._1.asInstanceOf[String]), new VectorWritable(x._2))).saveAsSequenceFile(path)

    } else if (ktag.runtimeClass == classOf[Long]) {
      rddInput.toDrmRdd()

        .map( x =>(new LongWritable(x._1.asInstanceOf[Long]), new VectorWritable(x._2))).saveAsSequenceFile(path)

    } else throw new IllegalArgumentException("Do not know how to convert class tag %s to Writable.".format(ktag))```

pferrel: So are you suggesting that we deprecate dfsWrite? Meaning leave it as is with a deprecated Spark function for now and replace it wherever it's called? Then when Spark removes the deprecated function we remove dfsWrite?

Can we handle your code with an implicit conversion that will get called before dfsWrite?

andrewpalumbo: That is actually using the non-deprecated `.saveAsSequenceFile(path)`  I'm just suggesting that we could skip all of the implicit conversions and we explicitly map the RDD to Writables ourselves.  Then call `.saveAsSequenceFile(path)` on the RDD of eg. `[IntWritable, VectorWritable]`. This is actually what Spark does in `.saveAsSequenceFile(path)` : https://github.com/apache/spark/blob/master/core/src/main/scala/org/apache/spark/rdd/SequenceFileRDDFunctions.scala#L97

if either a Key or a Value is not a `Writable`, it converts one or the other or both to a Writable using eg.: ```self.map(x => (anyToWritable(x._1), anyToWritable(x._2)))```

and then calls `.saveAsHadoopFile(...)` on the Mapped RDD.  

If it detects that both are Writables though as would be the case if we mapped them explicitly, it simply calls `.saveAsHadoopFile(...)`.  So By mapping them ourselves in `.dfsWrite(...)` we shouldn't incur any additional overhead.

Actually we may just be able to call `.saveAsHadoopFile(...)` directly on a mapped => Writable RDD from `.dfsWrite(...)`.

andrewpalumbo: implemented here (passes all tests):

https://github.com/andrewpalumbo/mahout/blob/MAHOUT-1653-jun/spark/src/main/scala/org/apache/mahout/sparkbindings/drm/CheckpointedDrmSpark.scala#L157

pferrel: Ok, need to rebase against your branch. 

If I can get this to build against the current master/0.11.0-SNAPSHOT I'd be inclined to push it so someone speak up if there is an objection. Once again, it will disable the shell build, for now.

andrewpalumbo: +1 (no objection that is..)

andrewpalumbo:
oh man.. @pferrel I didn't realize that i based that branch off of 0.10.x. I just did it with another one.  I'll have to fix them up.

andrewpalumbo:
https://github.com/apache/mahout/compare/master...andrewpalumbo:MAHOUT-1653-master?expand=1
rebased against master- This branch should also be compatible with < Spark 1.3 (though not has not been tested).  I can push it as part of #146 when that's ready. or if you need it ahead of that you can push now or whenever.

pferrel:
Andy pushed his 1.3 branch to master so no longer needed

PR #137, opened 2015-06-16
MAHOUT-1570: Flink backend for Mahout DSL

alexeygrigorev:
Implementation for MAHOUT-1570: Adding support for Apache Flink as a backend for the Mahout DSL


flink/src/main/scala/org/apache/mahout/flinkbindings/blas/FlinkOpAewScalar.scala
```diff
@@ -0,0 +1,105 @@
+/**
```
dlyubimov: A elementwise scalar is deprecated now. Better use unaryfunc operator. Then things like A+5 can be represented as 

OpUnaryFunc(A, x=> x+5). 

Unary function mechanism is more generic and affords for unary function fusion optimizations.

dlyubimov:
also on the topic of test suite coverage: we need to pass our standard tests. The base clases for them are:

https://github.com/apache/mahout/blob/master/math-scala/src/test/scala/org/apache/mahout/math/decompositions/DistributedDecompositionsSuiteBase.scala

https://github.com/apache/mahout/blob/master/math-scala/src/test/scala/org/apache/mahout/math/drm/DrmLikeOpsSuiteBase.scala

https://github.com/apache/mahout/blob/master/math-scala/src/test/scala/org/apache/mahout/math/drm/DrmLikeSuiteBase.scala

https://github.com/apache/mahout/blob/master/math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala

The technique here is to take these test cases as a base class for a distributed test case (you may want to see how it was done for Spark and H2O). This is our basic assertion that our main algorithms are passing on a toy problem for a given backend.

flink/pom.xml
```diff
@@ -0,0 +1,187 @@
...
+-->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.apache.mahout</groupId>
+    <artifactId>mahout</artifactId>
+    <version>0.11.0-SNAPSHOT</version>
```
smarthi: Update this to 0.11.1-SNAPSHOT, since we will be releasing 0.11.0 this week

alexeygrigorev:
I rebased and cleaned the commit history a bit to make sure every commit has a jira id. The documentation and the current status can be found at https://github.com/alexeygrigorev/mahout/wiki/Samsara-Flink-Bindings

math-scala/src/main/scala/org/apache/mahout/math/drm/logical/CheckpointAction.scala
```diff
@@ -44,5 +44,6 @@ abstract class CheckpointAction[K: ClassTag] extends DrmLike[K] {
     case Some(cp) => cp
   }
 
+  val classTag = implicitly[ClassTag[K]]
```
dlyubimov: maybe we should expose that in DrmLike. 
I so far had a need for this in logical operators (AbstractBinaryOp$classTagK, AbstractUnaryOp$classTagK) for optimizer's needs. so maybe it needs to be promoted up if it turns out every implementation must keep implicit classtag inside.

alexeygrigorev: I remember trying to attach it to DrmLike, but I had some problems because it's a trait

dlyubimov: yes you can't put implementation into trait. we'd have to make it optional
for the trait (=???) and override in the concrete things. Concrete things
are of course just of two types: logical operators based on the classes i
already mentioned, that already have this support, and the checkpoint
implementation, which probably does not yet. let me play with your branch a
little to see what we can do.

dlyubimov:
Hm. doesn't compile for me with maven 3.3.3 and jdk 1.8.

INFO] --- maven-assembly-plugin:2.4.1:single (dependency-reduced) @ mahout-flink_2.10 ---
[INFO] Reading assembly descriptor: src/main/assembly/dependency-reduced.xml
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Mahout Build Tools ................................. SUCCESS [  1.782 s]
[INFO] Apache Mahout ...................................... SUCCESS [  0.039 s]
[INFO] Mahout Math ........................................ SUCCESS [  7.477 s]
[INFO] Mahout HDFS ........................................ SUCCESS [  1.487 s]
[INFO] Mahout Map-Reduce .................................. SUCCESS [ 12.867 s]
[INFO] Mahout Integration ................................. SUCCESS [  2.695 s]
[INFO] Mahout Examples .................................... SUCCESS [ 13.358 s]
[INFO] Mahout Math Scala bindings ......................... SUCCESS [ 25.349 s]
[INFO] Mahout H2O backend ................................. SUCCESS [ 16.332 s]
[INFO] Mahout Spark bindings .............................. SUCCESS [ 26.458 s]
[INFO] Mahout Spark bindings shell ........................ SUCCESS [  4.863 s]
[INFO] Mahout Release Package ............................. SUCCESS [  0.663 s]
[INFO] Mahout Flink bindings .............................. FAILURE [06:09 min]
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 08:03 min
[INFO] Finished at: 2015-09-29T13:24:29-07:00
[INFO] Final Memory: 80M/859M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:2.4.1:single (dependency-reduced) on project mahout-flink_2.10: Error reading assemblies: Error locating assembly descriptor: src/main/assembly/dependency-reduced.xml
[ERROR] 
[ERROR] [1] [INFO] Searching for file location: /TB/dmitriy/projects/mahout/flink/src/main/assembly/dependency-reduced.xml
[ERROR] 
[ERROR] [2] [INFO] File: /TB/dmitriy/projects/mahout/flink/src/main/assembly/dependency-reduced.xml does not exist.
[ERROR] 
[ERROR] [3] [INFO] File: /TB/dmitriy/projects/mahout/src/main/assembly/dependency-reduced.xml does not exist.
[ERROR] -> [Help 1]
[E

alexeygrigorev:
I also remember having problems with building Mahout with jdk 1.8. Have you
tried 1.7?
Unfortunately I cannot test it myself at the moment...

alexeygrigorev:
I tried it with 1.7 and had the same problem. I'll look into it

alexeygrigorev:
So I fixed it and I can run `mvn clean package -DskipTests`. But since some of the tests don't pass for the Flink backend, if I remove the `-DskipTests`, I won't build. What do you think, should I disable the failing tests?

smarthi:
I think it's best to disable the failing tests, else it's gonna break the ci build

dlyubimov:
If they are standard algorithm tests coming from the abstract test suites
in math-scala, they currently cannot be disabled (in any way i know anyway)
without disabling them for other backends too.

Thanks Alexey, i will take a look again when i have time!

alexeygrigorev:
> If they are standard algorithm tests coming from the abstract test suites in math-scala, they currently cannot be disabled (in any way i know anyway) without disabling them for other backends too.

Yes, they are the standard tests, I unfortunately also don't a way to disable a specific test.

Probably I can just disable the suites with failing tests completely, with a note that these tests should be fixed (and run when developing/changing something). I hope that my tests should give enough coverage to detect errors if they are introduced to the existing code.

dlyubimov:
Ok here is what i suggest.

Since the code is still a bit raw (in a sense of passing tests), let's
indeed pull Alexey's work into a dev mahout branch and then create a new
request to merge to master to track merge conflicts (github will quick to
note those).

The reason is that i proably may want to do some small changes to it so the
assumption is that it would be more visible (and easier for me:) to do it
that way than to actually send a PR to Alexey's github project to enter
changes into the pr.

If ti is ok, i can do it.

-d

alexeygrigorev:
sure. please let me know if there's anything I can do (rebase, look into something, etc)

andrewpalumbo:
+1

andrewpalumbo:
I looked at this briefly last night, was able to get it down to only 2 failing tests (of 6 total) in the DistributedDecompositionSuite: dspca and dals.

It looks like `FlinkDrm.blockify()` is creating or adding at least one empty partition to the `BlockifiedFlinkDrm[K]` (possibly some idiosyncrasy with the backing flink Dataset..?  just a quick guess).

It's probably something deeper than that causing the problems, but by adding the below check for an empty partition, all DSSVD tests pass.

The DSPCA test runs without exception but fails with an incorrect finishing value.  The DALS test runs out of heap space even with -Xmx4g set.

FlinkDrm.scala:
```
def blockify(): BlockifiedFlinkDrm[K] = {
  val ncolLocal = ncol
  val classTag = implicitly[ClassTag[K]]

  val parts = ds.mapPartition(new MapPartitionFunction[DrmTuple[K], (Array[K], Matrix)] {
    def mapPartition(values: Iterable[DrmTuple[K]], out: Collector[(Array[K], Matrix)]): Unit = {
      val it = values.asScala.seq
        val (keys, vectors) = it.unzip

        if (!(vectors.isEmpty)) { // <-- adding this check makes DSSVD tests pass

        val isDense = vectors.head.isDense

        if (isDense) {
          val matrix = new DenseMatrix(vectors.size, ncolLocal)
          vectors.zipWithIndex.foreach { case (vec, idx) => matrix(idx, ::) := vec }
          out.collect((keys.toArray(classTag), matrix))
        } else {
          val matrix = new SparseRowMatrix(vectors.size, ncolLocal, vectors.toArray)
          out.collect((keys.toArray(classTag), matrix))
        }
      }
    }
  })

  new BlockifiedFlinkDrm(parts, ncol)
}
```

a comment from @dlyubimov on this partitioning issue:
>
> apparently flink plans (product splitting) may produce empty partitions. Spark physical operators were careful to avoid this when deciding on the product parallelism, so with Spark this is not known to happen (neither with h20 for that matter). 
>
> Technically of course the codes must survive empty partition situations. However, it has been patched only on a need-to-do basis, so obviously there are spots like blockify() that are not really fixed yet out of mere necessity. Since like i said the spark optimizer is careful to avoid this, it apparently never happened before (and i ran a lot of spark code on this). 
>
> So... of course we should patch the blockify code. However, the real problem IMO is the careful review of the splitting logic in Flink physical layer. Empty splits generally should not happen more often than is absolutely unavoidable (especially if this happens after drmParallelize that explicitly said to make 2 partitions out of 500-odd rows, give or take).

dlyubimov:
ok i pushed that branch (which is very nicely rebased, by the way, thank you!) to apache repo. I may want to make few changes.  When i do, i will be submitting PRs against this branch in mahout repo (and commit if ok'd).

pom.xml
```diff
@@ -121,6 +121,8 @@
     <scala.compat.version>2.10</scala.compat.version>
     <scala.version>2.10.4</scala.version>
     <spark.version>1.3.1</spark.version>
+    <!-- TODO: Remove snapshot dependency when Flink 0.9.1 is released -->
+    <flink.version>0.9-SNAPSHOT</flink.version>
```
hsaputra: Flink 0.9.1 is out so we could remove the SNAPSHOT label I suppose.

dlyubimov: @alexey : BTW is there such a thing as Flink shell?

hsaputra: Flink has Scala shell:
https://ci.apache.org/projects/flink/flink-docs-master/apis/scala_shell.html

smarthi: there is a Flink-scala-shell in the flink project. It should be available
with the flink project binaries.

alexeygrigorev: I also know there were plans to do Mahout DSL support for the shell, but I'm not sure if it got any further.

dlyubimov:
i guess we can continue using this PR to see the current diff if Alexey pulls our mahout branch here (it of course turns out there's no way to make PR off the github repo mirror)

alexeygrigorev:
@dlyubimov just to clarify: you're suggesting I pull from `mahout:flink-binding` to my `:flink-binding`, so all new commits from that branch appear here?

dlyubimov:
Maybe. we need something to review the diff of the flink branch vs. master.
can't create a PR since we do not own github's mahout repo.

smarthi:
@alexey, yes please update ur local flink-bindings repo from
mahout:flink-binding

smarthi:
Nevertheless, we have pushed some changes up to the mahout : flink-binding
that @alexey needs to pull from

smarthi:
@alexey, the VectorKryo and GenericMatrixKryo classes have been refactored
out to a common math-scala package.

After u update ur local branch with the latest from mahout:flink-binding, u
could try Till's suggested fixes for OOM (and if possible other failing
tests) and create a PR.

alexey:
guys its very interesting but please mention correct user during your comments :)
eg: @alexeygrigorev not @alexey

smarthi:
Fixed all references to correct @alexeygrigorev :)

alexeygrigorev:
the new commits are pulled

PR #140, opened 2015-06-22
MAHOUT-1738: implement MatrixFlavor for H2OBlockMatrix

avati:
MAHOUT-1738: implement MatrixFlavor for H2OBlockMatrix


dlyubimov:
thanks..
ok... so the MMul  related tests pass now?
and the h20 matrix is always row-backed?

dlyubimov:
i guess it is related to MAHOUT-1738... should be added to description for asfgit to duplicate to jira and dev@...

avati:
yes, MMul tests pass now

PR #141, opened 2015-06-23
NOJIRA: dynamically determine evalZeros in OpAewUnaryFunc

avati:
- saves sparse computation in operators like "+ 0.0", "- 0.0" etc.
  (when the scalar comes as a variable that happens to be 0.0)

- makes code less vulnerable to errors like a missed setting of
  "evalZeroes = true" parameter while adding a new operator in
  the future.


avati:
@dlyubimov - this changes code you recently added, hence bringing attention

dlyubimov:
sorry. -1. we already discussed this. this assumes all functions are deterministic (not true).

dlyubimov:
This also assumes all functions have 0.0 in their domain. not true either  (x => 1/ x doesn't ).

avati:
I agree about the deterministic part, therefore invalid.. but 1/0.0 is supposed to be Double.INFINITY, therefore in domain, right?

avati:
functions can be non-deterministic (like calling a rng). abandoning

dlyubimov:
1/0.0 produces NaN

PR #143, opened 2015-06-24
MAHOUT-1736: Implement allreduceBlock() on H2O

avati:
Closes apache/mahout#143

### DIFF
--- a/h2o/src/main/java/org/apache/mahout/h2obindings/H2OHelper.java
+++ b/h2o/src/main/java/org/apache/mahout/h2obindings/H2OHelper.java
@@ -33,13 +33,18 @@ import water.util.ArrayUtils;
 
 import java.util.Map;
 import java.util.HashMap;
+import java.io.Serializable;
 
 import org.apache.mahout.h2obindings.drm.H2ODrm;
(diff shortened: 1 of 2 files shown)


dlyubimov:
Seems like it is doing what it needs to do..

dlyubimov:
@avati question. 
so all in-core operations work for H20Matrix in the backend too? for example, if allreduceBlock defines reduce as x1 rbind x2 (stacking concatenation) that would work ?

i don't think we repeat all in-core tests for H20Matrix

mostly because we assumed the backends would use the same classes in the back from mahout-math. 

We confirm functionality, indirectly, by running current distributed algorithms, but they don't cover everything that in-core tests in mahout-math-scala cover for in-core matrices.

avati:
@dlyubimov H2OBlockMatrix should work like any other in-core matrix in the backend. In fact its lifecycle is valid only in the backend (you cannot spin-up a H2OBlockMatrix from scratch in the "frontend").

Yes, you are right that we are not running all the in-core matrix tests. We will have to redesign all the tests to be run in the context of a MapBlock or allreduceBlock closure. That's the only way to create a H2OBlockMatrix.

PR #145, opened 2015-06-24
Cumulative patch for MAHOUT-1746, MAHOUT-1753, MAHOUT-1754.

dlyubimov:


math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
```diff
@@ -160,6 +160,165 @@ package object drm {
...
+    import RLikeDrmOps._
+
+    val drmAcp = drmA.checkpoint()
+    
+    val mu = drmAcp colMeans
+
+    // Compute variance using mean(x^2) - mean(x)^2
+    val variances = (drmAcp ^ 2 colMeans) -=: mu * mu
+
+    mu → variances
```
avati: Is there a strong reason for the unicode right arrow? For me at least, the only place I can see it right is in this github view. It looks garbled in emacs, less, and every other console based tools I use.

math-scala/src/main/scala/org/apache/mahout/math/drm/RLikeDrmOps.scala
```diff
@@ -50,7 +50,11 @@ class RLikeDrmOps[K: ClassTag](drm: DrmLike[K]) extends DrmLikeOps[K](drm) {
 
   def *:(that: Double): DrmLike[K] = OpAewUnaryFunc[K](A = this, f = that * _)
 
-  def ^(that: Double): DrmLike[K] = OpAewUnaryFunc[K](A = this, f = math.pow(_, that))
+  def ^(that: Double): DrmLike[K] = that match {
+    case 2.0 ⇒ OpAewUnaryFunc[K](A = this, f = x ⇒ x * x)
```
avati: Can we have a comment here explaining the reason behind special-case for 2.0 and 0.5? This will help future code readers. Same comment applies to rest of the similar changes further below.

math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
dlyubimov: Other than it is valid Scala style? no. 

btw every tool i have shows it correctly for me, including less, web pages, latex/lyx, etc. etc. Use intellij. really.

math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
```diff
@@ -160,6 +160,165 @@ package object drm {
...
+    val sBcast = drmBroadcast(drmXcp ^ 2 rowSums)
+    val tBcast = drmBroadcast(drmYcp ^ 2 rowSums)
+
+    (drmX %*% drmY.t)
+
+      // Apply the rest of the formula
+      .mapBlock() { case (keys, block) =>
+
+      // Cache broadcast representations in local task variable
+      val s = tBcast: Vector
```
andrewpalumbo: Should this be: `val t = sBcast: Vector` ?
 (rather than `tBcast: Vector`)

andrewpalumbo: oops.. I had a typo in the above comment (edited now)..  

`val s = tBcast: Vector` should be:
`val s = sBcast: Vector`

dlyubimov: Of course it should. Typo when copied? Strange that tests ran fine... it obviously would be a wrong result at least and index oob at most

andrewpalumbo: Yeah- definite copy/paste typo.  The h2o/spark tests catch it.  After changing it, the tests run fine.

PR #146, opened 2015-07-03
MAHOUT-1653-master with shell upgrade to Spark 1.3

andrewpalumbo:
Upgrade to Spark 1.3.  The shell is not backwards compatible.


andrewpalumbo:
I will push this in the next couple of days if there are no objections.

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
```diff
@@ -48,55 +77,63 @@ class MahoutSparkILoop extends SparkILoop {
...
+  // this is technically not part of Sparks explicitly defined Developer API though
+  // nothing in the SparkILoopInit.scala file is marked as such.
   override def initializeSpark() {
-    intp.beQuietDuring {
-      command("""
+    _interp.beQuietDuring {
+      _interp.interpret("""
 
-         @transient implicit val sdc: org.apache.mahout.math.drm.DistributedContext =
+         @transient implicit val sc: org.apache.mahout.math.drm.DistributedContext =
```
dlyubimov: What's the rationale behind this rename? 

I'd rather keep it as is. 

My reasoning is, ok spark shells have 'sc' reserved for spark context, so we can't use the same variable name just to avoid clashes between mahout-enabled scripts and just spark-enabled scripts that potentially may expect spark context to be named "sc" (not likely since it is implicit, but still).

dlyubimov: comes to that, we could potentially initialize sc with spark context and sdc with mahout context. perhaps that would be ok and create compatibility throughout.

andrewpalumbo: the problem is that the clean up method that we were using is not part of the Spark REPL Developer API.  So the method that we were using to close the SparkContext:
```
def sparkCleanUp() {
      echoToShell("Stopping Spark context.")
      _interp.beQuietDuring {
        _interp.interpret("sdc.stop()")
      }
   }
```

is not available to us to override anymore.  So upon exiting the shell, the private `SparkIloop.sparkCleanUp()` is called and looks to call `sc.close()`: 
  
```
private def sparkCleanUp(){
    echo("Stopping spark context.")
    intp.beQuietDuring {
      command("sc.stop()")
    }
  }
 ```

and giving an error,  leaving our `sdc` context open.

Comment (andrewpalumbo) [inline spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala]: > comes to that, we could potentially initialize sc with spark context and sdc with mahout context. perhaps that would be ok and create compatibility throughout.

Are you thinking something like:

```
@transient val sc = {
           val _sc = org.apache.spark.repl.Main.interp.createSparkContext()
           println("Spark context available as sc.")
           _sc
         }
@transient implicit val sdc: org.apache.mahout.math.drm.DistributedContext =
            new org.apache.mahout.sparkbindings.SparkDistributedContext(sc)

        echoToShell("Mahout distributed context is available as \"implicit val sdc\".")
```

I'm not sure where we'd be able to call `sdc.close()`  but I'm not sure that that would be necessary anyway since that method in turn simply calls `sc.stop()` which is handled by `SparkILoop.sparkCleanUp()`

Comment (andrewpalumbo) [inline spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala]: Actually, now that I'm looking at it again without worrying about the possibility of backwards compatibility <1.3 we should also  be creating a SqlContext in that `initalizeSpark()` call:


From SparkILoop.initializeSpark():
```
      command("""
         @transient val sqlContext = {
           val _sqlContext = org.apache.spark.repl.Main.interp.createSQLContext()
           println("SQL context available as sqlContext.")
           _sqlContext
         }
        """)
      command("import org.apache.spark.SparkContext._")
      command("import sqlContext.implicits._")
      command("import sqlContext.sql")
      command("import org.apache.spark.sql.functions._")
```

Comment (dlyubimov) [inline spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala]: yes.

Except we are already calling "createMahoutContext" during createSparkContext, so we just need to stash its output as Mahout context private attribute and pass it in here to implicit. (perhaps via a public getter or val).

as for deinitalization, you are right -- all it does i think it stops the spark context.

Comment (andrewpalumbo) [inline spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala]: Ok gotcha.. thanks. I guess it also really doesn't matter where the spark context is stopped either.

File: spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala [hunk 2]
```diff
@@ -48,55 +77,81 @@ class MahoutSparkILoop extends SparkILoop {
...
             new org.apache.mahout.sparkbindings.SparkDistributedContext(
             org.apache.spark.repl.Main.interp.createSparkContext())
 
+                        """)
+      echoToShell("Mahout distributed context is available as \"implicit val sdc\".")
+
+      // get the spark context from the mahout distributed context.
+      _interp.interpret("import org.apache.spark.SparkContext._")
+      _interp.interpret("@transient val sc: org.apache.spark.SparkContext = sdc.sc")
+      echoToShell("Spark context is available as \"val sc\".")
```
andrewpalumbo: Reset the mahout distributed context to `implicit val sdc`.  Set `val sc` to the SparkContext in `sdc`.  Created a SqlContext as is done in the overridden method.

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
```diff
@@ -48,55 +77,81 @@ class MahoutSparkILoop extends SparkILoop {
 
     conf.set("spark.executor.memory", "1g")
 
-    sparkContext = mahoutSparkContext(
+    _interp.sparkContext = mahoutSparkContext(
       masterUrl = master,
       appName = "Mahout Spark Shell",
       customJars = jars,
       sparkConf = conf
     )
```
dlyubimov: ok what i meant is that this should really be saved as MahoutDistributed context :

    private val sdc = mahoutSparkContext(...) 

which is already of type SparkDistributedContext (mahout type), right?

and then use to initialize `sdc = this.sdc` and `sc = this.sdc.sc` later... We still seem to be crating DistributedSparkContext twice here, which is probably not a big deal but feels a bit hacky...

andrewpalumbo:
Ok- I see what you are saying..  I don't think that we can access that field From the interpreter though.  ie:
```
val sdc = mahoutSparkContext(...)  // or something like this...
_interp.interepret("@transient implicit val sdc = this.sdc")
``` 
will give an error like:
```
error: value sdc is not a member of iwC$iwC$iwC$iwC.....
```
I believe since the line is converted to an object by the interpreter not of type MahoutSparkILoop. 

What we are currently doing is actually following the Spark REPL blueprint:

// Create a SparkContext in SparkILoop
https://github.com/apache/spark/blob/branch-1.3/repl/scala-2.10/src/main/scala/org/apache/spark/repl/SparkILoop.scala#L1005

// Create an other SparkContext in the Intrepter:
https://github.com/apache/spark/blob/branch-1.3/repl/scala-2.10/src/main/scala/org/apache/spark/repl/SparkILoopInit.scala#L121

Let me know if you think there's a better way around it.

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
andrewpalumbo: Ok- I see what you are saying..  I don't think that we can access that field From the interpreter though.  ie:
```
val sdc = mahoutSparkContext(...)  // or something like this...
_interp.interepret("@transient implicit val sdc = this.sdc")
``` 
will give an error like:
```
error: value sdc is not a member of iwC$iwC$iwC$iwC.....
```
I believe since the line is converted to an object by the interpreter not of type MahoutSparkILoop. 

What we are currently doing is actually following the Spark REPL blueprint:

// Create a SparkContext in SparkILoop
https://github.com/apache/spark/blob/branch-1.3/repl/scala-2.10/src/main/scala/org/apache/spark/repl/SparkILoop.scala#L1005

// Create an other SparkContext in the Intrepter:
https://github.com/apache/spark/blob/branch-1.3/repl/scala-2.10/src/main/scala/org/apache/spark/repl/SparkILoopInit.scala#L121

Let me know if you think there's a better way around it.
 

oops.. this should have been in a line note..

dlyubimov: "@transient implicit val sdc =  org.apache.spark.repl.Main.interp.asInstanceOf[o.a.m.s.s.MahoutSparkILoop.sdc"
    "@transient implicit val sc = sdc.sc"

?

andrewpalumbo:
```
var sdc: SparkDistributedContext = _

override def createSparkContext(): SparkContext = {
[...]
   sdc = mahoutSparkContext(
      masterUrl = master,
      appName = "Mahout Spark Shell",
      customJars = jars,
      sparkConf = conf
    )

    _interp.sparkContext = sdc
[...]
}
  
_interp.interpret("""
      @transient implicit val sdc =
         org.apache.spark.repl.Main.interp.asInstanceOf[org.apache.mahout.sparkbindings.shell.MahoutSparkILoop].sdc
         """)
```
gives a NPE:
```
java.lang.NullPointerException
	at $iwC$$iwC$$iwC$$iwC.<init>(<console>:12)
	at $iwC$$iwC$$iwC.<init>(<console>:17)
	at $iwC$$iwC.<init>(<console>:19)
	at $iwC.<init>(<console>:21)
```

andrewpalumbo: uggh.. of course-- i was doing it backwards.  the SparkContext (along with the SparkDistributedContext) have to be created by a call to ```createSparkContext()``` before we can access sdc (hence the above NPE).
```
   _interp.interpret("""
         @transient val sc = {
           val _sc = org.apache.spark.repl.Main.interp.createSparkContext()
           println("Spark context available as sc.")
           _sc
         }
              """)

      _interp.interpret("""

         @transient implicit val sdc: org.apache.mahout.sparkbindings.SparkDistributedContext =
            org.apache.spark.repl.Main.interp.asInstanceOf[org.apache.mahout.sparkbindings.shell.MahoutSparkILoop].sdc

                        """)

      echoToShell("Mahout distributed context is available as \"implicit val sdc\".")
```

works fine.  Had it in my head that  2 SparkContexts were being created rather than 2 MahoutSparkContextes...

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
```diff
@@ -48,55 +83,87 @@ class MahoutSparkILoop extends SparkILoop {
...
-      command("""
 
-         @transient implicit val sdc: org.apache.mahout.math.drm.DistributedContext =
-            new org.apache.mahout.sparkbindings.SparkDistributedContext(
-            org.apache.spark.repl.Main.interp.createSparkContext())
+    _interp.beQuietDuring {
+
+      // get the spark context, at the same time create and store a mahout distributed context.
+      _interp.interpret("""
+         @transient val sc = {
```
andrewpalumbo: @dlyubimov I've made the necessary changes clean up the the redundant creation of a SparkDistributedContext.  Thanks alot for the input.  

You mentioned is that we could declare the SparkContext as `implicit val sc = ...`. Is there a reason that it should be implicit?

I've left it as `val sc =...` for now since that is the way Spark declares is in this method.     Thx.

dlyubimov: i honestly don't know if spark declared it as implicit. if it does not, we
don't have to either i suppose.

the reason mahout context is implicit is because all top-level routines
like drmParallelize(...) look for it.

andrewpalumbo: Thank you.  I'll leave it as is.

andrewpalumbo:
I've added in a final commit to set the `spark.executor.memory` value only if it is not already set eg.  in MAHOUT_OPTS.

This should be good to go.  I'll push it this afternoon if there are no objections.

dlyubimov:
good idea.
yes. +1

andrewpalumbo:
Commited to master on 7/10.. Closing manually.. did not close by commit message.

dlyubimov:
thanks!

PR #149, opened 2015-07-15
NOJIRA: minor row similarity cleanup

pferrel:
fixed an intermittent test failure, cleaned up row similarity help, fixed a minor bug reading into row similarity where the strength delimiter was in text but omitStrength was chosen

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -2,6 +2,8 @@ Mahout Change Log
 
 Release 0.11.0 - unreleased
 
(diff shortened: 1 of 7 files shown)


pferrel:
added par calls before any drm ops, which caused a significant speedup, went from 35 minutes on epinions with rate and trust data (cross-cooccurrence) to 23 minutes! This on a pseudo cluster.

pferrel:
I'm going to push this later today if no objection

dlyubimov:
noted 

("noted" in my reviews means i did not find any problem to the degree i am immersed to the problem, which is not a lot in this case). == +0.5

PR #153, opened 2015-07-17
MAHOUT-1756: Adding missing in-place operations in the form  5 +=: vec-expr and 5*=: vec-expr or vec *=: vec-expr

smarthi:


dlyubimov:
+1

dlyubimov:
can we please commit those to 0.11 and 0.10.2? thanks.

PR #157, opened 2015-08-29
MAHOUT-1773: Fix cluster-syntheticcontrol.sh for HDFS synthax

edniemeyer:
Added  '/' to some lines, so it can check the root of the HDFS, where it's possible to use -mkdir, -put and others. Otherwise the example keeps failing (Hadoop 2.4.1).

### DIFF
--- a/examples/bin/cluster-syntheticcontrol.sh
+++ b/examples/bin/cluster-syntheticcontrol.sh
@@ -66,13 +66,13 @@ if [ ! -f ${WORK_DIR}/synthetic_control.data ]; then
 fi
 if [ "$HADOOP_HOME" != "" ] && [ "$MAHOUT_LOCAL" == "" ]; then
   echo "Checking the health of DFS..."
-  $DFS -ls
+  $DFS -ls /
   if [ $? -eq 0 ];then 
     echo "DFS is healthy... "
     echo "Uploading Synthetic control data to HDFS"
     $DFSRM testdata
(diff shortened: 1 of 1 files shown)


dlyubimov:
I am guessing this is the MapReduce stuff.

edniemeyer:
Done! Sorry for not doing it before! :)

chu11:
I recently hit this issue.  As a note for other users, I solved this issue an alternate way (of course my issue may have been different than the OP).  In my environment the user's home directory had not been created (/user/USERNAME), thus leading to an error when trying to do HDFS operations with a relative path.  The prefixing of the "/" is a more than suitable fix ensuring it works regardless of the user's environment setup.

examples/bin/cluster-syntheticcontrol.sh
```diff
@@ -66,13 +66,13 @@ if [ ! -f ${WORK_DIR}/synthetic_control.data ]; then
 fi
 if [ "$HADOOP_HOME" != "" ] && [ "$MAHOUT_LOCAL" == "" ]; then
   echo "Checking the health of DFS..."
-  $DFS -ls
+  $DFS -ls /
   if [ $? -eq 0 ];then 
     echo "DFS is healthy... "
     echo "Uploading Synthetic control data to HDFS"
     $DFSRM testdata
```
andrewmusselman: Shouldn't this have a slash in it too?

PR #159, opened 2015-09-29
MAHOUT-1774 Enable compilation with a 1.8 jdk

dlyubimov:


dlyubimov:
is it ok for now?

pom.xml
```diff
@@ -659,7 +659,7 @@
             <configuration>
               <rules>
                 <requireJavaVersion>
-                  <version>[1.7, 1.8)</version>
+                  <version>[1.7, 1.9)</version>
```
smarthi: In the flink branch, I have it as [1.7, ). Keep it that way

PR #160, opened 2015-10-12
MAHOUT-1772

linahovanessian:
Sorry for 3 continues check-ins ! I had a problem with my SSH.
The valid change is the latest one , where I set the variable to transient.

  private final transient Splitter delimiterPattern;


dlyubimov:
Do we accept taste patches?

It would be great if Sean Owen provided a feedback on this.

PR #161, opened 2015-10-20
MAHOUT-1570, sub-pr: a siggestion: let's unify all key class tag extractors. 

dlyubimov:
Unifying "keyClassTag" of checkpoitns and "classTagK" of logical operators and

elevating "keyClassTag" into DrmLike[] trait. No more logical forks any more .

### DIFF
--- a/flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
+++ b/flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
@@ -20,7 +20,7 @@ package org.apache.mahout.flinkbindings.drm
 
 import scala.collection.JavaConverters._
 import scala.util.Random
-import scala.reflect.ClassTag
+import scala.reflect.{ClassTag, classTag}
 import org.apache.flink.api.common.functions.MapFunction
 import org.apache.flink.api.common.functions.ReduceFunction
 import org.apache.flink.api.java.hadoop.mapred.HadoopOutputFormat
@@ -75,7 +75,8 @@ class CheckpointedFlinkDrm[K: ClassTag](val ds: DrmDataSet[K],
     list.head
   }
 
-  def keyClassTag: ClassTag[K] = implicitly[ClassTag[K]]
+
+  override val keyClassTag: ClassTag[K] = classTag[K]
 
   def cache() = {
     // TODO
--- a/flink/src/main/scala/org/apache/mahout/flinkbindings/package.scala
+++ b/flink/src/main/scala/org/apache/mahout/flinkbindings/package.scala
@@ -108,11 +108,6 @@ package object flinkbindings {
     new CheckpointedFlinkDrm[K](dataset)
   }
 
-  private[flinkbindings] def extractRealClassTag[K: ClassTag](drm: DrmLike[K]): ClassTag[_] = drm match {
-    case d: CheckpointAction[K] => d.classTag
-    case d: CheckpointedFlinkDrm[K] => d.keyClassTag
-    // will not always return correct result, often result in Any
-    case _ => implicitly[ClassTag[K]]
-  }
+  private[flinkbindings] def extractRealClassTag[K: ClassTag](drm: DrmLike[K]): ClassTag[_] = drm.keyClassTag
 
 }
--- a/h2o/src/main/scala/org/apache/mahout/h2obindings/H2OEngine.scala
+++ b/h2o/src/main/scala/org/apache/mahout/h2obindings/H2OEngine.scala
@@ -112,7 +112,7 @@ object H2OEngine extends DistributedEngine {
       case op@OpRowRange(a, r) => RowRange.exec(tr2phys(a)(op.classTagA), r)
       // Custom operators
       case blockOp: OpMapBlock[K, _] => MapBlock.exec(tr2phys(blockOp.A)(blockOp.classTagA), blockOp.ncol, blockOp.bmf,
-        (blockOp.classTagK == implicitly[ClassTag[String]]), blockOp.classTagA, blockOp.classTagK)
+        (blockOp.keyClassTag == implicitly[ClassTag[String]]), blockOp.classTagA, blockOp.keyClassTag)
       case op@OpPar(a, m, e) => Par.exec(tr2phys(a)(op.classTagA), m, e)
       case cp: CheckpointedDrm[K] => cp.h2odrm
       case _ => throw new IllegalArgumentException("Internal:Optimizer has no exec policy for operator %s."
--- a/h2o/src/main/scala/org/apache/mahout/h2obindings/drm/CheckpointedDrmH2O.scala
+++ b/h2o/src/main/scala/org/apache/mahout/h2obindings/drm/CheckpointedDrmH2O.scala
@@ -18,6 +18,8 @@ class CheckpointedDrmH2O[K: ClassTag](
   val context: DistributedContext
 ) extends CheckpointedDrm[K] {
 
+  override val keyClassTag: ClassTag[K] = classTag[K]
+
   /**
     * Collecting DRM to in-core Matrix
     *
@@ -26,9 +28,6 @@ class CheckpointedDrmH2O[K: ClassTag](
     * as rowLabelBindings of the in-core matrix.
     */
   def collect: Matrix = H2OHelper.matrixFromDrm(h2odrm)
-
-  /** Explicit extraction of key class Tag   */
-  def keyClassTag: ClassTag[K] = implicitly[ClassTag[K]]
 
   /* XXX: call frame.remove */
   def uncache(): this.type = this
--- a/math-scala/src/main/scala/org/apache/mahout/math/drm/CheckpointedDrm.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/drm/CheckpointedDrm.scala
@@ -18,7 +18,6 @@
 package org.apache.mahout.math.drm
 
 import org.apache.mahout.math.Matrix
-import scala.reflect.ClassTag
 
 /**
  * Checkpointed DRM API. This is a matrix that has optimized RDD lineage behind it and can be
@@ -33,13 +32,6 @@ trait CheckpointedDrm[K] extends DrmLike[K] {
 
   /** If this checkpoint is already declared cached, uncache. */
   def uncache(): this.type
-
-  /**
-   * Explicit extraction of key class Tag since traits don't support context bound access; but actual
-   * implementation knows it
-   */
-  def keyClassTag: ClassTag[K]
-
 
   /** changes the number of rows without touching the underlying data */
   def newRowCardinality(n: Int): CheckpointedDrm[K]
--- a/math-scala/src/main/scala/org/apache/mahout/math/drm/DrmLike.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/drm/DrmLike.scala
@@ -25,6 +25,8 @@ import scala.reflect.ClassTag
  */
 trait DrmLike[K] {
 
+  val keyClassTag: ClassTag[K]
+
   protected[mahout] def partitioningTag: Long
 
   protected[mahout] def canHaveMissingRows: Boolean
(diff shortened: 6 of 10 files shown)


dlyubimov:
spark backend tests pass, but in h20 i get 

`
10-19 23:45:46.002 192.168.11.4:54321    13168  #onsSuite INFO: Cloud of size 1 formed [/192.168.11.4:54321]
*** RUN ABORTED ***
  java.lang.StackOverflowError:
  at org.apache.mahout.math.drm.DistributedEngine$.org$apache$mahout$math$drm$DistributedEngine$$pass1(DistributedEngine.scala:142)
  at org.apache.mahout.math.drm.DistributedEngine$.org$apache$mahout$math$drm$DistributedEngine$$pass1(DistributedEngine.scala:182)
  at org.apache.mahout.math.drm.DistributedEngine$class.optimizerRewrite(DistributedEngine.scala:44)
`

well the H20 build is known for that it never builds for me anyway :)

dlyubimov:
Ok so yes -- dspca produces just wrong result, and the dssvd test just fails with out of memory. 

 java.lang.OutOfMemoryError: Java heap space
  at java.util.Arrays.copyOf(Arrays.java:2271)
  at java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:113)
  at java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:93)
  at java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:140)


I would start with tracing problems with dssvd first. It looks like some memory management settings for flink. Any ideas? Should we just bump it the jvm memory of maven test?

hsaputra:
Hi @dlyubimov , is there any more information of the stack trace for the failure?

dlyubimov:
Full test report (with all failures) 

<script src="https://gist.github.com/dlyubimov/bfe0c7356ff775a8852b.js"></script>

dlyubimov:
Aha. so a lot of those are actually failures associated with empty partitions. It looks like. just what Andrew reported.
optimizer should not (willingly) produce empty splits.

andrewpalumbo:
with the memory bumped up to 4G (inside IntelliJ) the stack trace shows FlinkOpAtA as the source of the oom error in dals:
```
Exception in thread "main" java.lang.OutOfMemoryError: Java heap space
	at java.util.Arrays.copyOf(Arrays.java:2271)
	at java.io.ByteArrayOutputStream.grow(ByteArrayOutputStream.java:113)
	at java.io.ByteArrayOutputStream.ensureCapacity(ByteArrayOutputStream.java:93)
	at java.io.ByteArrayOutputStream.write(ByteArrayOutputStream.java:140)
	at java.io.ObjectOutputStream$BlockDataOutputStream.drain(ObjectOutputStream.java:1876)
	at java.io.ObjectOutputStream$BlockDataOutputStream.flush(ObjectOutputStream.java:1821)
	at java.io.ObjectOutputStream.flush(ObjectOutputStream.java:718)
	at java.io.ObjectOutputStream.close(ObjectOutputStream.java:739)
	at org.apache.flink.util.InstantiationUtil.serializeObject(InstantiationUtil.java:319)
	at org.apache.flink.util.InstantiationUtil.writeObjectToConfig(InstantiationUtil.java:268)
	at org.apache.flink.runtime.operators.util.TaskConfig.setStubWrapper(TaskConfig.java:273)
	at org.apache.flink.optimizer.plantranslate.JobGraphGenerator.createDataSourceVertex(JobGraphGenerator.java:864)
	at org.apache.flink.optimizer.plantranslate.JobGraphGenerator.preVisit(JobGraphGenerator.java:260)
	at org.apache.flink.optimizer.plantranslate.JobGraphGenerator.preVisit(JobGraphGenerator.java:103)
	at org.apache.flink.optimizer.plan.SourcePlanNode.accept(SourcePlanNode.java:87)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.SingleInputPlanNode.accept(SingleInputPlanNode.java:199)
	at org.apache.flink.optimizer.plan.OptimizedPlan.accept(OptimizedPlan.java:127)
	at org.apache.flink.optimizer.plantranslate.JobGraphGenerator.compileJobGraph(JobGraphGenerator.java:170)
	at org.apache.flink.client.LocalExecutor.executePlan(LocalExecutor.java:176)
	at org.apache.flink.api.java.LocalEnvironment.execute(LocalEnvironment.java:54)
	at org.apache.flink.api.java.ExecutionEnvironment.execute(ExecutionEnvironment.java:789)
	at org.apache.flink.api.java.DataSet.collect(DataSet.java:408)
	at org.apache.mahout.flinkbindings.blas.FlinkOpAtA$.slim(FlinkOpAtA.scala:57)
	at org.apache.mahout.flinkbindings.blas.FlinkOpAtA$.at_a(FlinkOpAtA.scala:39)
	at org.apache.mahout.flinkbindings.FlinkEngine$.flinkTranslate(FlinkEngine.scala:124)
	at org.apache.mahout.flinkbindings.FlinkEngine$.toPhysical(FlinkEngine.scala:91)
```

alexeygrigorev:
Thanks for looking into it! It may be my bug, but I'll let the Flink team know about it anyways.

tillrohrmann:
The problem of the `OutOfMemoryError` in `dals` of the `DistributedDecompositionSuite` is the following: 

In order parallelize the input matrix, the `FlinkEngine` creates a `Collection[DrmTuple[Int]]` which is given to Flink's `CollectionInputFormat`. The `CollectionInputFormat` is simply a wrapper for a collection which is serialized and shipped to the corresponding data source at runtime. The problem is that the matrix rows in the collection of `DrmTuple[Int]` are of type `MatrixVectorView`. Instances of this class hold a reference to the original matrix. Therefore, serializing an instance of `MatrixVectorView` effectively boils down to serializing the complete input matrix for each row. This means that the `CollectionInputFormat` needs memory of order `size(inputMatrix) * rows(inputMatrix)` to serialize the given collection of `DrmTuple[Int]` instances.

The input size for the dals test case are `500 * 500`. This means that the input matrix has roughly the size of 2 MB. The serialized `CollectionInputFormat` has then a size of roughly 1 GB, because we serialize the matrix for each row. If you replace `FlinkEngine.scala:231` with `val rows = (0 until m.nrow).map(i => (i, dvec(m(i, ::)).asInstanceOf[Vector]))`, the test will pass the serialization point (but fail with another error :-( The reason why 1 GB of memory, given that you set the max limit to 4 GB, is already enough to crash Flink is that Flink reserver per default 70% of the available memory for its managed memory (for sorting and hashing).

I think it would make sense to convert possible `Views` to its materialized form before distributing the data in the cluster. That way, you avoid the creation of unnecessary data which might be ok for the local use case where the data is shared.

I hope this sheds a little bit of light in the `OutOfMemoryError` you've encountered.

tillrohrmann:
I forgot to mention that the error is

```
Caused by: java.util.NoSuchElementException: head of empty list
	at scala.collection.immutable.Nil$.head(List.scala:337)
	at scala.collection.immutable.Nil$.head(List.scala:334)
	at org.apache.mahout.flinkbindings.drm.RowsFlinkDrm$$anon$1.mapPartition(FlinkDrm.scala:68)
	at org.apache.flink.runtime.operators.MapPartitionDriver.run(MapPartitionDriver.java:98)
	at org.apache.flink.runtime.operators.RegularPactTask.run(RegularPactTask.java:496)
	at org.apache.flink.runtime.operators.RegularPactTask.invoke(RegularPactTask.java:362)
	at org.apache.flink.runtime.taskmanager.Task.run(Task.java:559)
```

The error appears also in some of the other test cases, therefore I think it's not a problem of Flink itself but of the implementation of the Flink bindings.

smarthi:
That's been my conclusion too, Till. Its an issue in Flink bindings impl. Thanks for ur time and feedback.

dlyubimov:
On Wed, Oct 21, 2015 at 5:48 AM, Till Rohrmann <notifications@github.com>
wrote:

> The problem of the OutOfMemoryError in dals of the
> DistributedDecompositionSuite is the following:
>
> In order parallelize the input matrix, the FlinkEngine creates a
> Collection[DrmTuple[Int]] which is given to Flink's CollectionInputFormat.
> The CollectionInputFormat is simply a wrapper for a collection which is
> serialized and shipped to the corresponding data source at runtime. The
> problem is that the matrix rows in the collection of DrmTuple[Int] are of
> type MatrixVectorView. Instances of this class hold a reference to the
> original matrix. Therefore, serializing an instance of MatrixVectorView
> effectively boils down to serializing the complete input matrix for each
> row.
>
This has never been a problem for other mappings and should not be here.
What you serialize is either (1) collection of vectors, or (2) a matrix
view.

Java serialization of the mahout-math is not supported. so it has to be a
custom serialization in each case. Mahout core provides two serialization
means: (1) hadoop writable, (2) kryo.

in particular, spark backend uses kryo serialization.

In this case i guess we are speaking of collection of vectors, so the
classes supprting these types of serialization would be VectorWritable and
[1]. Kryo is generally a preferred way. We can move kryo support into
math-scala out of spark package if needed.

if neither kryo nor writable fits the bill, then additional thought is
required. I remember vaguely talking to Stephan and he told me that
supporting Writable serialization will suffice as a minimum requirement. So
serializing vectors with either of these two ways should be acceptable
regardless of actual implementation (view/non-view, doesn't matter. they
serialize any Vector implementation).

[1]
https://github.com/apache/mahout/blob/master/spark/src/main/scala/org/apache/mahout/sparkbindings/io/VectorKryoSerializer.scala

tillrohrmann:
I assume the reason why this has never been a problem for the spark and h2o bindings is exactly that you've registered (at least for the sparkbindings) a proper kryo serializer for the vector types. This is not done for the flink bindings.

Let me clarify a little bit how Flink serializes the collection of `MatrixVectorView`. Flink's type extractor detects that this type is not a POJO and thus assigns it a `GenericTypeInfo`. Types with this type info are serialized using Kryo. Since kryo has no special serializer registered it will use the default one. The default serializer in this case is the `FieldSerializer`. Thus, it will simply serialize all fields.

That's how the `CollectionInputFormat` can serialize the `MatrixVectorView`. Consequently, `CollectionInputFormat` can be serializable using Java serialization because it takes care of serializing its data collection.

However, I've just checked and noticed that Flink does not allow you to specify default serializers for Kryo. With that you can do the same as for the spark bindings. I'll fix this.

tillrohrmann:
Well, I just saw that you're using Flink `0.9-SNAPSHOT`. In `0.10-SNAPSHOT`, you can register default serializers for Kryo. I just tested to bump Flink's version to `0.10-SNAPSHOT` and it worked without problems. If you move `VectorKryoSerializer` and `GenericMatrixKryoSerializer` to a module which is accessible from the flink bindings module, then adding

```
env.addDefaultKryoSerializer(classOf[Vector], new VectorKryoSerializer())
env.addDefaultKryoSerializer(classOf[Matrix], new GenericMatrixKryoSerializer())
```

to the constructor of `FlinkDistributedContext` solved the OOM for me. Additionally, you have to make both serializers either serializable, otherwise Flink cannot ship them.

Alternatively, you can give both serializers a zero arg constructor (default values don't count a such a constructor for Java). Then the serializers don't have to be serializable and you can register the serializers via

```
env.addDefaultKryoSerializer(classOf[Vector], classOf[VectorKryoSerializer])
env.addDefaultKryoSerializer(classOf[Matrix], classOf[GenericMatrixKryoSerializer])
```

alexeygrigorev:
Thanks, Till! 
I could try to look into it during the weekend, but I'm not sure how it's going to work with 2 open PRs for the same code. The solution Till suggested doesn't look too complicated though...

smarthi:
Alexey,

Before we try out Till's suggested fixes, we are redoing some of the backend code to refactor common functionality out to a common module.

smarthi:
Fyi, We r running off of Flink 0.10-Snapshot

PR #162, opened 2015-10-21
MAHOUT-1775 FileNotFoundException caused by aborting the process of downloading Wikipedia dataset

psyclaudeZ:
Instead of putting mkdir, downloading dataset, and extracting inside one single if block, put them into three separate ones.


smarthi:
LGTM

smarthi:
Merged, fixes MAHOUT-1775

PR #164, opened 2015-10-23
MAHOUT-1778 move to run on Spark 1.5.1 and Hadoop 2.7.1

pferrel:


dlyubimov:
:100:

dlyubimov:
jira?

andrewpalumbo:
Do you wanna call this m-1778? That Jira is basically the upgrade to spark 1.5.

pferrel:
I've tested on Spark 1.5.1 and the shell gives the error below, the OOM is the one I worry about. 

Drivers seem to be fine, have run NB and Itemsimilarity but not the included tests.

```
Maclaurin:mahout pat$ mahout spark-shell
MAHOUT_LOCAL is not set; adding HADOOP_CONF_DIR to classpath.
15/10/28 15:02:51 WARN NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable

                         _                 _
         _ __ ___   __ _| |__   ___  _   _| |_
        | '_ ` _ \ / _` | '_ \ / _ \| | | | __|
        | | | | | | (_| | | | | (_) | |_| | |_
        |_| |_| |_|\__,_|_| |_|\___/ \__,_|\__|  version 0.11.0

      
Using Scala version 2.10.4 (Java HotSpot(TM) 64-Bit Server VM, Java 1.7.0_72)
Type in expressions to have them evaluated.
Type :help for more information.
15/10/28 15:02:56 WARN MetricsSystem: Using default name DAGScheduler for source because spark.app.id is not set.
Created spark context..
Spark context is available as "val sc".
Mahout distributed context is available as "implicit val sdc".
15/10/28 15:02:59 WARN General: Plugin (Bundle) "org.datanucleus.store.rdbms" is already registered. Ensure you dont have multiple JAR versions of the same plugin in the classpath. The URL "file:/Users/pat/spark/lib/datanucleus-rdbms-3.2.9.jar" is already registered, and you are trying to register an identical plugin located at URL "file:/usr/local/spark/lib/datanucleus-rdbms-3.2.9.jar."
15/10/28 15:02:59 WARN General: Plugin (Bundle) "org.datanucleus" is already registered. Ensure you dont have multiple JAR versions of the same plugin in the classpath. The URL "file:/Users/pat/spark/lib/datanucleus-core-3.2.10.jar" is already registered, and you are trying to register an identical plugin located at URL "file:/usr/local/spark/lib/datanucleus-core-3.2.10.jar."
15/10/28 15:02:59 WARN General: Plugin (Bundle) "org.datanucleus.api.jdo" is already registered. Ensure you dont have multiple JAR versions of the same plugin in the classpath. The URL "file:/Users/pat/spark/lib/datanucleus-api-jdo-3.2.6.jar" is already registered, and you are trying to register an identical plugin located at URL "file:/usr/local/spark/lib/datanucleus-api-jdo-3.2.6.jar."
Exception in thread "main" 
Exception: java.lang.OutOfMemoryError thrown from the UncaughtExceptionHandler in thread "main"

```

dlyubimov:
*I remember jokes about Microsoft (which doesn't really sound right now,
but was all the fun at the time): "Only 2 more Mb of RAM, and you would be
able to play the Mine sweeper game!".. Seems to me Spark has headed in this
tried and true MS direction*

dlyubimov:
> Do you wanna call this MAHOUT-1778? That Jira is basically the upgrade to spark 1.5.
Sure. why not. it's just discussion is not mirrorred anywhere if there's no Jira in the header.

buildtools/pom.xml
```diff
@@ -29,7 +29,7 @@
 
   <groupId>org.apache.mahout</groupId>
   <artifactId>mahout-buildtools</artifactId>
-  <version>0.11.1-SNAPSHOT</version>
+  <version>0.12.0-SNAPSHOT</version>
```
smarthi: Given that this is a Spark upgrade release + Bug fixes, lets keep the release at 0.11.1-Snapshot

spark-shell/src/main/scala/org/apache/mahout/sparkbindings/shell/MahoutSparkILoop.scala
```diff
@@ -157,7 +157,7 @@ class MahoutSparkILoop extends SparkILoop {
          _ __ ___   __ _| |__   ___  _   _| |_
         | '_ ` _ \ / _` | '_ \ / _ \| | | | __|
         | | | | | | (_| | | | | (_) | |_| | |_
-        |_| |_| |_|\__,_|_| |_|\___/ \__,_|\__|  version 0.11.0
+        |_| |_| |_|\__,_|_| |_|\___/ \__,_|\__|  version 0.11.1-SNAPSHOT
```
smarthi: Keep this as 0.11.1, this file is not modified by the release process. Its only the POMs that r updated from 'xxx-SNAPSHOT' to 'xxx' at the time of release. This one is part of the source code and needs to be updated manually. So lets have that as 0.11.1.

pom.xml
```diff
@@ -120,7 +120,7 @@
     <slf4j.version>1.7.10</slf4j.version>
```
smarthi: while at it, lets change this to the latest version = 1.7.12

bin/compute-classpath.sh
```diff
@@ -0,0 +1,167 @@
...
+#
+
+# This script computes Spark's classpath and prints it to stdout; it's used by both the "run"
+# script and the ExecutorRunner in standalone cluster mode.
+
+# Figure out where Spark is installed
+#FWDIR="$(cd "`dirname "$0"`"/..; pwd)"
+FWDIR="$SPARK_HOME"
+
+echo "Using FWDIR: $FWDIR"
```
andrewpalumbo: I think this should echo should comes out since it ends up in the path if not.

bin/compute-classpath.sh
```diff
@@ -0,0 +1,167 @@
...
+
+#. "$FWDIR"/bin/load-spark-env.sh # not executable by defult in $SPARK_HOME/bin
+
+"$MAHOUT_HOME"/bin/mahout-load-spark-env.sh
```
andrewpalumbo: The scala version doesn't seem to be set using the export in mahout-load-spark-env.sh:

When testing against s source build of spark 1.5.1:  ```$ compute-classpath.sh
Using FWDIR: /home/andy/apache/spark-1.5.1
Failed to find Spark assembly in /home/andy/apache/spark-1.5.1/assembly/target/scala-
You need to build Spark before running this program.
```
Setting the scala version manually, the classpath is computed correctly though.  Maybe we should enforce Mahout's scala version anyways?

bin/compute-classpath.sh
pferrel: was using a binary so missed that thanks

bin/compute-classpath.sh
pferrel: I need to install a build from source. 

Will make the Scala version a setting and hardcode for 0.11.1 since we haven't tried newer Scala afaik

pferrel: Shouldn't the code for the Scala version be put in compute-classpath ?

obviously I need to get the source version of Spark will move the Scala version check to compute-classpath before I've tested it on source build so be aware.

pferrel: after moving the Scala version check to compute-classpath.sh the Shell starts up without a glitch, no more warnings and no OOM. But it throw an exception on exit. BTW I built Spark 1.5.1 from source for Hadoop 2.4.1 (my version) compute-classpath seems to create the correct result but mahout-spark-class.sh can't be used independently. Have not tested the Scala 2.11 case--yikes this is getting complicated.

here is the exit error:

    mahout> :q
    Stopping spark context.
    15/10/30 09:29:49 ERROR TransportResponseHandler: Still have 1 requests outstanding when connection from 192.168.0.2/192.168.0.2:65385 is closed
    15/10/30 09:29:49 WARN NettyRpcEnv: Exception when sending RequestMessage(192.168.0.2:65376,NettyRpcEndpointRef(spark://Executor@192.168.0.2:65385),StopExecutor,false)
    java.io.IOException: Connection from 192.168.0.2/192.168.0.2:65385 closed
	at         org.apache.spark.network.client.TransportResponseHandler.channelUnregistered(TransportResponseHandler.java:104)
	at org.apache.spark.network.server.TransportChannelHandler.channelUnregistered(TransportChannelHandler.java:91)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:158)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:144)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelUnregistered(ChannelInboundHandlerAdapter.java:53)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:158)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:144)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelUnregistered(ChannelInboundHandlerAdapter.java:53)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:158)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:144)
	at io.netty.channel.ChannelInboundHandlerAdapter.channelUnregistered(ChannelInboundHandlerAdapter.java:53)
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelUnregistered(AbstractChannelHandlerContext.java:158)
	at io.netty.channel.AbstractChannelHandlerContext.fireChannelUnregistered(AbstractChannelHandlerContext.java:144)
	at io.netty.channel.DefaultChannelPipeline.fireChannelUnregistered(DefaultChannelPipeline.java:739)
	at io.netty.channel.AbstractChannel$AbstractUnsafe$8.run(AbstractChannel.java:659)
	at io.netty.util.concurrent.SingleThreadEventExecutor.runAllTasks(SingleThreadEventExecutor.java:357)
	at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:357)
	at io.netty.util.concurrent.SingleThreadEventExecutor$2.run(SingleThreadEventExecutor.java:111)

smarthi:
Ship it. Compiled against Spark 1.4.1 and tested against Spark 1.6-SnapShot, works cleanly.

pferrel:
Anyone else have a chance to check against maybe Spark 1.5.1 binary version, which has different classpaths? We'll also need to check against a binary Mahout for the same reason.

smarthi:
Tested with Spark 1.5.1 binary, works fine. Did see some warnings during Shell startup but nothing critical.

andrewpalumbo:
Tested shell with Spark-1.3 binaries np.  tested with Spark-1.5.1 source an binary distros and no problem shutting down.  Did get the datanucleus  warnings with 1.5.1- not sure if that's a configuration issue.

smarthi:
We have 3 +1s for this PR, committing this to 0.11.1 codebase.

PR #168, opened 2015-10-27
Spark 1.5.1 mahout-compute-classpath scripts

andrewpalumbo:
Pieced together scripts from Spark 1.3.1 compute-classpath.sh and  Spark 1.5.1 spark-class and load-spark-env scripts.

May be useful in conjunction with the `spark-submit`plan by @pferrel.  

Still needs some work- I'm not sure if we will want to use it or not.


bin/mahout-load-spark-env.sh
```diff
@@ -0,0 +1,59 @@
...
+# This script loads spark-env.sh if it exists, and ensures it is only loaded once.
+# spark-env.sh is loaded from SPARK_CONF_DIR if set, or within the current directory's
+# conf/ subdirectory.
+FWDIR="$SPARK_HOME"
+
+if [ -z "$SPARK_ENV_LOADED" ]; then
+  export SPARK_ENV_LOADED=1
+
+  # Returns the parent of the directory this script lives in.
+  parent_dir="$(cd "`dirname "$0"`"/..; pwd)"
```
andrewpalumbo: need to fix this to point to correct dir if this is gonna be used.

pferrel:
Some other projects create their own "compute-classpath.sh" and use it for constructing a command line for spark-submit. So I was heading in the same direction. If we construct the spark-submit cli we should not need to invoke the mahout script from scala since all jars will be provided in the spark-submit. This seems cleanest to me. 

I assume the mahout-load-spark-env.sh if for the shell? Does it work?

By "needs some work" what do you mean?

andrewpalumbo:
"mahout-load-spark-env.sh" is something from the /spark/bin directory.   I guess it just checks for the scala version of the spark build.  

by "There's still some work that needs to be done, I meant that this could be better implemented. It does seem to pick up all of the classes needed for the spark-trainnb drivers as is.  

the shell may be missing some classes though since it doesn't though it starts and has all basic mahout capabilities but gives the following error: 
```15/10/27 18:25:33 WARN ObjectStore: Failed to get database default, returning NoSuchObjectException```

re: picking up the mahout classes from scala, If we remove this feature then it would limit mahout use as an embedded library to only applications that are invoked with spark-submit.  

That's why I was thinking that it would be good to use both.  spark-submit and a compute-classpath script.

dlyubimov:
The only reason to call spark-classpath.sh from java was not to be
dependent on scala version (and perhaps packaging pecularities). Since they
now decided to do away with such dedicated script, there can no longer be a
goal of version independence.

which probably means we can pick the logic in the spark-class script
looking for the spark assembly jar, and copy-paste it where it is needed.
which is what i guess you are doing here.

pferrel:
This works for me. 

I had planned to remove the call to mahout's script to get the classpath and use spark-submit with all classes needed. 

This fixes the `mahout -spark classpath` so the scala code here now works: https://github.com/apache/mahout/blob/master/spark/src/main/scala/org/apache/mahout/sparkbindings/package.scala#L171

The only thing my method would do is remove this script call leaving it up to any launcher to assemble the classpath (in our case the mahout script would do it), which may be marginally cleaner. I don't like using Mahout to create a Spark context. This seems dangerous and is what spark-submit is for. But removing all the context creation stuff is more than i care to do so maybe this is all we need.

Since I've seen other projects calling this `compute-classpath.sh` is it ok to rename it? It seems to be the pattern people are falling into.

andrewpalumbo:
yeah- compute-classpath.sh sounds good.

pferrel:
ok I merged my 0.12.0-SNAPSHOT branch with this, will rename and create another PR.

andrewpalumbo:
Sounds good.

andrewpalumbo:
closing  new PR coming

PR #174, opened 2015-11-06
MAHOUT-1786: Make classes implements Serializable for Spark 1.5+

michellemay:
Add some "implements Serializable" for Apache Spark 1.5+
There might be other classes that would benefit from the same modification.

### DIFF
--- a/math/src/main/java/org/apache/mahout/math/AbstractMatrix.java
+++ b/math/src/main/java/org/apache/mahout/math/AbstractMatrix.java
@@ -24,13 +24,14 @@ import org.apache.mahout.math.flavor.MatrixFlavor;
 import org.apache.mahout.math.flavor.TraversingStructureEnum;
 import org.apache.mahout.math.function.*;
 
+import java.io.Serializable;
 import java.util.Iterator;
 import java.util.Map;
 
 /**
  * A few universal implementations of convenience functions for a JVM-backed matrix.
  */
-public abstract class AbstractMatrix implements Matrix {
+public abstract class AbstractMatrix implements Matrix, Serializable {
 
   protected Map<String, Integer> columnLabelBindings;
   protected Map<String, Integer> rowLabelBindings;
--- a/math/src/main/java/org/apache/mahout/math/AbstractVector.java
+++ b/math/src/main/java/org/apache/mahout/math/AbstractVector.java
@@ -17,6 +17,7 @@
 
 package org.apache.mahout.math;
 
+import java.io.Serializable;
 import java.util.Iterator;
 
 import com.google.common.base.Preconditions;
@@ -26,7 +27,7 @@ import org.apache.mahout.math.function.DoubleFunction;
 import org.apache.mahout.math.function.Functions;
 
 /** Implementations of generic capabilities like sum of elements and dot products */
-public abstract class AbstractVector implements Vector, LengthCachingVector {
+public abstract class AbstractVector implements Vector, LengthCachingVector, Serializable {
 
   private int size;
   protected double lengthSquared = -1.0;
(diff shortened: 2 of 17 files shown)


dlyubimov:
-1. Unfortunately, i think just adding "serializable" to every matrix class is not going to cut it. 

If custom java serialization were implemented a bit more efficiently (as it is currently done in Writeable and Kryo serializations), i would vote -0.1 (per apache voting guidelines). 

The reason is, java serialization is still not the best way to pack the tensor data. 

Another reason is there is no motivation in Mahout for java serialization support. All of our supported  backends support both Kryo and Writable protocols for the purpose of the backends. 

There are other minor reasons not to use java serialization as well (such as class compatibility checks etc.)

There admittedly may be an external reason but i feel like kryo serialzation should be an answer good enough for external reasons as well.

dlyubimov:
although i am not sure about efficiency point said in the original issue -- bytecode generated something in spark 1.5... i am dubious it will work well without custom serialization algorithms. needs a benchmark imo. there is still may be a significant difference between between serializing data structure vs. serializing data structure iterators and rebuilding a data structure.

michellemay:
Forcing the use of Kryo might not be a valid statu-quo for spark 1.5+ though..  

For reference, here is project Tungsten:
https://databricks.com/blog/2015/04/28/project-tungsten-bringing-spark-closer-to-bare-metal.html

"The above chart compares the performance of shuffling 8 million complex rows in one thread using the Kryo serializer and a code generated custom serializer. The code generated serializer exploits the fact that all rows in a single shuffle have the same schema and generates specialized code for that. This made the generated version over 2X faster to shuffle than the Kryo version."

michellemay:
I leave it to you to find a way to use that new codegen optimizations in spark.

dlyubimov:
Michel:

Just to be clear: can  you or can you not confirm  that this change

(1) is known to work correctly (as in all unit tests pass in a
java-serialized session); and

(2) that it is actually indeed faster than the same things in a kryo
session?

I think those are the main things to confirm in this issue.

michellemay:
I have a hard time testing vanilla 'master' on Windows..  (my dev environment)
I get tons of NPE (92 in total)  at __randomizedtesting.SeedInfo.seed, at java.lang.ProcessBuilder.start

michellemay:
I'm not 100% sure.. I still have other tests to do but it looks like default serializer of spark 1.5.1 is 19% SLOWER than kryo when performing large matrice AtA.

That is really sad since I see huge gains elsewhere in our spark tasks.

Is there any possibility to have best of both world ?

andrewmusselman:
I see this is old; @michellemay could you let us know if there's any room to progress?

dlyubimov:
Michel, I would like to revisit this issue but i need to understand whether there is indeed any merit to it. 

One thing i don't understand is why you think supporting java serialization has anything to do with enabling codegen serializer in spark? where is the information source that explicitly makes such guarantees? I couldn't immediately find any. 

All i can see is that it is used in catalyst, perhaps to serialize specific types there with keys and payloads (tuples?) but i don't seem to be able to understand how this may be applied to any non-dataframe pipeline or type. If you have any pointers, please let me know.

rawkintrevo:
@michellemay let me know if there is anything I can do to help push this forward. 

Thanks for your hard work!

rawkintrevo:
Follow on question- how does this eliminate Kryo?

dlyubimov:
@rawkintrevo i don't believe at this point this issue contains valid assertion w.r.t. being java serializable triggering codegen serializer. Nor  is there any experimental evidence of such. And there's spark 2.0.0 guidance to the contrary as i mentioned.

PR #175, opened 2015-11-06
MAHOUT-1785: Replace 'spark.kryoserializer.buffer.mb' from Spark conf…

smarthi:
…ig with 'spark.kryoserializer.buffer'

### DIFF
--- a/spark/src/main/scala/org/apache/mahout/drivers/MahoutSparkDriver.scala
+++ b/spark/src/main/scala/org/apache/mahout/drivers/MahoutSparkDriver.scala
@@ -75,7 +75,11 @@ abstract class MahoutSparkDriver extends MahoutDriver {
   override protected def start() : Unit = {
     if (!_useExistingContext) {
       sparkConf.set("spark.kryo.referenceTracking", "false")
-        .set("spark.kryoserializer.buffer.mb", "200")// this is default for Mahout optimizer, change it with -D option
+        .set("spark.kryoserializer.buffer.mb", "200m")// this is default for Mahout optimizer, change it with -D option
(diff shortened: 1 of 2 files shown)


spark/src/main/scala/org/apache/mahout/drivers/MahoutSparkDriver.scala
```diff
@@ -76,6 +76,10 @@ abstract class MahoutSparkDriver extends MahoutDriver {
     if (!_useExistingContext) {
       sparkConf.set("spark.kryo.referenceTracking", "false")
         .set("spark.kryoserializer.buffer.mb", "200")// this is default for Mahout optimizer, change it with -D option
+        // the previous has been marked deprecated as of Spark 1.4 by the below line,
+        // remove the above line when Spark finally retires above for below
+        .set("spark.kryoserializer.buffer", "200")
+
```
pferrel: I think you have to put the units in the string
   
    .set("spark.kryoserializer.buffer", "200m")

would be the equivalen iirc

spark/src/test/scala/org/apache/mahout/sparkbindings/test/DistributedSparkSuite.scala
```diff
@@ -40,6 +41,7 @@ trait DistributedSparkSuite extends DistributedMahoutSuite with LoggerConfigurat
       addMahoutJars = !isLocal,
       sparkConf = new SparkConf()
           .set("spark.kryoserializer.buffer.mb", "40")
+          .set("spark.kryoserializer.buffer", "40")
           .set("spark.akka.frameSize", "30")
```
pferrel: ditto above

smarthi:
Fixed both

PR #178, opened 2015-11-20
MAHOUT-1794: Support MAHOUT_WORK_DIR environment variable

chu11:
Support MAHOUT_WORK_DIR environment variable in examples to allow user to specify an alternate scratch/temp space instead of the defined one.  This will allow users in some environments without local drives or with small /tmp mounts to still run the examples.

### DIFF
--- a/examples/bin/classify-20newsgroups.sh
+++ b/examples/bin/classify-20newsgroups.sh
@@ -36,7 +36,11 @@ START_PATH=`pwd`
 # Set commands for dfs
 source ${START_PATH}/set-dfs-commands.sh
 
-WORK_DIR=/tmp/mahout-work-${USER}
+if [[ -z "$MAHOUT_WORK_DIR" ]]; then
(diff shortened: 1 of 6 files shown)


andrewmusselman:
Thanks for the addition, Al. The way this is written you will have to create a similar path on your local file system that will mirror the path you have in HDFS.

So for example, if you create a path in HDFS for your data, say '/user/al/mahout-tmp', and export it into MAHOUT_WORK_DIR, you will need to make the same path on the local file system at '/user/al/mahout-tmp'. Does that work for people do you think?

andrewmusselman:
Although cluster-syntheticcontrol.sh worked for both options, both default and specified path, and both local and remote, cluster-reuters.sh fails for option 1 in local mode with a specified path.

Al could you confirm you've tested the examples with all options, local and remote, and with default and specified path?

andrewmusselman:
False alarm, checked with a few scripts and this works fine.

PR #179, opened 2015-11-25
MAHOUT-1795: Build math & spark bindings under scala 2.11

mikekap:
The shell isn't building, so only enable it for 2.10.

Unfortunately this means that there's a 2x2 build matrix with hadoop/scala versions. The only way to resolve this
was to switch to activating profiles through properties, so if you were previously using -Phadoop1, you'll want
to use -Dhadoop1. The default config without any options is effectively -Phadoop2,scala-2.10.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -868,9 +868,6 @@
       <id>hadoop2</id>
       <activation>
         <activeByDefault>true</activeByDefault>
-        <property>
-          <name>env.JENKINS_URL</name>
-        </property>
       </activation>
       <properties>
         <hadoop.classifier>hadoop2</hadoop.classifier>
@@ -914,6 +911,26 @@
       <properties>
         <pmd.skip>true</pmd.skip>
         <checkstyle.skip>true</checkstyle.skip>
+      </properties>
+    </profile>
+    <profile>
+      <id>scala-2.10</id>
+      <activation>
+        <property><name>!scala-2.11</name></property>
+      </activation>
+      <properties>
+        <scala.compat.version>2.10</scala.compat.version>
+        <scala.version>2.10.4</scala.version>
+      </properties>
+    </profile>
+    <profile>
+      <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
+      <properties>
+        <scala.compat.version>2.11</scala.compat.version>
+        <scala.version>2.11.8</scala.version>
       </properties>
     </profile>
     <profile>
@@ -1031,7 +1048,6 @@
     <profile>
       <id>ci</id>
       <activation>
-        <activeByDefault>false</activeByDefault>
         <property>
           <name>env.JENKINS_URL</name>
         </property>
--- a/spark/src/main/assembly/dependency-reduced.xml
+++ b/spark/src/main/assembly/dependency-reduced.xml
@@ -38,12 +38,12 @@
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
-        <include>com.github.scopt</include>
+        <include>com.github.scopt_${scala.compat.version}</include>
         <include>com.tdunning:t-digest</include>
         <include>org.apache.commons:commons-math3</include>
         <include>it.unimi.dsi:fastutil</include>
-        <include>org.apache.mahout:mahout-native-viennacl_2.10</include>
-        <include>org.apache.mahout:mahout-native-viennacl-omp_2.10</include>
+        <include>org.apache.mahout:mahout-native-viennacl_${scala.compat.version}</include>
+        <include>org.apache.mahout:mahout-native-viennacl-omp_${scala.compat.version}</include>
         <include>org.bytedeco:javacpp</include>
       </includes>
     </dependencySet>


dlyubimov:
Ok.... what's not-so-terrible hack? where do we go from here?

mikekap:
Sorry, I was asked to open a pull request in the ticket. Given this, would there be interest in having a dual 2.10/2.11 build that excluded mahout-shell for 2.11?

mikekap:
Updated to not be a terrible hack, but formalizes excluding spark-shell from the 2.11 build. (If you look at the way spark itself tackles the 2.10/2.11 split - shell will need completely different code for 2.11.)

If this does get accepted, it would be nice to also set up CI for scala 2.11 (mvn -Dscala-2.11 ...) :)

dlyubimov:
Can somebody else to look this over -- we may want to include this if it truly enables 2.11 build

spark/src/main/scala/org/apache/mahout/sparkbindings/blas/ABt.scala
```diff
@@ -29,8 +29,6 @@ import org.apache.spark.SparkContext._
 import org.apache.mahout.math.drm.logical.OpABt
 import org.apache.mahout.logging._
 
-import scala.tools.nsc.io.Pickler.TildeDecorator
-
 /** Contains RDD plans for ABt operator */
```
dlyubimov: is this really there? oh. it is. good catch +1.

pom.xml
```diff
@@ -804,13 +798,15 @@
     <module>distribution</module>
     <module>math-scala</module>
     <module>spark</module>
-    <module>spark-shell</module>
```
dlyubimov: why are we dropping spark-shell from the build?

dlyubimov:
+1, preliminarily. still don't understand why this seems to exclude spark-shell from the build.

pom.xml
mikekap: Converting spark-shell is a nontrivial task that I don't have time to tackle. Spark itself has distinct shell code for 2.11 & 2.10. This change just formalizes excluding spark-shell from the 2.11 build; in a perfect universe the majority of this change wouldn't be needed.

andrewmusselman:
@mikekap can you explain why converting spark-shell for the 2.11 build is non-trivial, maybe give some more context?

mikekap:
I haven't looked too far in depth, but as I mentioned above, spark itself has completely separate (probably not API-compatible) source trees for spark [2.11](https://github.com/apache/spark/tree/master/repl/scala-2.11/src/main/scala/org/apache/spark/repl) and [2.10](https://github.com/apache/spark/tree/master/repl/scala-2.10/src/main/scala/org/apache/spark/repl). The code doesn't seem even remotely similar to me, and I likely don't have the expertise to convert that mahout shell code to work with 2.11.

Let me know if you'd like be to rebase this.

rawkintrevo:
Hey Mike- 

Sorry to let this thread go dead- can you rebase this?  Working on Spark 2.x and this becomes much more relevant. Also re: shell- I'm proposing dropping in favor of this approach.
https://github.com/apache/mahout/pull/271

Thanks for your contribution!

mikekap:
I'm happy to rebase this, but if you don't mind I'll wait until #271 is merged. It would be trivial to enable scala 2.11 if the shell is gone :)

rawkintrevo:
@mikekap #271 is merged.  Please re-base.  

Would _love_ to get this in for 0.13.0 if possible. 

Thank you again for the contribution!

rawkintrevo:
Bump- @mikekap can we get a rebase. Would like to merge this as soon as code-freeze for 0.13.0 is done.

andrewmusselman:
Now that 0.13.0 is out, is this still something we'd want to put in 0.13.1?

rawkintrevo:
hey @mikekap sorry I missed this- we changed the travis yml.  Can you rebase off of that.  This is failing on the viennaCL stuff, this is a known problem (scala 2.11 not working with viennaCL bindings).  The new travis.yml tests for spark 1.6/scala 2.10, spark 2.0/scala 2.11, spark 2.1/scala 2.11. 

But the scala 2.11's aren't tested with viennaCL. (only scala 2.10)

this should fix your build issue, and if you woundn't mind update the travis yml to utilize your profiles.

rawkintrevo:
bump

spark/src/main/assembly/dependency-reduced.xml
```diff
@@ -38,14 +38,14 @@
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
-        <include>com.github.scopt</include>
+        <include>com.github.scopt_${scala.compat.version}</include>
```
rawkintrevo: @pferrel @andrewpalumbo is this the thing ^^

pom.xml
```diff
@@ -917,6 +914,26 @@
...
+      </properties>
+    </profile>
+    <profile>
+      <id>scala-2.11</id>
+      <activation>
+        <property><name>scala-2.11</name></property>
+      </activation>
+      <properties>
+        <scala.compat.version>2.11</scala.compat.version>
+        <scala.version>2.11.8</scala.version>
```
rawkintrevo: We need to cap at 2.11.2 to maintain breeze compatibility. see https://github.com/apache/mahout/pull/326/files#diff-5d8d832c3c4016dd77a9c3d92a2c5b69

rawkintrevo:
going to merge this and fix outstanding issues with other PRs.

rawkintrevo:
@mikekap thanks for the contrib.  I went ahead an merged, and fixed up the rest.

PR #182, opened 2016-02-29
MAHOUT-1799: Read null row vectors from input file

statguy:


dlyubimov:
Noted, need Pat's review

pferrel:
Looks fine, I can't test it now. There may be cases where a row is null in the input but this treats those as non-error conditions so will not complain in logs. I'm a bit ambivalent about the need for this, can @statguy explain why a null row is not an error, might is be better to print a warning?

statguy:
@pferrel you are asking why null rows should not give an error? For example, when I find user-item recommendations, but not all users get recommendations. I write the matrix to an intermediate file and read it back to process it further to fill those null rows with some sort of average recommendations. I could write a unit test for this too, but I would like get an approval first.

smarthi:
@statguy Thanks for the PR, please also provide unit tests that can validate the problem and the fix.

statguy:
I have now provided the unit test.

pferrel:
OK, I'll have to hand merge this as per our process so it will be in the mainline for 0.13

PR #183, opened 2016-03-07
MAHOUT-1800: Pare down Classtag overuse

andrewpalumbo:
Currently, almost every operator requires an implicit parameter for the classtag context bound of drm rowset key type, even for things like drmA + drmB.

in reality though the DAG can already infer that similarly to e.g. it infers product geometry because classtags are already embedded in the logical plan.

for example, `classtag(drmA+drmB) == classtag(drmA) == classtag(drmB)`.

Not only does the DAG already contain this information, but also it opens doors to a loss of inference, since the optimizer doesn't verify that the new context bound is actually valid by retracing the inference. So any operation may introduce an invalid row key type, and as a consequence, invalid optimization information, without any further checks.

This patch does the following:
(1) eliminates ClassTag[K] context bound in majority of operations
(2) add keyClassTag:ClassTag[K] property getter to the DrmLike[K] trait itself
(3) ensures lazy inference of returned key parameter classtag via DAG inference.


smarthi:
+1 to merge this, tests pass locally

PR #185, opened 2016-03-08
MAHOUT-1802: Capture attached checkpoints (if cached)

andrewpalumbo:
Currently, the optimizer generates checkpoints and attaches them to actual logical elements of the DAG via CheckpointAction$cp. 

ie:

```
drmC = drmA+ drmB

val cp1 = drmC.checkpoint() // checkpoint
val cp2 = drmC.checkpoint() // cp2 == cp1

drmD = cp1 + drmE // cp1 + drmE
```

but, in:
`
drmD = drmC + drmE // computes drmA + drmB + drmC all over`

`drmC` already has` cp1` attached to it so we should assume the common computational path is the intent here regardless and should be used, instead of building plans that recompute it. That is,

`drmD = drmC + drmE` should imply `cp1 + drmE `as well even if checkpoint is not used explicitly.

This PR allows us to avoid excessive declarations like 

```
drmAcp = drmA.checkpoint

drmB = drmAcp %*%... 
```

and instead just use 

```
drmA.checkpoint()

drmB = drmA %*% ....
```


smarthi:
LGTM

PR #193, opened 2016-03-15
Mahout-1805 imjplement allreduceBlock in flink

andrewpalumbo:


andrewpalumbo:
Committed to flink-binding

PR #194, opened 2016-03-15
MAHOUT-1813: Functional "apply" DSL for distributed and in-memory matrices

andrewpalumbo:
We have functional "Assign" for in-memory matrices, e.g.:

```
    mxA := { x => x + 1 }
    mxA ::= { x=> x * 2 }
```

However, we lack similar unary elementwise function capability with distributed matrices, because distributed matrices are logically immutable.
The suggestion here is to use apply(func) to augment that capability for DRMs:

```
    drmA(x => x + 1)
    drmA(x => 2 * x)
```

Handling sparse/dense functional assignments: Second optional parameter, evalZeros, can be one of TRUE, FALSE, or AUTO (default AUTO).
It regulates treatement of 0 elements in matrices: whether we can skip assignments to 0 elements. AUTO (which is default) tests the
supplied function whether it changes 0 elements:

    evalZeroTest = f(0) != 0

Obviously, this will not work in all situations (e.g. non-deterministic functions), so we want to retain ability to explicitly override this
with `true` or `false`.

For completeness mostly, similar functional no-side-effect applies are implemented for in-memory types, e.g.:

```
    mxA(x => x + 1)
    mxA(x => 2 * x)
    mxA( (row, col, x) => x + 1)
```
etc.

### DIFF
--- a/math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala
+++ b/math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala
@@ -644,7 +644,7 @@ trait RLikeDrmOpsSuiteBase extends DistributedMahoutSuite with Matchers {
     val drmA = drmParallelize(mxA)
 
     (drmA(x => x + 1).collect - (mxAControl + 1)).norm should be < 1e-7
-    (drmA(x => x * 2).collect - (2 * mxAControl )).norm should be < 1e-7
+    (drmA(x => x * 2).collect - (2 * mxAControl)).norm should be < 1e-7
 
   }
 
--- a/math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala
+++ b/math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala
@@ -634,5 +634,19 @@ trait RLikeDrmOpsSuiteBase extends DistributedMahoutSuite with Matchers {
 
   }
 
+  test("functional apply()") {
+    val mxA = sparse (
+      (1 -> 3) :: (7 -> 7) :: Nil,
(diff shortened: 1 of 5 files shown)


math-scala/src/test/scala/org/apache/mahout/math/drm/RLikeDrmOpsSuiteBase.scala
```diff
@@ -634,5 +634,19 @@ trait RLikeDrmOpsSuiteBase extends DistributedMahoutSuite with Matchers {
...
+    val mxA = sparse (
+      (1 -> 3) :: (7 -> 7) :: Nil,
+      (4 -> 5) :: (5 -> 8) :: Nil
+    )
+
+    val mxAControl = mxA cloned
+    val drmA = drmParallelize(mxA)
+
+    (drmA(x => x + 1).collect - (mxAControl + 1)).norm should be < 1e-7
+    (drmA(x => x * 2).collect - (2 * mxAControl )).norm should be < 1e-7
```
dlyubimov: (2 * mxAControl ))

should be 

    (2 * mxAControl))

math-scala/src/test/scala/org/apache/mahout/math/scalabindings/MatrixOpsSuite.scala
```diff
@@ -131,6 +131,18 @@ class MatrixOpsSuite extends FunSuite with MahoutSuite {
...
 
+  test("functional apply()") {
+    val mxA = sparse (
+      (1 -> 3) :: (7 -> 7) :: Nil,
+      (4 -> 5) :: (5 -> 8) :: Nil
+    )
+    val mxAControl = mxA cloned
+
+    (mxA(x ⇒ x + 1) - (mxAControl + 1)).norm should be < 1e-7
+    (mxA(x ⇒ x * 2) - (2 * mxAControl )).norm should be < 1e-7
```
dlyubimov: same here as above, style spacing

dlyubimov:
+1 .

PR #197, opened 2016-03-17
MAHOUT-1815: dsqDist(X,Y) and dsqDist(X) failing in flink tests.

andrewpalumbo:
After taking the Very long way around trying to repartition, etc., it turns out that the row just needed to be properly re-keyed.

Tests pass now.

Though we may want to re-examine the implementation of FlinkOpAtB, as it seems pretty inefficient.


andrewpalumbo:
This is a small fix so if there are no objections I'm just gonna push it.

smarthi:
LGTM

andrewpalumbo:
thx

andrewpalumbo:
merged to flink-binding

PR #200, opened 2016-03-22
MAHOUT-1809: Bump JVM memory up to 4g for flink scalatests

andrewpalumbo:


andrewpalumbo:
This will allow `dspca` to pass in maven tests

andrewpalumbo:
Though 4g might be too much-- we may want to lower it before releasing.

PR #201, opened 2016-03-22
MAHOUT:1810 Use a method taken from FlinkMLTools for cache persistance

andrewpalumbo:
As a temporary measure, use this method to persist the dataset to the filesystem when caching rather that drmDfsRead()/Write.

Todo: 

1. Break up into  `persist` and  `readPersistedDataset` methods and only read a persisted dataset if it is already cached.
2. Use a property setting for the base dir.
3. Check to make sure that this method maintains parallelism deg for the dataset, if not set the new parallelism degree to the original

### DIFF
--- a/flink/src/main/scala/org/apache/mahout/flinkbindings/FlinkEngine.scala
+++ b/flink/src/main/scala/org/apache/mahout/flinkbindings/FlinkEngine.scala
@@ -357,12 +357,12 @@ object FlinkEngine extends DistributedEngine {
     res.collect().head
   }
 
-  private def generateTypeInformation[K: ClassTag]: TypeInformation[K] = {
+  def generateTypeInformation[K: ClassTag]: TypeInformation[K] = {
     val tag = implicitly[ClassTag[K]]
 
     generateTypeInformationFromTag(tag)
   }
-  
+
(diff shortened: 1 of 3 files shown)


flink/src/main/scala/org/apache/mahout/flinkbindings/FlinkEngine.scala
```diff
@@ -357,12 +357,12 @@ object FlinkEngine extends DistributedEngine {
     res.collect().head
   }
 
-  private def generateTypeInformation[K: ClassTag]: TypeInformation[K] = {
+  def generateTypeInformation[K: ClassTag]: TypeInformation[K] = {
     val tag = implicitly[ClassTag[K]]
```
dlyubimov: better (shorter) standard helper for this line is (I think)

    implicit val ktag = classTag[K]

andrewpalumbo: Thanks, I probably should have closed this since i already merged it,  I'll reopen another with for the list in the description. and make the above referenced fix.

andrewpalumbo:
Will reopen a new PR for this with a fresh branch.

PR #203, opened 2016-03-25
MAHOUT-1817  Implement caching in Flink Bindings

andrewpalumbo:
As a temporary measure, use this method to persist the `DataSet` to the filesystem when caching rather that drmDfsRead()/Write.

Todo:
 
1.  ~~Break up into `persist` and `readPersistedDataset` methods and only read a persisted dataset if it is already cached.~~
2.     ~~Use a property setting for the base dir~~.
3.     ~~Check to make sure that this method maintains parallelism deg for the dataset, if not set the new parallelism degree to the original~~

### DIFF
--- a/flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
+++ b/flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
@@ -18,7 +18,7 @@
  */
 package org.apache.mahout.flinkbindings.drm
 
-import org.apache.flink.api.common.functions.{MapFunction, ReduceFunction}
+import org.apache.flink.api.common.functions.{MapFunction, ReduceFunction, RichFunction, RichMapFunction}
 import org.apache.flink.api.common.typeinfo.TypeInformation
 import org.apache.flink.api.java.io.{TypeSerializerInputFormat, TypeSerializerOutputFormat}
 import org.apache.flink.api.scala._
@@ -58,7 +58,7 @@ class CheckpointedFlinkDrm[K: ClassTag:TypeInformation](val ds: DrmDataSet[K],
   var persistanceRootDir: String = _
 
   // need to make sure that this is actually getting the correct propertirs for {{taskmanager.tmp.dirs}}
-  val mahoutHome = System.getProperty("MAHOUT_HOME")
+  val mahoutHome = getMahoutHome()
 
   GlobalConfiguration.loadConfiguration(mahoutHome + "/conf/flink-config.yaml")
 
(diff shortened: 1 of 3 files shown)


smarthi:
lgtm

flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
```diff
@@ -76,20 +94,31 @@ class CheckpointedFlinkDrm[K: ClassTag:TypeInformation](val ds: DrmDataSet[K],
...
     }
-    implicit val typeInformation = createTypeInformation[(K,Vector)]
+    val _ds = readPersistedDataSet(cacheFileName, ds)
 
-    val _ds = persist(ds, persistanceRootDir + cacheFileName)
+    // We may want to look more closely at this:
+    // since we've cached a drm, triggering a computation
+    // it may not make sense to keep the same parallelism degree
+    if (!(parallelismDeg == _ds.getParallelism)) {
+      _ds.setParallelism(parallelismDeg).rebalance()
```
andrewpalumbo: not sure what to do here, this `rebalance` call seems to adding some time to the caching call.

smarthi: Is it needed? IIRC it was added when we had skewed partitions in previous flink releases.

andrewpalumbo: I guess my real question is whether or not to check and re-set the parallelism after  reading in from the cache:
```if (!(parallelismDeg == _ds.getParallelism)) {
      _ds.setParallelism(parallelismDeg).rebalance()
}
```

several operations may result in a much smaller matrix than the original DRM but it is not calculated until here.  Its really a case by case thing, so difficult to figure out.

andrewpalumbo: yes @smarthi  so I will I will likely remove this check tomorrow and then push the rest as is.   I guess we need to go through and make sure that all operations are setting the parallelism if they need to.

flink/src/main/scala/org/apache/mahout/flinkbindings/drm/CheckpointedFlinkDrm.scala
```diff
@@ -76,20 +94,38 @@ class CheckpointedFlinkDrm[K: ClassTag:TypeInformation](val ds: DrmDataSet[K],
...
+      * final drm after computation is actually triggered
+      *
+      *  // We may want to look more closely at this:
+      *  // since we've cached a drm, triggering a computation
+      *  // it may not make sense to keep the same parallelism degree
+      *  if (!(parallelismDeg == _ds.getParallelism)) {
+      *    _ds.setParallelism(parallelismDeg).rebalance()
+      *  }
+      *
+      */
```
andrewpalumbo: commented out getParallelism setting in cache() for now.  will push soon if no objections.

smarthi: In light of MAHOUT-1819, wherein its been agreed that parallelism would only be set inside MahoutFlinkContext for the ExecutionEnvironment, I think this jira can be safely marked as 'Resolved'.

PR #206, opened 2016-03-28
MAHOUT-1819: Set the default Parallelism for Flink execution in FlinkDistributedContext

smarthi:


flink/src/main/scala/org/apache/mahout/flinkbindings/FlinkEngine.scala
```diff
@@ -249,12 +249,10 @@ object FlinkEngine extends DistributedEngine {
...
     FlinkByteBCast.wrap(v)
 
-
   /** Broadcast support */
   override def drmBroadcast(m: Matrix)(implicit dc: DistributedContext): BCast[Matrix] =
     FlinkByteBCast.wrap(m)
 
-
   /** Parallelize in-core matrix as flink distributed matrix, using row ordinal indices as data set keys. */
   override def drmParallelizeWithRowIndices(m: Matrix, numPartitions: Int = 1)
```
andrewpalumbo: we should probably note in the comments that this breaks the drmParallelize() contract, and ignores the numPartitions parameter.

andrewpalumbo:
+1

flink/src/main/scala/org/apache/mahout/flinkbindings/FlinkEngine.scala
smarthi: yeah will do. the only reason we still have numPartitions as a parameter is because this method in its present form is more useful to Spark and H2O, but not in Flink.

PR #213, opened 2016-04-09
MAHOUT-1821: Use a mahout-flink-conf.yaml configuration file for Maho…

smarthi:
…ut specific Flink configuration

### DIFF
--- a/conf/flink-config.yaml
+++ b/conf/flink-config.yaml
@@ -21,49 +21,13 @@
 # Common
 #==============================================================================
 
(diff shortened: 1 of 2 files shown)


andrewpalumbo:
I think that we should only set the properties that we need for our stuff.  This will override any properties set by the user in their `flink-config.yaml` file.

andrewpalumbo:
+1

smarthi:
Closing this PR, merged to 'flink-binding'

PR #226, opened 2016-04-20
(nojira) enable flink tests

andrewpalumbo:


flink/src/test/scala/org/apache/mahout/flinkbindings/DistributedFlinkSuite.scala
```diff
@@ -58,15 +59,15 @@ trait DistributedFlinkSuite extends DistributedMahoutSuite { this: Suite =>
...
-      1,
-      parallelism,
-      false,
-      false,
-      true)
+    GlobalConfiguration.loadConfiguration(getMahoutHome() + "/conf/flink-config.yaml")
+
+    val conf = GlobalConfiguration.getConfiguration
+    
+    val cl = TestBaseUtils.startCluster(conf, false)
```
andrewpalumbo: @smarthi I'm not sure if the mini-cluster is picking up the `flink-config.yaml` settings.  I was trying to get Flink to test with Travis the other night, and  kept hitting this 

``` Cause: java.io.IOException: Insufficient number of network buffers: required 32, but only 0 available. The total number of network buffers is currently set to 2048. You can increase this number by setting the configuration key 'taskmanager.network.numberOfBuffers'.```

error even with 

```taskmanager.network.numberOfBuffers: 4096``` in the  `flink-config.yaml` file.

may be another Flink bug?

PR #230, opened 2016-04-30
MAHOUT-1838: simple 3d plots

andrewpalumbo:
Simple first cut of 3d plots and surfaces. Surfaces are a bit more complicated in that data may need to be ordered.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/visualization/MahoutPlot.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/visualization/MahoutPlot.scala
@@ -31,14 +31,16 @@ import smile.plot.{PlotCanvas, PlotPanel}
 trait MahoutPlot  {
 
   var canvas : PlotCanvas = _
-  var plotPanel: PlotPanel =_
+  var plotPanel: PlotPanel = _
   var plotFrame: JFrame = _
   var mPlotMatrix: Matrix = _
-  def contentPane = plotPanel
+  def contentPane = canvas
 
   // export a PNG of the plot to /tmp/test.png
   def exportPNG(path: String ="/tmp/test.png") = {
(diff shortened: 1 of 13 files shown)


andrewpalumbo:
plot3drotated
plot3d

andrewpalumbo:
2dhisto_20buckets_gaussian
2dhisto_21buckets_100pct_gaussian
2dhisto_10buckets_10pct_gaussian

andrewpalumbo:
3d_histo
3d_histo_rotated
3d_histo_rotated_5000

math-scala/src/main/scala/org/apache/mahout/visualization/MahoutPlot.scala
```diff
@@ -18,7 +18,36 @@
...
+      new BufferedImage(contentPane.getWidth, contentPane.getHeight, BufferedImage.TYPE_INT_ARGB)
+
+    val g2d: Graphics2D = bi.createGraphics
+
+    contentPane.printAll(g2d)
+
+    val file: File = new File(path)
+
+    ImageIO.write(bi, "PNG", file)
+  }
```
andrewpalumbo: @pferrel Setup to easily export plot PNGs to the Filesystem.  Maybe we can we can pipe these to the Zeppelin server via JSON?

andrewpalumbo:
For the most part this is good to go.  Still bare-bones and functionality needs to be built in (eg. adding titles, plotting multiple datasets, plotting decision boundaries, etc). Much will depend on the direction we take this in.

The 3d surface and mesh functions need some work.  

This PR ended up covering 5 or so subtasks of MAHOUT-1838.

If there are no Objections I'll probably push it tomorrow.

PR #231, opened 2016-05-01
MAHOUT-1841 change UNIFORM_DIVISOR to 2^63 from 2^64

andrewpalumbo:
Matrices.SymmetricUniformView is returning numbers on the range of (-.5,.5).  Documentation calls for [1,1).

### DIFF
--- a/math/src/main/java/org/apache/mahout/math/Matrices.java
+++ b/math/src/main/java/org/apache/mahout/math/Matrices.java
@@ -127,7 +127,7 @@ public final class Matrices {
     return gaussianGF;
   }
 
-  private static final double UNIFORM_DIVISOR = Math.pow(2.0, 64);
+  private static final double UNIFORM_DIVISOR = Math.pow(2.0, 63);
 
   /**
    * Uniform [-1,1) matrix generator function.


andrewpalumbo:
@smarthi this seems to pass all tests for me

smarthi:
Very well I'll merge the name to master

smarthi:
I can't get any of the dals (H2O, Spark or Flink) to pass on Mac OpenJDK with this fix, not sure we wanna push this until we have others confirm on both Mac and Linux.

andrewpalumbo:
Sounds good.

smarthi:
So it does indeed pass on Fedora Linux with OpenJDK 8, but fails on Mac OSX.

smarthi:
The dals failures seem local to my Mac and its failing now even without the change, I guess this PR is good to go, merging it.

smarthi:
I don't think this fixed it, Jenkins broken post this fix. Its best to change the javadocs to read [-0.5, 0.5)

andrewpalumbo:
Yeah it looks like we should revert for now.

andrewpalumbo:
It failed in travis also... hadn't noticed that:
```
- dals *** FAILED ***
  9043.008205882092 was not less than 0.15 (DistributedDecompositionsSuiteBase.scala:215)
```

andrewpalumbo:
@dlyubimov do you have any thoughts on why this might fail for `dals` only and only on certain java distributions?

andrewpalumbo:
reverted in master by 0d6c4e02409a6f8da776d00dc661fd178a7630b2

andrewmusselman:
Should this stay open or be closed?

smarthi:
Close it, too risky a fix and it's not easy to reproduce

PR #236, opened 2016-05-27
MAHOUT-1864: Twenty Newsgroups Classification Example fails in case running with MAHOUT_LOCAL=true

girirajsharma:
Twenty Newsgroups Classification Example fails in case running with `MAHOUT_LOCAL=true` or else when `HADOOP_HOME` env variable is not set.

[Newsgroups](https://mahout.apache.org/users/classification/twenty-newsgroups.html) lists instructions in order to run this classifier. When running in standalone mode(`MAHOUT_LOCAL=true`), i.e., running `$ ./examples/bin/classify-20newsgroups.sh`, the script runs `./examples/bin/set-dfs-commands.sh` internally to export hadoop related env variables.

`set-dfs-commands.sh` attempts to check for hadoop version despite running with `MAHOUT_LOCAL` set as true. IMHO, the script works fine considering the prerequisites, but, it will as well make sense if we can update the script `./examples/bin/set-dfs-commands.sh` to export hadoop env varibales only in case `MAHOUT_LOCAL` is not set to true.

### DIFF
--- a/examples/bin/set-dfs-commands.sh
+++ b/examples/bin/set-dfs-commands.sh
@@ -24,31 +24,32 @@
 # Run by each example script.
 
 # Find a hadoop shell
-if [ "$HADOOP_HOME" != "" ] && [ "$MAHOUT_LOCAL" == "" ] ; then
+if [ "${HADOOP_HOME}" != "" ] && [ "${MAHOUT_LOCAL}" == "" ] ; then
   HADOOP="${HADOOP_HOME}/bin/hadoop"
   if [ ! -e $HADOOP ]; then
     echo "Can't find hadoop in $HADOOP, exiting"
     exit 1
   fi
+
+  # Check Hadoop version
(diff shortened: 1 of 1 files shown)


andrewpalumbo:
Thank you for the patch, @girirajsharma.  As you point out, `HADOOP_HOME` is a requisite in both local and distributed mode.  So they are in fact working properly.  IMO It would be best not to overcomplicate these scripts.  

Mahout has actually phased out MapReduce over the past two years (see MAHOUT-1510) and now uses the Mahout "Samsara" framework: http://mahout.apache.org/users/sparkbindings/home.html.  Mahout is not accepting new MapReduce features.  Thank you.

andrewmusselman:
Let's close this; thanks Giri

rawkintrevo:
@andrewpalumbo @andrewmusselman this was marked as resolved in JIRA
https://issues.apache.org/jira/browse/MAHOUT-1864

But it does not appear to have actually been merged- are you ok with me merging now and updating jira to reflect 0.13.1?

rawkintrevo:
I'm a dummy- it's marked as 'Resolved: Won't Fix'
never mind.

PR #237, opened 2016-05-29
MAHOUT-1866: Add matrix-to-tsv string function

smarthi:


math-scala/src/main/scala/org/apache/mahout/math/drm/package.scala
```diff
@@ -148,6 +148,42 @@ package object drm {
...
+  /**
+    * Convert a sampled DRM into a Tab Separated Vector (TSV) to be loaded into an R-DataFrame
+    * for plotting and sketching
+    * @param drmX - DRM
+    * @param samplePercent - Percentage of Sample elements from the DRM to be fished out for plotting
+    * @tparam K
+    * @return TSV String
+    */
+  def sampleMatrixToTSV[K](drmX: DrmLike[K], samplePercent: Double = 1): String = {
+
```
andrewpalumbo: Minor point: maybe rename to `drmSampleToTSV` or something along those lines so that it is obvious that its a DRM and not a Matrix?  Other than that +1 from me.

PR #239, opened 2016-06-06
Fix assignment of k12 and k21

mbrio:
The values for these variables were incorrectly swapped according
to the description of the algorithm at:

http://tdunning.blogspot.de/2008/03/surprise-and-coincidence.html

According to the table, `k12` defines a variable that reflects the count of interactions with B but not A; the way it was defined in the source code was the interactions with A but not B, which should be the value of `k21`.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/cf/SimilarityAnalysis.scala
@@ -203,8 +203,8 @@ object SimilarityAnalysis extends Serializable {
     numInteractionsWithAandB: Long, numInteractions: Long) = {
 
     val k11 = numInteractionsWithAandB
-    val k12 = numInteractionsWithA - numInteractionsWithAandB
-    val k21 = numInteractionsWithB - numInteractionsWithAandB
+    val k12 = numInteractionsWithB - numInteractionsWithAandB
+    val k21 = numInteractionsWithA - numInteractionsWithAandB
     val k22 = numInteractions - numInteractionsWithA - numInteractionsWithB + numInteractionsWithAandB
 
     LogLikelihood.logLikelihoodRatio(k11, k12, k21, k22)


dlyubimov:
> The values for these variables were incorrectly swapped accordin

I highly doubt that. The LLR code has been ironed out for generations. 

 @tdunning  could explain it most accurately why.

dlyubimov:
actually this is a newer code, should be delegated to @pferrel or @ssc

tdunning:
I doubt that this matters since LLR(K) = LLR(K') for the 2x2 case that we
are talking about.

mbrio:
@tdunning I won't argue against that, I've done the tests and I can see that the swapping of those values has very limited effect on the outcome of the algorithm. I guess my issue with it is, when a developer, who may not have as much experience as your team has with this algorithm, looks at the source code, then reads your article that's linked to within the comments, they may question whether the algorithm has been implemented correctly. That question may lead developers to have less trust in the quality of the Mahout project; which we both know is false, because the algorithms within this package are solid, proven algorithms.

tdunning:
Well, I guess that we have an existence proof that there is at least one person confused.

But what about a putative next time when we are comparing Q versus R instead of A versus B? The use of variable names like A and B to refer to events that we don't even really have yet doesn't really matter. Nor does swapping of names around.

The point here is that there is an important level of abstraction that makes strict adherence to previous naming a bit of a trivial issue.

That said, making the change will not matter. I tend to not like code changes on this level and for this kind of reason, but it should be benign.

mbrio:
I think if the variables were not following the naming scheme put forth in your article there would not be confusion, such as your example of using Q and R. It's fine, I won't push further with this, it's a change that only has an effect on the clarity of the code that's written as opposed to the functionality. Admittedly that clarity is confused only when compared to the article that's linked to, not when compared to the rest of the code within Mahout.

sscdotopen:
I think we can safely close this.

tdunning:
Agree

PR #242, opened 2016-06-13
(NoJIra) re-apply Vienna math scala 

andrewpalumbo:
Re-apply the base `vienna` math module after its removal for 0.12.2 release.  This patch includes base directory, `pom.xml`, headers and simple passing hello world tests in both `vienna` and math-scala dirs.


dlyubimov:
so the decision is to take and include all VCL code in the build?

andrewpalumbo:
Yes I'd thought it would be be easiest originally, in order to preserve the tree structure and since I didn't know yet didn't know yet exactly what we classes we were going to be using.  @smarthi had concerns about this being too much code also.  And looking closer the sources are much larger than I'd thought. I'd thought that I'd commit the whole thing first and then remove what wasn't used, but in the long run I guess its better to add them only as they are they are needed.

dlyubimov:
wouldn't it be perhaps a better strategy to ask user to have their own
build of VCL? VCL is under development, it will morph a lot, we will have
to support a fixed release of it?

PR #245, opened 2016-07-07
Mahout 1869

skanjila:
Added the ability to dump output to csv file


andrewmusselman:
Looks like a lot of extras in here; what is the intended functionality of this, running some perf tests and writing out to CSV?

The akka/spray additions seem out of scope if so, and there is a lot of extra stuff in the pom that could be inherited from the parent. Also, the CSV library from a github repo will have to be justified, otherwise there are existing libraries that would be preferable.

Please follow the convention from existing files regarding preamble comment containing licensing.

Thanks!

skanjila:
Here's the intended functionality:

For each type of algorithm in mahout

     run a set of perf tests with different data sets and different hyperparameters

     connect results to zeppelin for visual verification/comparison of results

End


As I mentioned earlier ideally this becomes part of any build that gets kicked out during the release process, ,  I will take the spray and akka functionality out, that was there to make this a rest based service that one could call if necessary.   I can also replace the csv library, what are the suggestions for the different libraries here?  Also any thoughts about the general approach, if not I will start incorporating into zeppelin as well as hooking it into the other algorithms.  Currently I am launching this as a separate module within mahout, is this workable for folks, the module will live parallel to math/samsara etc.  One last thing, are there a preferred set of algorithms that this should be hooked to first (trying to figure out if we could prioritize certain algorithms over others?).


I will also add the preamble comment about licensing to files and attach a pull request for your review.


Thanks



________________________________

perf/pom.xml
```diff
@@ -0,0 +1,247 @@
...
+
+    <dependency>
+      <groupId>com.github.tototoshi</groupId>
+      <artifactId>scala-csv_2.11</artifactId>
+      <version>1.3.0</version>
+    </dependency>
+
+
+    <!-- https://mvnrepository.com/artifact/org.apache.zeppelin/zeppelin-interpreter -->
+    <dependency>
```
andrewpalumbo: We are not going to want to bring in the Zeppelin interpreter to the Mahout tree for this.  This version is relatively small but the latest version is 6MB.

PR #246, opened 2016-08-04
[MAHOUT-1856][WIP] create a framework for new Mahout Clustering, Classification, and Optimization Algorithms 

rawkintrevo:
Relevant JIRA: [https://issues.apache.org/jira/browse/MAHOUT-1856](https://issues.apache.org/jira/browse/MAHOUT-1856)

Readme.md provides a more comprehensive (yet still incomplete) overview.

Key Points:
Top Level Class: 
Model has one method- fit, and coefs.

Transformers map a vector input to a vector output (same or different length)
Regressors map a vector input to a single output (e.g. a Double)
Classifiers extend Transformers which have created a probability vector by 'selecting' the class and returning the label (instead of the entire p-vector)

Pipelines and Ensembles are models as well, except they are composed from other models listed above, or from other pipelines and ensembles.

ToDo:
- [ ] All models need a uniform way to expose their tuning parameters -> this will be required for a auto-tuning algo.  
- [ ] Pipelines / Ensembles must be able to account and report the tunable paremeters of their sub models
- [ ] Need fitness functions
- [ ] Native method wrappers- Underlying engines and third party packages have implementations of many ML models, let's not recreate the wheel by exposing YET ANOTHER sgd algorithm. Instead should be able to convert matrix to expected format of 'other' library, run model, get results, package back into matrix and pass on in pipeline or ensemble. (This is especially useful for DeepLearning4J integration). Also Native implementations on engine of some algos probably more efficient by leveraging engine specific tricks (think Flink delta iterators) than implementations we would make. 
- [ ] Lots more, open for discussion. 

This is merely a conversation started on what to do.  

I've included OLS as an example regressor and a normalizer as an example transformer, only for illustrative purposes.  I really don't want to pack to many algos in to this initial commit, just an example/ proof of concept so we can say, yea- this framework makes sense for this kind of model OR ooh, we probably want to have these features too.


andrewpalumbo:
Thanks @rawkintrevo.. This is a great start.  I'd originally thought that we'd put the algos in the math-scala module, but looking at it,  I think this makes sense.

andrewpalumbo:
Nice work on the slides also! :100: 

Another thought that I had is that we may want to allow In-Core matrices as parameters.  Just throwing it out there for discuassion.  I cant think of a particular use case off the top of my head but It seems that there should be.

rawkintrevo:
This is WIP so it doesn't really matter that it's failing atm but [this](https://travis-ci.org/apache/mahout/builds/180124934#L176) isn't good.. Couldn't find maven on the `wget` command.

andrewpalumbo:
@dlyubimov, @mahout-team could you review/provide feedback on this?  Originally Trevor had a separate module for this, and I asked him to move it into math-scala.

andrewpalumbo:
I should note that my above comment was meant as a question to all concerned: "Does this (the algorithms base traits package) Belong in math-scala or in its own module... Trevor initially had it in its own module, I asked that he move it to math-scala, due in most part to size concerns of the entire binary distribution.. however I am now questioning whether that is the correct call.. I can think of several good reasons that it should have its own module, (and others that it should be in math-scala).. For now, I think that math-scala is an ok place, and we can just move it to another module if we want later.  @smarthi @smarthi @rawkintrevo @dlyubimov please weigh in here.  Not a big deal right now, but thinking ahead it may be good to get the location right.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
```diff
@@ -0,0 +1,33 @@
...
+
+import org.apache.mahout.math.algorithms.Model
+import org.apache.mahout.math.drm.DrmLike
+
+/**
+  * Abstract of Regressors
+  */
+abstract class Regressor extends Model {
+
+  def fit[Int](drmY: DrmLike[Int], drmX: DrmLike[Int]): Unit
```
dlyubimov: also : i am not sure fitter should extend a model. Rather, fitter should return a model, i.e.,  
`fit[k](...) : Model`, right? i think that's how this pattern goes in most kits. Fitter is just a startegy.

And i'd abstain from doing abstract classes in Scala, unless trait absolutely cannot do it. (and it can in this case). Abstract class points to a specific, single and necessary base implementation in hirerarchy, which is too constraining without need for the actual implementations.

rawkintrevo: +1 on swapping Y and X- I've had to catch myself more than once on that already.  I think the original motivation was a tip of the hat to R's `Y ~ x` but I agree with you.

Re `[Int]` I realized that later, but haven't gone through to swap them all back to `K`.  It is (or was) `K` in some places.

rawkintrevo:
Thanks for the quick review @dlyubimov will incorporate your sugguestions

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
dlyubimov: i guess if this is abstract enough, we also need to be able admit hyperparameters which are of course specific for every fitter. in R this is trivial (any call can be made a bag of whatever named arguments), but in Scala this may need a bit of a thought (if this abstraction needs to be that high). otherwise, i guess most scala kits just create a concrete fit signature per implementation.

if the Regressor trait is meant to be common to all possible regression class algorithms, we either need a way to universally pass in the hyperparameters, or just not have fit abstraction in the regressor trait at all . (then what i guess :) )

dlyubimov: like perhaps `fit[K](X:drmLike[K],T:drmLike[K], (Symbol,Any)*):Model` where the optional list is hyperparameter list. Then hyperparameterized calls could be something like: ```
fit(X,Y, 'alpha ->10.0, 'lambda -> 1.e-15)
``` etc.
This loses strong type-iness of the signature but call is not that ugly, and it might be ok if specific implementations are cleanly documented .

File: math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala [hunk 2]
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.mahout.math.drm.DrmLike
+
+/**
+  * Abstract of Regressors
+  */
+trait Regressor[K] extends Model {
+
+  var residuals: DrmLike[K] = _
+
+  def fit(drmFeatures: DrmLike[K], drmTarget: DrmLike[K]): Unit
```
andrewpalumbo: As well, `drmFeatures` is somewhat confusing,  with the  Drm being a matrix of samples x features, i.e. features are columns,  I think something like `drmSamples`, or `drmObservations`, or even `drmX` may be more straightforward.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/OrdinaryLeastSquares.scala
```diff
@@ -0,0 +1,96 @@
...
+  * 
+  */
+class OrdinaryLeastSquares[K](hyperparameters: Map[String, Any] = Map("" -> None)) extends LinearRegressor[K] {
+  // https://en.wikipedia.org/wiki/Ordinary_least_squares
+
+  var calcStandardErrors: Boolean = hyperparameters.asInstanceOf[Map[String, Boolean]].getOrElse("calcStandardErrors", true)
+  var addIntercept: Boolean = hyperparameters.asInstanceOf[Map[String, Boolean]].getOrElse("addIntercept", true)
+
+  var summary = ""
+  def fit(drmFeatures: DrmLike[K], drmTarget: DrmLike[K]): Unit = {
```
andrewpalumbo: Continuing on from our discussion on Slack I would think that Fit may be a more appropriate place for Hyperparameters eg:
```
fit(observed_independent: Drm[K], observerd_targets: Drm[K],  hyperparamters: Option[List[double]]): List[double]
```
I think that this may be a matter of Convention, so If you're following a convention that I am not familiar with, this may be fine.  However I feel that this may be more robust.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/AutocorrelationTests.scala
```diff
@@ -0,0 +1,54 @@
...
+      If d < dL,α, there is statistical evidence that the error terms are positively autocorrelated.
+      If d > dU,α, there is no statistical evidence that the error terms are positively autocorrelated.
+      If dL,α < d < dU,α, the test is inconclusive.
+
+      Rule of Thumb:
+       d < 2 : positive auto-correlation
+       d = 2 : no auto-correlation
+       d > 2 : negative auto-correlation
+  */
+  def DurbinWaton[K](model: Regressor[K]): Regressor[K] = {
```
andrewpalumbo: misspelling- "DurbinWatson"

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/AsFactor.scala
```diff
@@ -0,0 +1,79 @@
...
+
+class AsFactor extends Transformer{
+
+  var factorMap: MahoutVector = _
+  var k: MahoutVector = _
+  var summary = ""
+
+  def transform[K: ClassTag](input: DrmLike[K]): DrmLike[K] ={
+    if (!isFit) {
+      //throw an error
```
andrewpalumbo: Is this complete? I.e., do you want to throw an error here for this release?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/MeanCenter.scala
```diff
@@ -0,0 +1,93 @@
...
+        copy.foreach(row => row -= bcastV.value)
+        (keys, copy)
+    }
+    output
+  }
+
+  def invTransform[K: ClassTag](input: DrmLike[K]): DrmLike[K] = {
+
+    if (!isFit) {
+      //throw an error
```
andrewpalumbo: Same qestion here as before- do you want to throw an error here?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/StandardScaler.scala
```diff
@@ -0,0 +1,119 @@
...
+    variance = mNv._2
+    stdev = mNv._2.sqrt
+    isFit = true
+
+  }
+
+  def transform[K: ClassTag](input: DrmLike[K]): DrmLike[K] = {
+
+    if (!isFit) {
+      //throw an error
```
andrewpalumbo: I find this a bit confusing (as well as the other mentions of 'if (!isFit) {// throw }.  So if i want to standardize a Drm to (mean = 0,std_dev = 1),    I would need to do something like:
```
drmStandardized = StandardScaler(unscaledDrm).fit().transform()
```
?

andrewpalumbo: As well, I think that this could be another argument for moving  Hyperparamaters into `fit(...)`, e.g. If for some reason we wanted to standardize on N(mean = 0, stdDev = 2),  we could still call `StandardScaler` and `fit (Map["mu" -> 0, "sigma" ->2])`:
```
val drmStandardized = StandardScaler(unscaledDrm).fit(Map["mu" -> 0, "sigma" ->2]).transform()
```

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionSuite.scala
```diff
@@ -0,0 +1,82 @@
...
+    val model = new OrdinaryLeastSquares[Int]()
+    model.fit(drmY, drmX)
+    val estimate = model.beta
+    val Ranswers = dvec(-1.336265, -13.157702, -4.152654, -5.679908, 163.179329)
+
+    val epsilon = 1E-6
+    (estimate - Ranswers).sum should be < epsilon
+
+  }
+
```
andrewpalumbo: It would be good to have a couple of more tests here; at least one for `Transform(...)`

andrewpalumbo:
Trevor- looks really good to me- I've left some comments mainly about hyperparameter being moved to fit(...) from model, I think that his makes sense in many ways, E.g, When doing an highly iterative Hyperparameter search, It would eliminate a good amount of overhead to call:

  ```aModel.fit(....,HyperParameters: Map["hParameter1" -> value , "hParameter2" -> ...])``` 

 rather than re-constructing the entire class each time.  As well as i noted in line, I think that the `fit(...)` method should have the ability to return at least a `List[double]` of errors per row if needed, So I would suggest that it return `Any` rather than Unit in the base Traits. (unless the convention that you're following is to rely on predict for this.

andrewpalumbo:
I want to emphasize though that if you're building up to follow a convention (sk-learn-like, MlLib etc) which I am not familiar with That may e better to follow than my suggestions to make this framework as easy on new users as possible.

rawkintrevo:
Thanks for the review @andrewpalumbo !

[sklearn parameters are set when the `Estimator` is instantiated.](http://scikit-learn.org/stable/tutorial/statistical_inference/settings.html). 

MLlib on the other hand, [passes parameter maps in `fit` as you suggest](https://spark.apache.org/docs/2.0.1/api/java/org/apache/spark/ml/Estimator.html#fit(org.apache.spark.sql.Dataset,%20org.apache.spark.ml.param.ParamMap))

BOTH however, allow hyper parameters to be updated.  And in the case you refer too, the model would not be re-instantiated, but something like this:

```scala
model.param1 = 1
model.fit(X, y) 
model.param2 = 2
```

To your point, I also want to make this as easy as possible for new users- so I think it would be best to leave the option to pass a parameter map at initiation, and also expose it as a optional parameter of the `fit` method.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
rawkintrevo: by errors- do you mean like code errors, or the residuals?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/StandardScaler.scala
rawkintrevo: correct and agreed- sklearn convention is to also have a `fitTransform` method to do it in one step.  

will add that.

andrewpalumbo:
@rawkintrevo Great- best of both worlds then!

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
andrewpalumbo: yes eg. residuals, a hessian matrix, etc, anything that a user  who is developing their own algorithm might like to see returned from their own `fit(..)` method.  My point was No reason to limit the return value to a `Unit`, rather I'd think `Any` would be a more appropriate return value from the base trait. Though maybe This would not work with the pipeline structure that you're setting up?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/CochraneOrcutt.scala
```diff
@@ -0,0 +1,83 @@
...
+
+  var betas: Array[MahoutVector] = _
+
+  var summary = ""
+
+  def fit(drmFeatures: DrmLike[K],
+          drmTarget: DrmLike[K],
+          hyperparameters: Map[String, Any] = Map("" -> None)): Unit = {
+
+    var hyperparameters: Option[Map[String,Any]] = None
```
rawkintrevo: there should be a `setHyperparameters` right about here...

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/AutocorrelationTests.scala
```diff
@@ -0,0 +1,54 @@
...
+      If d > dU,α, there is no statistical evidence that the error terms are positively autocorrelated.
+      If dL,α < d < dU,α, the test is inconclusive.
+
+      Rule of Thumb:
+       d < 2 : positive auto-correlation
+       d = 2 : no auto-correlation
+       d > 2 : negative auto-correlation
+  */
+  def DurbinWatson[K](model: Regressor[K]): Regressor[K] = {
+    val e: DrmLike[K] = model.residuals(1 until model.residuals.nrow.toInt, 0 until 1)
```
dlyubimov: `nrow.toInt` is generally dangerous as it does not catch the algorithm limitation. The problem with this (and i actually have run into this before) is that the algorithm obviously has in this case a limitation of 2 bln rows, and it should explicitly fall apart once this limit is reached, instead of silently producing a nonsense. I think there's a method specifically for this purpose in one of our math-scala package `drm`,  `safeToNonNegInt` that would throw IllegalArgument if conversion loses significant bits.

Of course the best approach is to avoid such limitations in the first place, but if unavoidable, please use `safeToNonNegInt`.

dlyubimov:
If not too much hassle, please consider using unicode characters →,  ⇒, ←. 
In intelliJ this is easily facilitated by adding substitution 'live' templates:
spectacle kp8565

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
rawkintrevo: @andrewpalumbo re `drmFeatures` vs `drmX` and @dlyubimov 
We're mixing conventions-
I feel like, for consistency- we either use `drmFeatures` and `drmTarget` (similar to sparkML) or `drmX` and `drmY` similar to sklearn and R- leaving for now, but open to debate- I have a slight bias towards `drmX` , `drmY`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
```diff
@@ -0,0 +1,49 @@
...
+import org.apache.mahout.math.{Vector => MahoutVector}
+import org.apache.mahout.math.algorithms.Model
+import org.apache.mahout.math.drm.DrmLike
+
+import scala.reflect.ClassTag
+
+/**
+  * Abstract of Regressors
+  */
+trait Regressor[K] extends Model {
```
rawkintrevo: I agree that Spark and Flink follow the paradigm you are suggesting, however sklearn doesn't.  If we're just going off of what others do, following the other larger packages- yea we should probably follow conventions of what other scala based "Big Data" packages do.  However, I can't understand WHY they do it that way- it makes the code hard to read/follow and I assume is an artifact of all the serialization and the way they execute models (having to ship object around for map / reduce phases), that is to say they do it because _they are forced to_ and _at the expense of_ readability. 

In Mahout, most of that is taken care of at the distributed engine level.  

If we start going down the rabbit hole of "do as Spark and Flink do" we may find ourselves with [entire class just for the summary of a linear model](https://github.com/apache/spark/blob/master/mllib/src/main/scala/org/apache/spark/ml/regression/LinearRegression.scala#L620).  I for one, want to stay as far away from that as possible.  I'd like to see algorithm code (these and future) be as succinct and tractable as possible so that 

1. New contributors aren't intimidated (that is to say are encouraged to commit algorithms)
2. Those algorithms can be easily reviewed and maintained with minimal Scala knowledge (as it limits the pool of willing and able contributors who understand the actual math at play)

That isn't to say, at the end of the day, your proposal is incorrect- you usually are correct and I value and appreciate you taking the time to review.  I am saying, " i think that's how this pattern goes in most kits." is neither necessary nor sufficient imo, as in some respects I'm explicitly trying to avoid the approach of other packages, in this case- refactoring something to be more complex with no clear understanding of the benefit.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -0,0 +1,52 @@
...
+    val sumSquareTotal = totalResiduals.assign(SQUARE).sum
+    val r2 = 1 - (sumSquareResiduals / sumSquareTotal)
+    model.testResults += ("r2" -> r2)
+    model.summary += s"\nR^2: ${r2}"
+    model
+  }
+
+  // https://en.wikipedia.org/wiki/Mean_squared_error
+  def MeanSquareError[R[K] <: Regressor[K], K](model: R[K]): R[K] = {
+    val mse = model.residuals.assign(SQUARE).sum / model.residuals.nrow
```
rawkintrevo: will update this to `safeToNonNegInt(`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/CochraneOrcutt.scala
```diff
@@ -0,0 +1,89 @@
...
+    betas = new Array[MahoutVector](iterations)
+    regressor.fit(drmFeatures, drmTarget)
+    betas(0) = regressor.beta
+
+    drmY = drmTarget
+
+    val Y = drmTarget(1 until drmTarget.nrow.toInt, 0 until 1).checkpoint(cacheHint)
+    val Y_lag = drmTarget(0 until drmTarget.nrow.toInt - 1, 0 until 1).checkpoint(cacheHint)
+    val X = drmFeatures(1 until drmFeatures.nrow.toInt, 0 until 1).checkpoint(cacheHint)
+    val X_lag = drmFeatures(0 until drmFeatures.nrow.toInt - 1, 0 until 1).checkpoint(cacheHint)
```
rawkintrevo: missed all of these- but have since updated with `safeToNonNegInt(`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/Regressor.scala
dlyubimov: in most kits i know (including two i am working on right now myself) the pattern is that fit returns a model object. sckit seems to be on the outlier end on this.

Also i think the approach "package A does it like this and we do/don't like it therefore it is good/bad" to be a dogma fallacy. IMO We just need to do what makes sense. 

And it makes sense to me to serialize or persist the model, not the (fitting algorithm+model). This will cause problems both on user and implementation ends IMO

andrewpalumbo: +1 to `model` persistence.  Though we need some thinking here.  We could use persist/read the model as a set of Drms, since we have native-to-mahout support for this. ie. drm.dfsWrite(..), drmdfsRead(...). this can be quite clunky though.  (see NaiveBayes model.persist(...) eg.)

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/MeanCenter.scala
```diff
@@ -0,0 +1,96 @@
...
+        val copy: Matrix = block.cloned
+        copy.foreach(row => row -= bcastV.value)
+        (keys, copy)
+    }
+    output
+  }
+
+  def invTransform[K: ClassTag](input: DrmLike[K]): DrmLike[K] = {
+
+    if (!isFit){
```
andrewpalumbo: Style point:   `if (!isFit){` - space after close parens before brace ie: `if (!isFit) {`  
Here and other places.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/ModelFactory.scala
```diff
@@ -0,0 +1,8 @@
+package org.apache.mahout.math.algorithms
```
andrewpalumbo: File needs License

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -0,0 +1,22 @@
+package org.apache.mahout.math.algorithms.regression
```
andrewpalumbo: License

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/AbstractTest.scala
```diff
@@ -0,0 +1,8 @@
+package org.apache.mahout.math.algorithms.regression.tests
```
andrewpalumbo: license

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/AbstractTest.scala
```diff
@@ -0,0 +1,8 @@
...
+
+/**
```
andrewpalumbo: need to remove all references to your name. (autogenereated by Intellij, etc..)

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionTestsSuiteBase.scala
```diff
@@ -0,0 +1,135 @@
+package org.apache.mahout.math.algorithms
```
andrewpalumbo: license

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionSuiteBase.scala
```diff
@@ -0,0 +1,84 @@
...
+      (3, 2, 13,   7,  40.400208),  // Clusters
+      (3, 3, 13,   4,  45.811716)), numPartitions = 2)
+
+    //drmData.collect(::, 0 until 4)
+
+    val drmX = drmData(::, 0 until 4)
+    val drmY = drmData(::, 4 until 5)
+
+    val model = new OrdinaryLeastSquares[Int]().fit(drmX, drmY, 'calcStandardErrors → false)
+
```
andrewpalumbo: If I understand the new factory method correctly, could this just be somethinng like:  
```
val olsModel: OrdinaryLeastSquares[Int] = fit(drmX, drmY, 'calcStandardErrors → false)
``` ?

File: .gitignore [hunk 21]
```diff
@@ -20,3 +20,6 @@ mr/temp
 temp
 foo
 math-tests/
+derby.log
+bin/*
```
andrewpalumbo: is this meant to be added? -- `bin/` don' think we want that in `.gitignore`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
```diff
@@ -0,0 +1,91 @@
...
+import JavaConversions._
+import org.apache.mahout.math._
+import org.apache.mahout.math.drm.{drmBroadcast, _}
+import org.apache.mahout.math.drm.RLikeDrmOps._
+import org.apache.mahout.math.scalabindings.{dvec, _}
+import org.apache.mahout.math.scalabindings.RLikeOps._
+
+import scala.reflect.ClassTag
+
+class AsFactor extends PreprocessorModelFactory{
```
andrewpalumbo: space after opening `{`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
```diff
@@ -0,0 +1,91 @@
...
+          output(i, ::) =  svec(fm.get(block.getQuick(i,0)).get -> 1.0 :: Nil, cardinality = bcastK.get(0).toInt)
+        }
+        (keys, output)
+      }
+    }
+    res
+  }
+
+  override def invTransform[K](input: DrmLike[K]) = {
+    // not yet implemented
```
andrewpalumbo: throw `NotImplementedException` here?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/MeanCenter.scala
```diff
@@ -0,0 +1,93 @@
...
+/**
+  * A model for mean centering each column of a data set at 0 or some number specified by the setCenters method.
+  * @param centers
+  */
+class MeanCenterModel(means: MahoutVector) extends PreprocessorModel {
+
+  var colCentersV: MahoutVector = means
+
+  def setCenters(centers: MahoutVector): Unit ={
+    if (means.length != centers.length){
```
andrewpalumbo: space before opening brace here and line above

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/OrdinaryLeastSquaresModel.scala
```diff
@@ -0,0 +1,123 @@
...
+    model.drmY = drmTarget
+    if (drmFeatures.nrow != drmTarget.nrow){
+      throw new Exception(s"${drmFeatures.nrow} observations in features, ${drmTarget.nrow} observations in target, must be equal.")
+    }
+
+    var X = drmFeatures
+    if (addIntercept) {
+      X = X cbind 1
+    }
+    val drmXtXinv = solve(X.t %*% X)
```
andrewpalumbo: Do we have any around pulling `X.t %*% X` into memory here?  I thought that recently I heard that this was being done differently here.

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionSuiteBase.scala
andrewpalumbo: Looking closer at the code, I see that I was wrong in last comment.. Do we need helper objects for this?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/transformer/MeanCenter.scala
rawkintrevo: isFit has been more or less abstracted away by the Model/ModelFactory paradigm. need to clean this up

.gitignore
rawkintrevo: nope- intellij run amok

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/OrdinaryLeastSquaresModel.scala
rawkintrevo: short answer is  you can solve for beta without doing this- but not the var/co-var matrix.  I can refactor such that if `'calcStandardErrors -> false` then it does it the distributed way. 

Also, so long as XtX isn't toooo fat here, this isn't going to be that brutal.  Even 50k columns dense results in a fairly manageable matrix (50k x 50k)

andrewpalumbo: hmm.. that might be useful- the re-factor for the case w/o a needed covariance matrix.  You could also calculate the covariance matrix in a distributed manner on the out of core Drms..  My concern is that people love to hammer on new features.. Although point is moot if the answer is to use `thinQR(..)` to solve which is also bottlenecked by in-memory ` X.t %*% X` 

Maybe a question for after the platform is finished...

math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
```diff
@@ -28,10 +28,11 @@ import org.apache.mahout.math.drm.{drmBroadcast, _}
 import org.apache.mahout.math.drm.RLikeDrmOps._
 import org.apache.mahout.math.scalabindings.{dvec, _}
 import org.apache.mahout.math.scalabindings.RLikeOps._
+import sun.reflect.generics.reflectiveObjects.NotImplementedException
```
andrewpalumbo: my mistake - in my previous comment.. this should  be `UnsupportedOperationException`.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
```diff
@@ -0,0 +1,117 @@
...
+    import org.apache.mahout.math.function.VectorFunction
+    val factorMap = input.allreduceBlock(
+      { case (keys, block) =>
+        // someday we'll replace this with block.max: Vector
+        // or better yet- block.distinct
+        dense(block.aggregateColumns( new VectorFunction {
+            def apply(f: Vector): Double = f.max
+        }))
+      })(0, ::)
+    new AsFactorModel(factorMap.sum.toInt, factorMap)
```
dlyubimov: style nitpicking. since this really does not overload a trait-signed method, it would be best readable if the return type is explicitly stipulated in the declaration.
```
def fit[K](input: DrmLike[K]):MyModelType = {
```

PR #256, opened 2016-10-04
added minimal log4j.xml to examples project

krichter722:
in order to make logger work when running main classes


smarthi:
Thanks for the PR, but FYI the examples project is not being maintained anymore since its legacy MapReduce.

PR #262, opened 2016-12-15
[MAHOUT-1895] Add convenience methods for converting Vectors to Scala Types

rawkintrevo:
adds the following methods to Vector
.toMap
.toArray

* [x] Add Methods
* [x] Add Tests

Minor Improvement

Relevent JIRA:
[MAHOUT-1895](https://issues.apache.org/jira/browse/MAHOUT-1895) 

Requires no license updates


dlyubimov:
> "While dense and sparse vectors may be created from `TraversableOnce[Double]` such as `Array[Double]` and `TraversableOnce[(Int, AnyVal)]` such as `Map[Int,Double]` respectively"

This is, iirc,  is not quite true. 
only dense vectors can be created from arrays, sparse vectors are explicitly disallowed (or not supported, depends on your take).

So i have a few problems with this. 
(1) similarly, if we want to create vector 2 array transformation, it probably should only be allowed for dense vectors only, otherwise it invites misuses.

(2) even when i do need an array out of dense vector, i usually use it without copying (shallowly). In some implementations, it is adding array() method which returns shallow backing array, if any, similarly to ByteBuffer.array(); and in some implementations I am using a reflection to access the private attribute (which is admittedly is a hack). The former is preferred. 

Either way, the gist of this point is that one quite more often needs a backing array rather than a copy of such. This implementation forces copy-only approach. Note that ByteBuffer doesn't have copying implementation (.toArray()) and only has backing array access (array()) for perhaps the same reasons. Inviting misuse and inefficiency.

(3) Assuming (1) and (2) are overcome, this implementation is still suboptimal for sparse implemetnations.

rawkintrevo:
Thanks for the comments @dlyubimov .  Do you mean we should add a method to expose [this](https://github.com/apache/mahout/blob/master/math/src/main/java/org/apache/mahout/math/DenseVector.java#L29) for dense implementations? 

For `SequentialAccessSparseVector` could we [expose the `OrderedIntDoubleMapping`](https://github.com/apache/mahout/blob/master/math/src/main/java/org/apache/mahout/math/SequentialAccessSparseVector.java#L45) underlying and then add a method [to expose `indices` and `values`](https://github.com/apache/mahout/blob/master/math/src/main/java/org/apache/mahout/math/OrderedIntDoubleMapping.java#L26) then zip them into a Map as a convenience.

Similarly for [`Int2DoubleOpenHashMap` backing `RandomAccessSparseVector`] we could [expose `values`](https://github.com/apache/mahout/blob/master/math/src/main/java/org/apache/mahout/math/RandomAccessSparseVector.java#L36) and then using the interface for `it.unimi.dsi.fastutil.ints.Int2DoubleOpenHashMap` we could [get the `keySet()`](http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/Int2DoubleOpenHashMap.html#keySet--) and `values` which in turn are [`IntSet`](http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/ints/IntSet.html) and [`DoubleCollection`](http://fastutil.di.unimi.it/docs/it/unimi/dsi/fastutil/doubles/DoubleCollection.html) respectively, however both have `toArray()` methods. 

If these are unacceptable, do you have any alternative approaches (and apologies if I entirely missed your point regarding the `DenseVector`s)

dlyubimov:
@rawkintrevo there are really two reasons to want toArray(). One is to have its contracts (I guess it is what you are appealing to when you say you want to wrtie summaries etc.) and one is because Array is common part of somebody else's contract.

On reason one, the refute is simple because Vector does all Array contract does, i.e., it has random access contract with exact the same scala signature. Actually it has much more, including things like pretty print that are useful for summaries.

On reason two, it is a more valid case. However, i found that one much more often wants avoid copying memory anyway and that optional array() is much more desired in all of my usecases. It is true i am an algorithm developer and so my use cases are skewed that way.  One example would be repackaging of dense matrices and vectors we had in VCL integration. that's what we wanted there -- the optionally supported  non-copying array() method. but we did not want a copying method at all since we were repackaging it into native memory anyway. So were some similar other cases.

I am sure you can provide a specific case where you want toArray(), which i suspect might be the motivation where you have to use some other 3rd party contract  and then it might be a better shaped argument if you give us some concrete examples of that 3rd party contract you've encountered a need to use.

> Thanks for the comments @dlyubimov . Do you mean we should add a method to expose this for dense implementations?

yes. This is an alternative i would need.

exposing non-copying collection.Map implementation makes sense for RandomAccessSparseVector and whatever other implementations that adhere to Map's promise of O(1) lookup. For Sequential i don't see it as appropriate non-copying for the same very reason -- it violates O(1) contract promise. 

In general it has occured to me that I am somehwat more ready to consider toMap, toSeq (and even toArray) that implement correspondent Scala collections because it is consistent with strict Scala collection contract use (i.e., admittance of a copying contract) as long as it is available in Scala only. Regardless, implementation must consider individual underlying structure for the copying efficiency. I.e., it has to use non-default element iterators where appropriate (just like we do with VCL adapting).

But i happen to want implementing collection.Map for random access and optional array() for DenseVector more than the copying contracts.

dlyubimov:
Another though i had, if you feel strongly going about toMap, toSeq, toArray method support, perhaps you could group it into a separate decorator implementation, similar to JavaCollections._ approach ((say `import MahoutCollections._`), , so that it can be invoked only when needed, but is cleanly separated from the algebra dialects (which are enabled by `import RlikeOps._`). In other words, i don't what RlikeOps to enable this by default.

andrewpalumbo:
+1 from me

rawkintrevo:
@dlyubimov per your comments- I moved to `MahoutCollections`.

PR #265, opened 2016-12-27
Add optional manual configuration of AtA's number of partitions.

BertrandDechoux:
```scala
// Determine how many partitions the new matrix would need approximately. We base that on
// geometry only, but it may eventually not be that adequate. Indeed, A'A tends to be much more
// dense in reality than the source.
```
Allow override when estimation is not adequate because AtA is indeed more dense...

### DIFF
--- a/spark/src/main/scala/org/apache/mahout/sparkbindings/blas/AtA.scala
+++ b/spark/src/main/scala/org/apache/mahout/sparkbindings/blas/AtA.scala
@@ -37,6 +37,7 @@ object AtA {
 
   final val PROPERTY_ATA_MAXINMEMNCOL = "mahout.math.AtA.maxInMemNCol"
   final val PROPERTY_ATA_MMUL_BLOCKHEIGHT = "mahout.math.AtA.blockHeight"
+  final val PROPERTY_ATA_NUM_PARTS = "mahout.math.AtA.numParts"
 
   /** Materialize A'A operator */
   def at_a(operator: OpAtA[_], srcRdd: DrmRddInput[_]): DrmRddInput[Int] = {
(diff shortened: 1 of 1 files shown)


BertrandDechoux:
cf MAHOUT-1899 https://issues.apache.org/jira/browse/MAHOUT-1899

andrewpalumbo:
Thank you Bertrand for the contribution!  Someone will review it shortly.

dlyubimov:
My problem with this is that it sets up a fixed splitting for an operator (in this case A'A) rather than an estimate, which may work for a particular situation but does not work for any given plan (which may include more than one A'A in a session with totally different properties).

Instead, perhaps it is time to add some clever sparsity heuristics to obtain a better task size estimators if there is indeed a demonstrable problem here.

I meant to do such density based estimators for some time, except (1) i couldn't find a clever way of deriving such heuristics in a plan (or even inside the operator implementation itself) and (2) i did not encounter this being a problem yet. There seems to be no principled approach to predicting product density for A'A  or especially more general AB operators (not to mention other things we do, such as custom elementwise fusions).

dlyubimov:
ps in practice severe "densification" of products occurs due to anomalous input records, e.g., if we have a recommender with anomalous user behavior where user has reviewed every possible movie. In practice often not only could these things be ignored, but actually should be filtered as they represent noisy (anomalous) user behavior that does not reflect the trend.

dlyubimov:
perhaps naive logic of taking product density lower bound estimate as pow(max(rowAdensity),2) could be a good start to detect anomalous cases like that.

BertrandDechoux:
I can provide more information about my use case. Maybe it will help you, maybe not.

If I am not mistaken, for spark-itemsimilarity, the estimation is the following
number_reduces tasks = number_partitions * number_products / number_users

spark-itemsimilarity does a sampling if the number of product interactions for a specific user is too high (500 by default). The high number of interactions is well handled. But indeed this job has only a single AtA so configuring the number of partitions with a fixed value is convenient in that specific case but not in general.

The core of the issue is that, in my case, (number_products / number_users) ~= 1/300

In order to increase parallelism at the aggregation step for throughput and resilience, the multiplication step has to greatly waste ressources with a huge number of partitions...

rawkintrevo:
Can I get a status on this? @BertrandDechoux @dlyubimov

PR #267, opened 2017-01-08
MAHOUT-1900: Add a getter to DenseMatrix for the double[][] values fi…

smarthi:
…eld.


andrewpalumbo:
@smarthi +1 from me-  You might want to check out the conversation around #262 between DL and TG.

PR #268, opened 2017-01-16
MAHOUT-1906: Ensure customJars are added to the MahoutContext under c…

andrewpalumbo:
…ertain conditions for spark 1.6+

When creating MahoutContext: if 1.6+ && backend is spark-standalone && addMahoutContextJars = false && customJars.isEmpty() Context creation fails with serialization error on task serialization.

Fix is to ensure that jars are added in this case.


andrewpalumbo:
merged to master

PR #274, opened 2017-02-04
MAHOUT-1921 Propagate cacheHint in dssvd

rawkintrevo:
Propogate cache hint in dssvd

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
@@ -1,6 +1,6 @@
 package org.apache.mahout.math.decompositions
 
-import org.apache.mahout.math.{Matrix, Matrices, Vector}
+import org.apache.mahout.math.{Matrices, Matrix, Vector}
 import org.apache.mahout.math.scalabindings._
 import RLikeOps._
 import org.apache.mahout.math.drm._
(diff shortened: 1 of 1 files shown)


math-scala/src/main/scala/org/apache/mahout/math/decompositions/DSSVD.scala
```diff
@@ -22,13 +22,18 @@ object DSSVD {
    * @return (U,V,s). Note that U, V are non-checkpointed matrices (i.e. one needs to actually use them
    *         e.g. save them to hdfs in order to trigger their computation.
    */
-  def dssvd[K](drmA: DrmLike[K], k: Int, p: Int = 15, q: Int = 0):
+  def dssvd[K](drmA: DrmLike[K],
+               k: Int,
+               p: Int = 15,
+               q: Int = 0,
+               cacheHint: CacheHint.CacheHint = CacheHint.MEMORY_ONLY):
```
andrewpalumbo: Not sure what we should make the default `CacheHint` here and other places, e.g. #276.

for `drmDfsRead(...)` we use `CacheHint.NONE` but that may be something that we should let the user supply as well.  Most parallaization methods use `CacheHint.NONE`.  I would think that if we have anything hardcoded or as a default, we should use `CacheHint.MEMORY_AND_DISK` since we have no idea of the size of the user's data..  but I suppose that is making Assumptions as well.

PR #276, opened 2017-02-04
MAHOUT-1923 Propagate cacheHint in distributed qr decomp

rawkintrevo:


andrewpalumbo:
same +1 contingent upon approval and decision from #274

PR #277, opened 2017-02-04
MAHOUT-1935 fix double calculation of XtX

rawkintrevo:
Collect XtX only once, and pass it to `solve`.


andrewpalumbo:
+1.. BTW you could also compute and cache it, (if it were actually needed the first time) to avoid recalculation of `X.t %*% X` down the line.

PR #278, opened 2017-02-04
MAHOUT-1936 fix AsFactor allReduce block

rawkintrevo:
Issue in AsFactor fit method was max was being found in "map" phase, not reduce phase.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
@@ -38,11 +38,13 @@ class AsFactor extends PreprocessorFitter {
 
     import org.apache.mahout.math.function.VectorFunction
     val factorMap = input.allreduceBlock(
-      { case (keys, block: Matrix) =>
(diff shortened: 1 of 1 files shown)


math-scala/src/main/scala/org/apache/mahout/math/algorithms/preprocessing/AsFactor.scala
```diff
@@ -38,11 +38,13 @@ class AsFactor extends PreprocessorFitter {
...
     val factorMap = input.allreduceBlock(
-      { case (keys, block: Matrix) =>
+      { case (keys, block: Matrix) => block },
+      { case (oldM: Matrix, newM: Matrix) =>
         // someday we'll replace this with block.max: Vector
         // or better yet- block.distinct
-        dense(block.aggregateColumns( new VectorFunction {
-            def apply(f: Vector): Double = f.max
+
+        dense((oldM rbind newM).aggregateColumns( new VectorFunction {
```
andrewpalumbo: +1

andrewpalumbo:
+1 LGTM -  Do we have any unit tests for the pipeline?

PR #284, opened 2017-02-26
MAHOUT-1938: Switch to Intel AVX 2 instruction set

andrewpalumbo:
A question that we may need to consider- do we want to release the binaries for AVX2?  There does not seem to be a fall-back to AMD arch.

### DIFF
--- a/viennacl-omp/pom.xml
+++ b/viennacl-omp/pom.xml
@@ -141,7 +141,7 @@
                 <argument>-jar</argument>
                 <argument>${org.bytedeco:javacpp:jar}</argument>
                 <argument>-propertyfile</argument>
-                <argument>linux-haswell.properties</argument>
+                <argument>linux-x86_64-viennacl.properties</argument>
                 <argument>-classpath</argument>
                 <argument>${project.build.outputDirectory}:${org.scala-lang:scala-library:jar}</argument>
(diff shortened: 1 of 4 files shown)


andrewpalumbo:
@andrewmusselman could you test this patch out so that wee can ensure that It works on Haswell?

andrewpalumbo:
Could someone with an Intel chipset pre-Haswell (most i-7s) test this to ensure that it does have a fallback to non-avx2 instruction sets?

andrewpalumbo:
build line:
`mvn clean install -Pviennacl -Phadoop2 -DskipTests && cd viennacl && mvn test`

viennacl-omp/pom.xml
```diff
@@ -133,15 +133,15 @@
             </goals>
             <configuration>
               <environmentVariables>
-                <LD_LIBRARY_PATH>{project.basedir}/target/classes/org/apache/mahout/javacpp/linalg/linux-x86_64_omp/
+                <LD_LIBRARY_PATH>{project.basedir}/target/classes/org/apache/mahout/javacpp/linalg/linux-x86_64/
```
andrewpalumbo: For now we can just leave this as a hardcoded hack, I think, but it would be good to have the directory read in a name from the properties file, and use that as the top level path name.

PR #286, opened 2017-02-26
[MAHOUT-1903][MAHOUT-1907]   VCL Vector fix.

andrewpalumbo:
After several weeks of discussion with Karl, it turns out that this was a simple fix.  Vectors *must* be copied back to MAIN_MEMORY, before trying to read elements out.  This seems not to be an issue with Matrices. 

Fixed the memory copy issue, however seeing a small numerical error:
``` 
- VCL Dense Matrix %*% Dense vector *** FAILED ***
  2.1868857395626282E-16 was not less than 1.0E-16 (ViennaCLSuiteVCL.scala:346)
  + Mahout dense matrix %*% dense vector multiplication time: 0 ms. 
  + ViennaCL/OpenCL dense matrix %*% dense vector multiplication time: 5 ms. 
```
Relaxing the assertion by an order of magnitude (-1E16 to -1E15) will solve this problem, however these small numerical differences between the JVM and the Device are  interesting and possibly significant for highly iterative algorithms.

### DIFF
--- a/viennacl/src/main/scala/org/apache/mahout/viennacl/opencl/javacpp/MatrixBase.scala
+++ b/viennacl/src/main/scala/org/apache/mahout/viennacl/opencl/javacpp/MatrixBase.scala
@@ -66,9 +66,9 @@ class MatrixBase extends Pointer {
   @native
   def memoryDomain: Int
 
-  @Name(Array("switch_memory_context"))
-  @native
-  def switchMemoryContext(@ByRef ctx: Context)
+   @Name(Array("switch_memory_context"))
+   @native
+   def switchMemoryContext(@ByRef ctx: Context)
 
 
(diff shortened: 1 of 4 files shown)


andrewpalumbo:
There seems to be some loss of precision when performing matrix %*% vector multiplication on GPU. for each few orders of magnitude I raised the num elements, i had to relax epsilon by an order.

We know that dense algebra is not where OpenCL shines but I am wondering if it is possible that vectors are being converted to fp32 from fp64.  

The same test, with same values, when run in main memory on OpenMP gives the same precision as the Mahout Jvm  with an epsilon of -1e16.  

Not a blocker IMO.  As Well GPU vectors give us access to an entire library of Native Iterative solvers.

PR #290, opened 2017-02-27
Mahout 1913

andrewmusselman:


viennacl/src/main/scala/org/apache/mahout/viennacl/opencl/GPUMMul.scala
```diff
@@ -242,10 +242,11 @@ object GPUMMul extends MMBinaryFunc {
...
        getNumNonDefaultElements() and getNumNonZeroElements() which we do not always
        have access to. We created MAHOUT-1882 for this.
-         val hasElementsA = a.zSum() >  0.0
-         val hasElementsB = b.zSum() >  0.0
     */
 
+    val hasElementsA = a.zSum() >  0.0
+    val hasElementsB = b.zSum() >  0.0
+
     // A has a sparse matrix structure of unknown size.  We do not want to
```
andrewpalumbo: No problems with it thanks @andrewmusselman.  
There should be more println("gpurwrw"), etc,  in this class could you use the logger for that please?

andrewpalumbo: sorry was looking only at one commit.. i see that you already did.

andrewpalumbo:
+1 ship it thanks.

PR #292, opened 2017-03-06
MAHOUT-1762 Utilize spark-submit in bin/mahout script

rawkintrevo:


pferrel:
@rawkintrevo the real problem is missing classes from the dependency-reduced jar, right? How does this solve that?

rawkintrevo:
@pferrel no, the real problem was that the classes were not being shipped out to the cluster.  This solves by leveraging spark-submit to ship jars.

rawkintrevo:
closing this

pferrel:
Was this added to Mahout?

PR #295, opened 2017-03-17
no serializable 

mzjgithub:
…ed due to stage failure: Task 0.0 in stage 0.0 (TID 0) had a not serializable result: org.apache.mahout.math.RandomAccessSparseVector  ...


dlyubimov:
We've already discussed Serializable support for matrices, we will not support that for foreseeable future.  Instead, optimized Kryo support is provided. How do you set up the session? Samsara requires  Kryo serialization enabled.

dlyubimov:
Also it is highly desireable to have Jira issue filed before PR and have it in the header -- that way ASF github bot will update Jira with comments on PR automatically.

PR #296, opened 2017-03-19
MAHOUT-1949 [WIP] Docker Based ITs

rawkintrevo:
- [x] Create Docker Images with Maven of Spark and Hadoop Env
- [x] mount project directory in docker
- [ ] **UPDATE CI ENV to utilize dockers**
- [ ] A script which runs through an exhaustive list of tests (cli drivers/examples/etc)
- [ ] A way to tell weather those tests passed or failed (checking the output?)
- [ ] A way to fail the build if if the examples/etc fail. (no idea how this works, I've always tried to make build successful, never tried to fail one).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,7 @@ foo
 math-tests/
 metastore_db/*
 dockerITs/logs/*.out
+*.jar
+tmp/
+bin/derby.log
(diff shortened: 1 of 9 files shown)


rawkintrevo:
closed JIRA

PR #300, opened 2017-04-04
MAHOUT-1962  Added Ftest to OLS implemenations

dustinvanstee:
Here are a few other mods
-- added 2 tests for OLS
-- modified print statement for summary
-- changed degreeOfFreedom to X.ncols - 1 to remove intercept
-- proposing to move intercept to first column ....


math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -70,17 +76,43 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
     val tScore = model.beta / se
-    val tDist = new org.apache.commons.math3.distribution.TDistribution(n-k)
+    val tDist = new TDistribution(n-k)
+
+    // This is the residual sum of squares for just the intercept
+    //println(" drmTarget.ncol) = " +  drmTarget.ncol)
+    val interceptCol = model.beta.length - 1
+    val targetMean: Double = drmTarget.colMeans().get(0)
+    val rssint: Double = ((drmTarget - targetMean  ).t %*% (drmTarget - targetMean)).zSum()
+    val rssmod: Double = ete.zSum()
```
rawkintrevo: as I recall `ete` is a 1x1 matrix and you can also access the value via `ete(0,0)` see line 76.  

Since we keep doing this, it might make sense just to collect the value a little ways upstream

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -70,17 +76,43 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
     // ^^ TODO bug in this calculation- fix and add test
     //degreesFreedom = k
-
-
     modelOut.se = se
     modelOut.tScore = tScore
     modelOut.pval = pval
-    modelOut.degreesFreedom = X.ncol
+    // for degrees of freedom, dont count the intercept term that was added
+    modelOut.degreesFreedom = X.ncol - 1
```
rawkintrevo: Not saying I'm above making mistakes, but why are we subtracting 1 here?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -123,9 +156,33 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
+    Call:
+      lm(formula = target ~ a + b + c + d, data = df1)
+
+    Residuals:
+    1       2       3       4       5       6       7       8       9
+    -4.2799  0.5059 -2.2783  4.3765 -1.3455  0.7202 -1.8063  1.2889  2.8184
+
+    Coefficients:
+      Estimate Std. Error t value Pr(>|t|)
+    (Intercept)  163.179     51.915   3.143   0.0347 *
```
rawkintrevo: +1 for moving intercept to top of the list

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -41,16 +41,22 @@ object FittnessTests {
     val r2 = 1 - (sumSquareResiduals / sumSquareTotal)
     model.r2 = r2
     model.testResults += ('r2 -> r2)  // need setResult and setSummary method incase you change in future, also to initialize map if non exists or update value if it does
-    model.summary += s"\nR^2: ${r2}"
+    // this is already being done in  -> calculateCommonStatistics
```
rawkintrevo: so it is- good catch.  That said, I think it _should_ happen here, and be taken out of calculateCommonStatistics (adding to the summary string)

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -41,16 +41,22 @@ object FittnessTests {
...
   }
 
   // https://en.wikipedia.org/wiki/Mean_squared_error
   def MeanSquareError[R[K] <: RegressorModel[K], K](model: R[K], residuals: DrmLike[K]): R[K] = {
+    // TODO : I think mse denom should be (row - col) ??
     val mse = residuals.assign(SQUARE).sum / residuals.nrow
     model.mse = mse
     model.testResults += ('mse -> mse)
-    model.summary += s"\nMean Squared Error: ${mse}"
+    // this is already being done in  -> calculateCommonStatistics
```
rawkintrevo: same comment as above

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
rawkintrevo: Looking at this more closely, I don't think you actually implemented like this?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -123,9 +156,33 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
+    Signif. codes:  0 ‘***’ 0.001 ‘**’ 0.01 ‘*’ 0.05 ‘.’ 0.1 ‘ ’ 1
+    */
+
     val k = model.beta.length
-    "Coef.\t\tEstimate\t\tStd. Error\t\tt-score\t\t\tPr(Beta=0)\n" +
-      (0 until k).map(i => s"X${i}\t${model.beta(i)}\t${model.se(i)}\t${model.tScore(i)}\t${model.pval(i)}").mkString("\n")
+
+      // Using Formatted Print here to pretty print the columns
+      "\nCoef.\t\tEstimate\t\tStd. Error\t\tt-score\t\t\tPr(Beta=0)\n" +
+      (0 until k).map(i => "X%-3d\t\t%+5.5f\t\t%+5.5f\t\t%+5.5f\t\t%+5.5f".format(i,model.beta(i),model.se(i),model.tScore(i),model.pval(i))).mkString("\n") +
```
rawkintrevo: +1 for better formatting,

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -41,16 +41,22 @@ object FittnessTests {
...
     model.testResults += ('r2 -> r2)  // need setResult and setSummary method incase you change in future, also to initialize map if non exists or update value if it does
-    model.summary += s"\nR^2: ${r2}"
+    // this is already being done in  -> calculateCommonStatistics
+    //model.summary += s"\nR^2: ${r2}"
     model
   }
 
   // https://en.wikipedia.org/wiki/Mean_squared_error
   def MeanSquareError[R[K] <: RegressorModel[K], K](model: R[K], residuals: DrmLike[K]): R[K] = {
+    // TODO : I think mse denom should be (row - col) ??
```
rawkintrevo: There are `nrow` errors from `nrow` observations?  Is this a d.f. thing?  Link?

math-scala/src/test/scala/org/apache/mahout/math/regression/OrdinaryLeastSquaresTest.scala
```diff
@@ -0,0 +1,64 @@
...
+import org.apache.mahout.math._
+import org.apache.mahout.math.scalabindings._
+import org.apache.mahout.test.DistributedMahoutSuite
+import org.apache.mahout.test.MahoutSuite
+import org.scalatest.{FunSuite, Matchers}
+import collection._
+import JavaConversions._
+import collection.JavaConversions
+/**
+  * Created by dustinvanstee on 3/30/17.
```
rawkintrevo: remove boilerplate from intellij

math-scala/src/test/scala/org/apache/mahout/math/regression/OrdinaryLeastSquaresTest.scala
```diff
@@ -0,0 +1,64 @@
...
+    model.tScore(2) - (-2.32654000) should be < epsilon
+    model.tScore(3) - (-3.01022444) should be < epsilon
+    model.tScore(4) -  3.143183937  should be < epsilon
+
+    model.fScore - 16.38542361  should be < epsilon
+    //println(model.summary)
+
+  }
+
+}
```
rawkintrevo: file ends in blank line

spark/src/test/scala/org/apache/mahout/math/algorithms/regression/OlsSparkTestSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+import org.apache.mahout.math.regression._
+import org.apache.mahout.sparkbindings.test.DistributedSparkSuite
+import org.apache.mahout.test.MahoutSuite
+import org.scalatest.FunSuite
+
+
+class OlsSparkTestSuite extends FunSuite with MahoutSuite with DistributedSparkSuite with OrdinaryLeastSquaresTest {
+  // Common tests located in OrdinaryLeastSquaresTest.scala
+  // The test below is common to spark as I created an random RDD for larger size
+  test("Simple Medium Model2 - Spark Specific") {
```
rawkintrevo: very interesting... @dlyubimov thoughts on this-  doesn't jam anything else up...

I get what you're doing- may be redundant (good test coverage in `math-scala` should cover the spark case)

This would maybe be more appropriate if refactored as an example...

spark/src/test/scala/org/apache/mahout/math/algorithms/regression/OlsSparkTestSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    val n = 10000
+    val features = 100
+    val eps = 10 // i'm guessing error term, poorly documented
+    val partitions = 2
+    val intercept = 10.0
+    //val sc = new SparkContext()
+
+    val synDataRDD = LinearDataGenerator.generateLinearRDD(mahoutCtx, n, features, eps, partitions, intercept)
+
+    // Maybe refine this with something from the sparkbindings, but for now this is good...
```
rawkintrevo: any thing worth doing is worth doing right... you'll find comments like this littered through out the code base dating back to 2008 i think...

https://github.com/apache/mahout/blob/master/spark/src/main/scala/org/apache/mahout/sparkbindings/package.scala#L156

spark/src/test/scala/org/apache/mahout/math/algorithms/regression/OlsSparkTestSuite.scala
```diff
@@ -0,0 +1,80 @@
...
+    val tempRDD = synDataRDD.zipWithIndex.map( lv => {
```
rawkintrevo: `val tempRDD = drmWrapMLLibLabeledPoint(synDataRDD)`

math-scala/src/test/scala/org/apache/mahout/math/regression/OrdinaryLeastSquaresTest.scala
```diff
@@ -0,0 +1,64 @@
+package org.apache.mahout.math.regression
+
```
rawkintrevo: We shouldn't need a new file- refactor the FTest to be like the coefficient of determintaiton / mse, move them to `FitnessTests` and add test to https://github.com/apache/mahout/blob/master/math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionTestsSuiteBase.scala  with an R-Prototype

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -35,6 +37,10 @@ trait LinearRegressorModel[K] extends RegressorModel[K] {
   var tScore: MahoutVector = _
   var pval: MahoutVector = _
   var degreesFreedom: Int = _
+  var trainingExamples :Int = _
+  var fScore: Double = _
```
rawkintrevo: these do belong here

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -70,17 +76,43 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
+    val rssmod: Double = ete.zSum()
+    // ete above is the RSS for the calculated model
+
+    //println(" model.beta(0) = " +  model.beta(0))
+    //println(" model.beta(interceptCol) = " +  model.beta(interceptCol))
+    //println("rssint = " + rssint)
+    //println("rssmod = " + rssmod)
+
+    val groupDof = X.ncol-1
+    val fScore = ((rssint - rssmod) / groupDof) / ( rssmod / (X.nrow - groupDof- 1 ))
```
rawkintrevo: this needs to be refactored as a fitness test- the rationale is unlike R for larger datasets the user will not invoke calcStandardStatistics, and instead will execute the tests she needs a la cart.

rawkintrevo:
Hey Dustin, thanks for the contribution. 

Sorry this slipped through the cracks.

Just reviewed and made comments. 

High level-
- Awesome work on your first PR, this maybe was a more advanced 'beginner' issue
- I think the F-Test needs to be refactored as a fitnessTest so the user can calculate it ala cart later. 
- The unit test for the F-Test needs to have an R prototype, e.g. I ran this same example in R, here was the F- Value, see the comments in fitness tests, you can probably use those and see what the F-score was, and just report it there. 
- Not sure the spark specific test belongs as a unit test.  Maybe better as an example.

rawkintrevo:
As a second note- if you put MAHOUT-1962 in the title of the PR a lot of things will automatically integrate, not sure if you can edit it now- but that is a standard proceedure

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
dustinvanstee: Not yet, I was going to explore moving the intercept to the first index in the array for seperate PR.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
dustinvanstee: ok, will go back and fix that ..

spark/src/test/scala/org/apache/mahout/math/algorithms/regression/OlsSparkTestSuite.scala
rawkintrevo: Never mind- build is failing on this. 

Keep the math tests in the math suite.  They will be tested on spark as well as others.

rawkintrevo:
Just beezing through this again... I would drop `spark/src/test/scala/org/apache/mahout/math/algorithms/regression/OlsSparkTestSuite.scala` all together.

andrewpalumbo:
this is not really apropos of this PR, but looking through, reminded me that we have some a classification stats package in https://github.com/apache/mahout/tree/master/math-scala/src/main/scala/org/apache/mahout/classifier/stats.  There may be some useful code there, that could be refactored into to the new framework.

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionTestsSuiteBase.scala
```diff
@@ -106,4 +106,83 @@ trait RegressionTestsSuiteBase extends DistributedMahoutSuite with Matchers {
     (myAnswer - correctAnswer) should be < epsilon
   }
 
+  test("OLS w/Ftest and Ttest validation") {
+
```
rawkintrevo: Needs an R prototype

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
...
     val mse = residuals.assign(SQUARE).sum / residuals.nrow
     model.mse = mse
     model.testResults += ('mse -> mse)
     model.summary += s"\nMean Squared Error: ${mse}"
     model
   }
+
+  // https://en.wikipedia.org/wiki/xxxx
```
rawkintrevo: might as well finish out the link.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
 
   // https://en.wikipedia.org/wiki/Mean_squared_error
   def MeanSquareError[R[K] <: RegressorModel[K], K](model: R[K], residuals: DrmLike[K]): R[K] = {
+    // TODO : I think mse denom should be (row - col) ??
```
rawkintrevo: you may be right, why do you think this?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
...
+    //val targetMean: Double = drmTarget
+    val targetMean: Double = drmTarget.colMeans().get(0)
+
+    val rssint: Double = ((drmTarget - targetMean  ).t %*% (drmTarget - targetMean)).zSum()
+    // ete above is the RSS for the calculated model
+
+    //println(" model.beta(0) = " +  model.beta(0))
+    //println(" model.beta(interceptCol) = " +  model.beta(interceptCol))
+    //println("rssint = " + rssint)
+    //println("rssmod = " + rssmod)
```
rawkintrevo: get rid of these.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
...
     model.summary += s"\nMean Squared Error: ${mse}"
     model
   }
+
+  // https://en.wikipedia.org/wiki/xxxx
+  def FTest[R[K] <: RegressorModel[K], K](model: R[K],  drmFeatures: DrmLike[K], drmTarget: DrmLike[K]): R[K] = {
+
+    // This is the residual sum of squares for just the intercept
+    //println(" drmTarget.ncol) = " +  drmTarget.ncol)
+    val interceptCol = drmTarget.ncol - 1
```
rawkintrevo: should this be `drmFeatures.ncol`?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -54,46 +58,72 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
     //degreesFreedom = k
-
-
     modelOut.se = se
     modelOut.tScore = tScore
     modelOut.pval = pval
-    modelOut.degreesFreedom = X.ncol
-    modelOut.summary = generateSummaryString(modelOut)
+    // for degrees of freedom, dont count the intercept term that was added
+    modelOut.degreesFreedom = X.ncol - 1
```
rawkintrevo: you're correct you but you need to check if a intercept was added- `if (model.addIntercept) {`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -54,46 +58,72 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
+
+    modelOut.trainingExamples = n.toInt
```
rawkintrevo: This is dangerous- could be to big of a number to fit in an Int. `safeToNonNegInt(residuals.nrow)`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -54,46 +58,72 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
+
     if (calcCommonStatistics){
-      modelOut = calculateCommonStatistics(modelOut, drmTarget, residuals)
+      modelOut = calculateCommonStatistics(modelOut, X, drmTarget, residuals)
```
rawkintrevo: OK- you changed this to take `X` but the only thing you use X for is to calculate teh d.o.f. and `model.trainingExamples` which are already included in teh model (you just added them).  I would refactor this so X isn't a parameter.

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/LinearRegressorModel.scala
```diff
@@ -54,46 +58,72 @@ trait LinearRegressorFitter[K] extends RegressorFitter[K] {
...
     }
+
+    // Let Statistics Get Calculated prior to assigning the summary
+    modelOut.summary = generateSummaryString(modelOut)
+
     modelOut
   }
 
+  // Since rss is needed for multiple test statistics, use this function to cache this value
+  def calculateResidualSumOfSquares[M[K] <: LinearRegressorModel[K]](model: M[K],residuals: DrmLike[K]) : M[K] ={
```
rawkintrevo: As a stylistic point and because other non-linear regressors may/will need this- please move to `FitnessTests`

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/OrdinaryLeastSquaresModel.scala
```diff
@@ -55,6 +55,8 @@ class OrdinaryLeastSquares[K] extends LinearRegressorFitter[K] {
     }
 
     var X = drmFeatures
+
+    // TODO : move intercept betas to the first column, not the last column ??
```
rawkintrevo: Start a JIRA, don't introduce a comment into the code in a file you're otherwise not touching ;)

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/RegressorModel.scala
```diff
@@ -30,6 +30,9 @@ trait RegressorModel[K] extends SupervisedModel[K] {
   // Common Applicable Tests- here only for convenience.
   var mse: Double = _
   var r2: Double = _
+  var fpval: Double = _
+  var rss:Double = -9999.0
```
rawkintrevo: Sure there is a reason, but why are you initializing to -9999.0 instead of _ ?

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
rawkintrevo: In addition- it seems that this is only used in some print statements that need to be deleted anyway. Probably can drop this line all together

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
...
+
+    val rssint: Double = ((drmTarget - targetMean  ).t %*% (drmTarget - targetMean)).zSum()
+    // ete above is the RSS for the calculated model
+
+    //println(" model.beta(0) = " +  model.beta(0))
+    //println(" model.beta(interceptCol) = " +  model.beta(interceptCol))
+    //println("rssint = " + rssint)
+    //println("rssmod = " + rssmod)
+
+    val groupDof = drmFeatures.ncol-1
```
rawkintrevo: replace this with `model.degreesFreedom` (and refactor to not need the features in the signature)

math-scala/src/main/scala/org/apache/mahout/math/algorithms/regression/tests/FittnessTests.scala
```diff
@@ -47,10 +52,44 @@ object FittnessTests {
...
+    val fScore = ((rssint - model.rss) / groupDof) / ( model.rss / (drmFeatures.nrow - groupDof- 1 ))
```
rawkintrevo: this line is the crux of it- Looks good, but I'll feel better if you can verify with an R-prototype

math-scala/src/test/scala/org/apache/mahout/math/algorithms/RegressionTestsSuiteBase.scala
rawkintrevo: actually this all can be moved to the "fitness tests" , use the R-prototype in there and just confirm the F-test, should be in good shape.

rawkintrevo: e.g. you don't need these separate tests.

rawkintrevo:
Big job @dustinvanstee ! Looking awesome. Made a few more comments, you're getting pretty close. 

Thanks again!

rawkintrevo:
sorry for delay @dustinvanstee 

I'm still not in love with moving all of the summary string generation to one function- but I think really we should re think the entire thing and try to do as much as possible in one pass. 

This however would be a separate JIRA. 

lgtm unless any further comments

PR #302, opened 2017-04-20
[WIP][DISCUSS] Initial work on JCuda bindings

andrewpalumbo:
Initial work On JCuda Bindings by @nsakharnykh + some quick timing tests.

### DIFF
--- a/math-scala/src/main/scala/org/apache/mahout/math/backend/RootSolverFactory.scala
+++ b/math-scala/src/main/scala/org/apache/mahout/math/backend/RootSolverFactory.scala
@@ -57,6 +57,14 @@ final object RootSolverFactory extends SolverFactory {
   def getOperator[C: ClassTag]: MMBinaryFunc = {
 
     try {
+      logger.info("Creating org.apache.mahout.cuda.GPUMMul solver")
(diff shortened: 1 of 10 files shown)


math-scala/src/main/scala/org/apache/mahout/math/backend/RootSolverFactory.scala
```diff
@@ -57,6 +57,14 @@ final object RootSolverFactory extends SolverFactory {
   def getOperator[C: ClassTag]: MMBinaryFunc = {
 
     try {
+      logger.info("Creating scala.cuda.GPUMMul solver")
```
rawkintrevo: Does this remove ViennaCL from the picture all together then?

andrewpalumbo: @rawkintrevo No, we'll maintain separate profiles for ViennaCL (`-Pviennacl` and `-Pcuda` ) and JCuda to begin with. `RootSolverFactory` has a a lot of work to be done on it and will be completely reworked/refactored, in the coming releases.  In fact we have Jiras open for much of this work for 0.13.1.    

Currently its not a huge issue to simply check the classpath at runtime (from the source build), however we will need to do some thinking on how and where to place these jars when this is released for the binary distro.

andrewpalumbo:
see  #310

PR #303, opened 2017-04-20
MAHOUT-1965 Multiple Spark-Scala builds in TravisCI

rawkintrevo:
Add ASLv2 to travis.yml

Uncommented Flink Module

Told Maven to skip Flink in CI build 

Build / Test Strategy based on what was formerly in runtests.sh

Deleted runtests.sh

Bumped TravisCI environment from precise to trusty

Added installations nessecary for build OMP/ViennaCL


pom.xml
```diff
@@ -843,7 +843,7 @@
     <module>distribution</module>
     <module>math-scala</module>
     <module>spark</module>
-    <!--module>flink</module-->
+    <module>flink</module>
```
andrewpalumbo: Since is in the works to be fixed with a profile, I'm +1.  That should come relatively soon though- since Flink often breaks the Nightlies and almost always Crashes Hudson.

andrewpalumbo:
lgtm +1

pom.xml
rawkintrevo: we can stand down until the profile is in- should be a day or two.

PR #344, opened 2017-10-09
Add an IfStatement to check the return value of resultFile.delete().

BruceKuiLiu:
This statement returns a value that is not checked.
The return value should be checked since it can indicate an unusual or unexpected function execution.
The statement returns false if the file could not be successfully deleted (rather than throwing an Exception).
If the result was not checked, developers would not notice if the statement signals an unexpected behavior by returning an atypical return value.
http://findbugs.sourceforge.net/bugDescriptions.html#RV_RETURN_VALUE_IGNORED_BAD_PRACTICE


rawkintrevo:
Hey @BruceKuiLiu,

PR #345, opened 2017-10-09
Remove the redundant null check statements of a non-null value.

BruceKuiLiu:
This IfStatement is a redundant check of a known non-null otherObj because the null check of otherObj is contained in the previous instanceof check.
http://findbugs.sourceforge.net/bugDescriptions.html#RCN_REDUNDANT_NULLCHECK_OF_NONNULL_VALUE


rawkintrevo:
Hey @BruceKuiLiu,

PR #346, opened 2017-10-09
Consider returning a zero length array rather than null.

BruceKuiLiu:
It is often a better design to return a length zero array rather than a null reference to indicate that there are no results (i.e., an empty list of results).
This way, no explicit check for null is needed by clients of the method.
On the other hand, using null to indicate "there is no answer to this question" is probably appropriate.
http://findbugs.sourceforge.net/bugDescriptions.html#PZLA_PREFER_ZERO_LENGTH_ARRAYS


rawkintrevo:
Hi @BruceKuiLiu thanks for the contribution and welcome to the community.

Could you please join dev@mahout.apache.org and bring this up for discussion, or at least start JIRA issue where we can discuss. 

https://issues.apache.org/jira/projects/MAHOUT/issues/MAHOUT-2023?filter=allopenissues

rawkintrevo:
Hey @BruceKuiLiu,

PR #353, opened 2017-12-04
MAHOUT-2023 Fix Scopt Target

rawkintrevo:
### Purpose of PR:

### Important ToDos
Please mark each with an "x"
- [x] A JIRA ticket exists (if not, please create this first)[https://issues.apache.org/jira/browse/ZEPPELIN/]
- [x] Assigned JIRA to self
- [x] Successfully built and ran all unit tests, verified that all tests pass locally.

If all of these things aren't complete, but you still feel it is
appropriate to open a PR, please add [WIP] after MAHOUT-XXXX before the
descriptions- e.g. "MAHOUT-XXXX [WIP] Description of Change"

no 

no

### DIFF
--- a/spark/src/main/assembly/dependency-reduced.xml
+++ b/spark/src/main/assembly/dependency-reduced.xml
@@ -38,7 +38,8 @@
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
-        <include>com.github.scopt:scopt_${scala.compat.version}</include>
+        <include>com.github.scopt:scopt_2.10</include>
+        <include>com.github.scopt:scopt_2.11</include>
         <include>com.tdunning:t-digest</include>
         <include>org.apache.commons:commons-math3</include>
         <include>it.unimi.dsi:fastutil</include>
--- a/spark/src/main/assembly/dependency-reduced.xml
+++ b/spark/src/main/assembly/dependency-reduced.xml
@@ -38,7 +38,7 @@
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
-        <include>com.github.scopt_${scala.compat.version}</include>
+        <include>com.github.scopt:scopt_${scala.compat.version}</include>
         <include>com.tdunning:t-digest</include>
         <include>org.apache.commons:commons-math3</include>
         <include>it.unimi.dsi:fastutil</include>
--- a/spark/src/main/assembly/dependency-reduced.xml
+++ b/spark/src/main/assembly/dependency-reduced.xml
@@ -34,7 +34,7 @@
       </unpackOptions>
       <scope>runtime</scope>
       <outputDirectory>/</outputDirectory>
-      <useTransitiveFiltering>true</useTransitiveFiltering>
+      <useProjectArtifact>true</useProjectArtifact>
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
@@ -42,8 +42,6 @@
         <include>com.tdunning:t-digest</include>
         <include>org.apache.commons:commons-math3</include>
         <include>it.unimi.dsi:fastutil</include>
-        <include>org.apache.mahout:mahout-native-viennacl_${scala.compat.version}</include>
-        <include>org.apache.mahout:mahout-native-viennacl-omp_${scala.compat.version}</include>
         <include>org.bytedeco:javacpp</include>
       </includes>
     </dependencySet>
--- a/buildtools/change-scala-version.sh
+++ b/buildtools/change-scala-version.sh
@@ -46,10 +46,17 @@ check_scala_version() {
 
 check_scala_version "$TO_VERSION"
 
+SCALA_210_LONG="2.10.6"
+SCALA_211_LONG="2.11.8"
+
 if [ $TO_VERSION = "2.11" ]; then
   FROM_VERSION="2.10"
+  FROM_LONG_VERSION=$SCALA_210_LONG
+  TO_LONG_VERSION=$SCALA_211_LONG
 else
   FROM_VERSION="2.11"
+  FROM_LONG_VERSION=$SCALA_211_LONG
+  TO_LONG_VERSION=$SCALA_210_LONG
 fi
 
 sed_i() {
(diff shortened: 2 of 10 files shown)


pferrel:
The problem is not just scopt but all artifacts that have {scala.compat.version} in the artifact name. That means 

        <include>org.apache.mahout:mahout-native-viennacl_${scala.compat.version}</include>
        <include>org.apache.mahout:mahout-native-viennacl-omp_${scala.compat.version}</include>

too

rawkintrevo:
At very least they were both problems. 

This fixes the lack of scopt. Considering that scopt now gets picked up, my gut tells me this is something else. The warning message knows what they are looking for, just can't find them.

rawkintrevo:
I seem to have validated my suspicion. 
The viennacl's are being skipped because they are not in fact dependencies, so Maven skips them. 
You can verify this by first ensuring you have installed viennacl:
`mvn clean install -Pviennacl`

Then go back to `mahout/spark/pom.xml` and add the dependency:
```
    <dependency>
      <groupId>org.apache.mahout</groupId>
      <artifactId>mahout-native-viennacl_2.10</artifactId>
      <version>0.13.1-SNAPSHOT</version>
    </dependency>
```

Now run `mvn clean package` and you will see
```
[WARNING] The following patterns were never triggered in this artifact inclusion filter:
o  'org.apache.mahout:mahout-native-viennacl-omp_2.10'

```

Scopt is required by `mahout-math-scala` (which to me is silly, but there it is- my .02 is to move the drivers to examples). So vis-a-vis math-scala, `scopt` IS a dependency and therefor will get picked up. ViennaCL is in NO way a dependency and therefor will never be included in the dependeny reduced jar. 

I am updating my PR to remove the two offending lines @pferrel noted, as they simply don't belong at all.  The user will need to add those jars to the class path of the spark job they run, this could stand to be documented better- I leave it as an exercise for @andrewpalumbo or any passing contributor.

andrewpalumbo:
I am almost positive that we need  
```
<include>org.apache.mahout:mahout-native-viennacl_${scala.compat.version}</include>
<include>org.apache.mahout:mahout-native-viennacl-omp_${scala.compat.version}</include>
```
They are never actually dependencies, however are picked up on the back ends at run time if they exist on the path, hence including them in the `dependency-reduced.xml`.  I will mess with it a little bit later today.

pferrel:
The fix doesn’t seem to be working for me. what am I doing wrong?

@rawkintrevo I 
1) pulled your mahout-2023 branch
2) `./buildtools/change-scala-version.sh 2.11`
3) `mvn clean install -Pscala-2.11,spark-2.1`
4) `jar tf mahout-spark_2.11-0.13.1-SNAPSHOT-dependency-reduced.jar | grep scopt`

and get nothing, it's not going into the dependency-reduced.jar. 
What am I doing wrong? (edited)


[10:34] 
also `mahout spark-itemsimilarity` reports a missing scopt as you'd expect


[10:36] 
This is what the `dependency-reduced.xml` looks like:
```      <includes>
        <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
        <include>com.google.guava:guava</include>
        <include>com.github.scopt:scopt_${scala.compat.version}</include>
        <include>com.tdunning:t-digest</include>
        <include>org.apache.commons:commons-math3</include>
        <include>it.unimi.dsi:fastutil</include>
        <include>org.bytedeco:javacpp</include>
      </includes>
```


[10:37] 
after `./buildtools/change-scala-version.sh 2.11`


[10:38] 
does that script need to hit `dependency-reduced.xml`?

rawkintrevo:
I did two updates. 

One, both scopts are included now. Because maven will automatically filter the one that isn't a dependency, no conflicts here. 

I also refactored to drop the scala-2.11, scala-2.10 profiles.  It's all handled by buildtools/change-scala-version.sh

In the parent pom you will still see $scala.version / $scala.compat.version. These are only needed to handle the ant-copy plugin, which is itsself a hack and should probably be removed, but requires refactoring where and how the scala-shell picks up jars.

spark/src/main/assembly/dependency-reduced.xml
```diff
@@ -34,16 +34,15 @@
...
       <includes>
         <!-- guava only included to get Preconditions in mahout-math and mahout-hdfs -->
         <include>com.google.guava:guava</include>
-        <include>com.github.scopt_${scala.compat.version}</include>
+        <include>com.github.scopt:scopt_2.10</include>
+        <include>com.github.scopt:scopt_2.11</include>
         <include>com.tdunning:t-digest</include>
         <include>org.apache.commons:commons-math3</include>
         <include>it.unimi.dsi:fastutil</include>
-        <include>org.apache.mahout:mahout-native-viennacl_${scala.compat.version}</include>
```
andrewpalumbo: can do the same thing as is done with scopt here,  ie. include both versions and then use `buildtools/change-scala-versions.sh` to switch?

rawkintrevo: deal- there's no switch happening with change-scala-version.sh though

rawkintrevo: I take it back- viennacl won't be included bc there is no dependency... sorry.

andrewpalumbo: I think after we discussed yesterday, if adding these jars to the classpath is the better practice, and we want this jar to be a slim as possible, leaving them out would be fine.   I am unclear on how often its shipped around on the backend.

 I think that the spark method- just adding to the `spark.executor.extraClassPath` and `spark.driver.extraClassPath` makes sense.  @rawkintrevo this is what you were saying we should to, correct and to just add it to the documentation?  I don't think that the `dependency-reduced.jar` is the way we want to be shipping everything in the future.

I think that this is something that we as a project need to discuss at some point.  The original goal for these solver modules (and the somewhat current, hacky implementation) was to probe the classpath for the existence of a solver, and then to use e.g. viennacl-omp#Mmul if it is on the classpath.  Not sure that adding it to the dependency reduce jar is the way to go.  

I say we just ship this as is-  still need to do more testing on a cluster due to pat's next scopt find:

```
Exception in thread "main" java.lang.NoClassDefFoundError: scala/Product
    at java.lang.ClassLoader.defineClass1(Native Method)
    at java.lang.ClassLoader.defineClass(ClassLoader.java:763)
```

pferrel: No we can't ship this with spark-itemsimilarity driver not working. about 50% of the few questions on the list, not to mention the IBM folks are using it.

andrewpalumbo:
Sry @pat @rawkintrevo.. I'd meant ship w/o Viennacl jars in dependency-reduced.xml.  yeah def have to fix scopt issue.



Sent from my Verizon Wireless 4G LTE smartphone


rawkintrevo:
Cool. I'm out. Someone else can give this a shot.

spark/src/main/assembly/dependency-reduced.xml
rawkintrevo: I don't think IBM is using this- they compiled their own code.

pferrel: As I have because we are having trouble here

PR #358, opened 2018-07-19
MAHOUT-2052 [WIP] Broken codeblock in a samsara tutorial html file.

zzinny:
### Purpose of PR:
As you can see below,
there is a broken codeblock and the tag ``<pre>`` is shown in a samsara tutorial html file.

image

This is because of ``<pre class="codehilite"></pre>`` and it should be changed to ``<div class="codehilite"><pre> ... </pre></div>``

### Important ToDos
Please mark each with an "x"
- [x] A JIRA ticket exists (if not, please create this first)[https://issues.apache.org/jira/browse/ZEPPELIN/]
- [ ] Created unit tests where appropriate
- [ ] Added licenses correct on newly added files
- [ ] Assigned JIRA to self
- [ ] Added documentation in scala docs/java docs, and to website
- [ ] Successfully built and ran all unit tests, verified that all tests pass locally.

If all of these things aren't complete, but you still feel it is
appropriate to open a PR, please add [WIP] after MAHOUT-XXXX before the
descriptions- e.g. "MAHOUT-XXXX [WIP] Description of Change"

No

No

### DIFF
--- a/website/docs/latest/tutorials/samsara/play-with-shell.md
+++ b/website/docs/latest/tutorials/samsara/play-with-shell.md
@@ -41,10 +41,12 @@ We describe how to do a quick toy setup of Spark & Mahout on your local machine,
 
  1. Goto the directory where you unpacked Spark and type ```sbin/start-all.sh``` to locally start Spark
  1. Open a browser, point it to [http://localhost:8080/](http://localhost:8080/) to check whether Spark successfully started. Copy the url of the spark master at the top of the page (it starts with **spark://**)
- 1. Define the following environment variables: <div class="codehilite"><pre>export MAHOUT_HOME=[directory into which you checked out Mahout]
-export SPARK_HOME=[directory where you unpacked Spark]
-export MASTER=[url of the Spark master]
-</pre></div>
+ 1. Define the following environment variables: 
+    <pre class="codehilite">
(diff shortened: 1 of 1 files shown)


andrewmusselman:
Thanks for the edit! I am not seeing a repro case in two browsers (chrome and brave on ubuntu) but if you are able to check that this fixes this in your browser then I think it's good. I think `pre` is a block-level element so wrapping in a `div` wouldn't change much.

Which browser are you in?

wwooww374:
Thanks for your comment!

I tested my code in Chrome, Safari, and Brave by running ``jekyll serve`` and there was no change. As you said, I think this is not a problem with ``<pre>`` not wrapped in ``<div>``.

Instead, I separated ``<pre>`` tag section and text in new commit because all other codeblocks without this problem is separated from surrounding text. When I tested it in Chrome, Safari, and Brave, it fixed the problem.

image

rawkintrevo:
@wwooww374 first of all, thanks for the contribution!

Second, this file was written in the old CMS website (pre jekyll) there are two additional options. 
`<code> </code>` could replace `<pre></pre>` all together. 

OR

In Markdown, code can be denoted by the back tick- e.g. `Shift+~`  or ` it's hard to make it show up, bc Github also supports this kind of denoting.

If everything is working now thought, let me know- I'll test it on my system too.

PR #371, opened 2019-04-28
(nojira) fix for http://openjdk.java.net/jeps/212; update SCM tag to 0.14.1

andrewpalumbo:
JavaDoc errors from java 8 generated html; still has warnings, but project builds.


andrewpalumbo:
72 hrs- pushing; lazy consensus

PR #424, opened 2023-02-14
MAHOUT-2155 Add Javadocs for RandomWrapper

rawkintrevo:
### Purpose of PR:

Add javadocs for RandomWrapper

### Important ToDos
Please mark each with an "x"
- [x] Assigned JIRA to self
- [x] Added documentation in scala docs/java docs, and to website

Does this change break earlier versions? No

Is this the beginning of a larger project for which a feature branch should be made? No

### DIFF
--- a/core/src/main/java/org/apache/mahout/common/RandomWrapper.java
+++ b/core/src/main/java/org/apache/mahout/common/RandomWrapper.java
@@ -79,11 +79,17 @@ public final class RandomWrapper extends Random {
     return random;
   }
 
-
+  /**
+   * This method is not supported. It will throw an {@link UnsupportedOperationException}.
+   * Reason: This method is protected in {@link Random} and cannot be delegated.
+   *
+   * Callers can't use this method and other methods are delegated, so shouldn't matter.
+   *
+   * @param bits - why are you still reading?
+   * @return An error, don't use it.
+   */
   @Override
   protected int next(int bits) {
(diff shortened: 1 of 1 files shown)


core/src/main/java/org/apache/mahout/common/RandomWrapper.java
```diff
@@ -47,14 +65,21 @@ public void setSeed(long seed) {
...
   }
 
+    /**
+     * @return the delegate random number generator.
+     */
   public RandomGenerator getRandomGenerator() {
     return random;
   }
 
+
```
andrewmusselman, line 82: Do we need this extra newline?

core/src/main/java/org/apache/mahout/common/RandomWrapper.java
```diff
@@ -47,14 +65,21 @@ public void setSeed(long seed) {
...
   @Override
   protected int next(int bits) {
```
andrewmusselman, line 84: While we're at it can we add a comment block to this method?

andrewmusselman (approved):
Looks good with changes suggested

core/src/main/java/org/apache/mahout/common/RandomWrapper.java
rawkintrevo: Addressed.

PR #426, opened 2023-02-23
MAHOUT-2161Fix website link in sidebar

rawkintrevo:
### Purpose of PR:
Fixes link (it did on my machine, but not sure why link wasn't working before soo may or may not actually fix)

### Important ToDos
Please mark each with an "x"
- [x] Assigned JIRA to self

Does this change break earlier versions? No

Is this the beginning of a larger project for which a feature branch should be made? No

### DIFF
--- a/website/_includes/sidebar.html
+++ b/website/_includes/sidebar.html
@@ -6,7 +6,7 @@
     <div class="card-body">
         <ul class="list-unstyled mb-0">
             <li>
-                <a href="http://www.apache.org/foundation/how-it-works.html">Apache Software Foundation</a>
+                <a href="https://www.apache.org/foundation/how-it-works.html">Apache Software Foundation</a>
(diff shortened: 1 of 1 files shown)


andrewmusselman (commented):
I don't think this fixes what I'm seeing.

This should be changing to a hover state with an underline and it's not. Similar but different weird behavior is happening on the LICENSE link too: takes a second for it to show the hover state, and then a second later the hover state disappears.

rawkintrevo:
I did a right-click-inspect on it, and I _think_ the issue is the canvas from the cool wavy line graphic is sometimes being rendered on top of it. If that's the case- my recco is to punt on this until #2158 (the jira, not the PR, i don't get github anymore)

andrewmusselman:
Okay sounds good to me

Issue #468, opened 2024-10-04
Add a Parameter object


andrewmusselman:
Needed for PQC work

PR #472, opened 2024-11-05
MAHOUT-468 &  MAHOUT-469 Add parameter object and bind at execution

rawkintrevo:
### Purpose of PR:
#468 Add a parameter object
#469 Bind Parameters at execution
with bonus notebook for optimization loop external to qumat

### Important ToDos
Please mark each with an "x"
- [ ] Created unit tests where appropriate

- [x] Assigned JIRA to self
- [x] Added documentation in scala docs/java docs, and to website

https://colab.research.google.com/drive/1QgeqKZtkC_hOmffRprxJWMcC1g9YsMOE?usp=sharing qiskit
https://colab.research.google.com/drive/1M5tT7h2wRiGNYx48g_FDGPTKuogKxEzc?usp=sharing cirq

bracket untested

no

no

### DIFF
--- a/examples/Optimization_Example.ipynb
+++ b/examples/Optimization_Example.ipynb
@@ -1,11 +1,11 @@
 {
-"nbformat": 4,
- "nbformat_minor": 0,
+ "nbformat": 4,
+ "nbformat_minor": 2,
  "cells": [
- {
+  {
    "cell_type": "code",
    "source": [
-    "!pip install git+https://github.com/apache/mahout.git@main"
+    "!pip install git+https://github.com/apache/mahout.git@468-469"
    ],
    "metadata": {
     "colab": {
@@ -111,16 +111,16 @@
     "# Run the optimization loop\n",
     "for iteration in range(10):  # Reduced iteration count for brevity\n",
     "    cost = your_cost_function()  # Evaluate the cost function\n",
-    "    \n",
+    "\n",
     "    # Update parameter(s) based on some optimization logic\n",
     "    # This is a placeholder update mechanism\n",
     "    current_parameters['theta'] += 0.1\n",
     "    current_parameters['phi'] += 0.1\n",
     "    current_parameters['lambda'] += 0.1\n",
-    "    \n",
+    "\n",
     "    # Execute the circuit with the updated parameters\n",
     "    result = qumat_instance.execute_circuit(parameter_values=current_parameters)\n",
-    "    \n",
+    "\n",
     "    print(f\"Iteration {iteration}, Result: {result}, Parameters: {current_parameters}\")"
    ]
   },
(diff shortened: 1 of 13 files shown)


rawkintrevo:
On the last commit- i had to specify my fork in the install (bc I added functionality that isn't available in main), we need a more elegant solution to do this.

examples/Simple_Example.ipynb
```diff
@@ -29,7 +29,7 @@
   {
    "cell_type": "code",
    "source": [
-    "!pip install git+https://github.com/apache/mahout.git@main"
+    "# pip install git+https://github.com/apache/mahout.git@main"
```
andrewmusselman, line 32: Is this supposed to be commented out?

examples/Optimization_Example.ipynb
```diff
@@ -0,0 +1,144 @@
+{
```
andrewmusselman, line 1: Does this need an "Open in Colab" button?

examples/Optimization_Example.ipynb
```diff
@@ -0,0 +1,144 @@
...
+   "cell_type": "code",
+   "execution_count": 2,
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "# Create a quantum circuit with 1 qubit\n",
+    "qumat_instance.create_empty_circuit(1)\n",
+    "\n",
+    "# Apply parameterized RX, RY, and RZ gates\n",
+    "qumat_instance.apply_rx_gate(0, 'theta')\n",
```
andrewmusselman, line 81: I don't see where `theta`, `phi`, and `lambda` are defined, should we add those definitions?

andrewmusselman (commented):
I've read through and left some comments; can we add a button to run the Optimization_Example.ipynb in Colab so we can show it working?

examples/Simple_Example.ipynb
rawkintrevo: Yes, for notebooks to actually be tested on the code that is being checked in it does.

examples/Optimization_Example.ipynb
rawkintrevo: It could, what is the standard on those? (I've seen in notebook, in a readme file, both, neither)

examples/Optimization_Example.ipynb
rawkintrevo: That's a fair call out. They're standard in rotational gates, but it wouldn't hurt to add some docs calling them out.

examples/Optimization_Example.ipynb
andrewmusselman: I don't have a rule of thumb but I know it's on the Simple Example and it's great. Let's just add to `.ipynb` files?

examples/Optimization_Example.ipynb
andrewmusselman: If this runs for you I'm good approving; I'll make a new issue to add these def'ns later.

andrewmusselman:
Added these two issues to track minor improvements

PR #528, opened 2025-04-08
Update Gemfile

krishna-dave206:
Added a few extra lines in the Gemfile so that it doesn't trouble users who have older Ruby versions to setup the website locally.


rawkintrevo:
The gemfile as it is builds for the asf, which is paramount. Rejecting this.

Issue #529, opened 2025-04-08
Deprecation Banners for MapReduce & Samsara Pages


shajiyakhan1309:
As Apache Mahout continues to shift its focus toward Quantum Computing (Qumat), it's important to clearly communicate the deprecated status of legacy components like MapReduce and Samsara. These technologies are no longer actively maintained, but their documentation remains for historical reference.

To improve transparency and user experience, I have added consistent deprecation banners to all pages related to MapReduce and Samsara.

**Proposed Changes:**

- Added a styled HTML banner at the top of each legacy documentation page.

- Banner informs users about the deprecated status and provides a link to explore Qumat primer.

- Existing notes about "sunsetting support" have been preserved and displayed beneath the banner for continuity.

**Example Banner Text:**

`<div style="background-color: #fff3cd; color: #856404; border: 1px solid #ffeeba; padding: 15px; border-radius: 5px; margin-bottom: 20px;"> <strong>⚠️ Deprecation Notice:</strong> This component (<em>Mahout/Samsara</em>) is no longer actively maintained. It remains accessible for historical reference but is not recommended for new development. 
  Please check out <a href="https://mahout.apache.org/quantum-computing-primer/" style="color: #0c5460; text-decoration: underline;">Qumat - Mahout's Quantum Computing Primer</a> for the latest innovations.</div>`

**Benefits:**

- Enhances clarity for users navigating older documentation

 - Aligns the website with Mahout’s current development direction

 - Non-invasive update that keeps legacy content fully accessible

rawkintrevo:
This is a great idea

shajiyakhan1309:
Hello @rawkintrevo ,
I’d like to work on this issue as I’ve already implemented the changes locally. Could you please assign it to me?

guan404ming:
I think this issue has been resolved, let's close this. Feel free to reopen when needed. Thanks for the contribution!

PR #539, opened 2025-04-26
Update deprecated-docs.html

shajiyakhan1309:
This PR fixes a small but important typo in the deprecation banner used across legacy documentation pages #529  .

Previously, the banner incorrectly read:

“This component (Mahout/Samsara) is no longer actively maintained...”

It has now been corrected to:

“This component (MapReduce/Samsara) is no longer actively maintained...”

The term "Mahout" was mistakenly used instead of "MapReduce," which could cause confusion since Mahout as a whole is still active and evolving (especially with Qumat).

✅ The fix is live in the shared layout to ensure consistency across all deprecated pages.

### DIFF
--- a/website/_layouts/deprecated-docs.html
+++ b/website/_layouts/deprecated-docs.html
@@ -1,7 +1,7 @@
 ---
 layout: doc-default
 ---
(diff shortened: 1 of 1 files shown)


rawkintrevo:
Thanks @shajiyakhan1309 ! LGTM

PR #544, opened 2025-05-30
Moved the Classification, misc, and basics to the Map_Reduce_Folder

krishna-dave206:


website/Gemfile
```diff
@@ -18,6 +18,10 @@
 #
 
 source "https://rubygems.org"
+gem "logger"
```
rawkintrevo, line 21: if you're just moving files, why do you need all of these new gems?

rawkintrevo (commented):
Why do you need all of these new gems if you're just moving files?

rawkintrevo:
Also, if you're moving the files don't you want to setup redirects?

krishna-dave206:
> Also, if you're moving the files don't you want to setup redirects?

Those are just Documentation files like not the part of actual code of website... I'll redirect them when I'm gonna work with the actual website code. Documentation part's gonna stay in github only, so no need to redirect them as such

website/Gemfile
krishna-dave206: No I forgot to remove them. I had added those for my system actually.

krishna-dave206:
If edits by maintainers is allowed, can you remove those new Gems? Coz otherwise, I'll have to raise a new PR in all leaving this open.

rawkintrevo:
> Those are just Documentation files like not the part of actual code of website... I'll redirect them when I'm gonna work with the actual website code. Documentation part's gonna stay in github only, so no need to redirect them as such

This is not accurate. They are markdown files that are rendered into the website. You might want to wait a minute before you do this

rawkintrevo:
The code here looks fine, but the rendered website is built from these markdown files. 

Maybe review https://jekyllrb.com/docs/

and

https://github.com/jekyll/jekyll-redirect-from

krishna-dave206:
I've updated the links and redirected them to the MapReduceFolder. Also, checked that every link is working, @rawkintrevo Do review and merge if everything's good.

website/Gemfile
```diff
@@ -18,6 +18,10 @@
 #
 
 source "https://rubygems.org"
+<<<<<<< HEAD
```
rawkintrevo: this is failed merge, you need to clean this up

rawkintrevo:
> I've updated the links and redirected them to the MapReduceFolder

if you did this, did you forget to commit and push it?

rawkintrevo (changes requested):
The only change other than renaming files is an artifact from a local merge.

rawkintrevo:
I see your commit @krishna-dave206 , and I don't know if you're ready for another review, but did you want to put redirects in now, or is it ready to review as is?

krishna-dave206:
The redirect links are working for all the files that I moved, so there aren't any dead links. You can review it and merge this one.

rawkintrevo:
You've changed links, but you haven't set up redirects. We can talk about this f2f.

rawkintrevo:
lgtm

PR #615, opened 2025-11-12
MAHOUT-608: handle 0-qubit circuits in cirq backend to prevent crash

### LINKED ISSUE #608: [Bug] Cirq backend crashes when adding measurements to 0-qubit circuits
### Description

The Cirq backend crashes when attempting to add measurements to a circuit with zero qubits. The `execute_circuit` function in `cirq_backend.py` calls `cirq.measure(*circuit.all_qubits())` on empty circuits, which raises an error because you cannot measure an empty list of qubits.

### Steps to Reproduce

1. Create a QuMat instance with Cirq backend
2. Create an empty circuit with 0 qubits
3. Call `execute_circuit()` directly on the backend (bypassing the main QuMat wrapper)

**Example code:**
from qumat import QuMat

config = {
    "backend_name": "cirq",
    "backend_options": {"shots": 1024}
}
qumat = QuMat(config)
qumat.create_empty_circuit(0)

### Actual Behavior

Crashes with backend-specific error when attempting to call `cirq.measure(*[])` on an empty qubit list. The error message is cryptic and doesn't explain the root cause (0 qubits).

shiavm006:
### Purpose
Fixes crash in Cirq backend when executing 0-qubit circuits with measurements.

### Related Issues
closes #608 

### Changes Made
- Added guard in `execute_circuit()` to handle empty circuits before adding measurements
- Returns Cirq-formatted result `[{0: shots}]` for 0-qubit case
- Prevents crash when `cirq.measure()` is called on empty qubit list

### Technical Details
The issue occurred at line 106 where `circuit.append(cirq.measure(*circuit.all_qubits(), key="result"))` crashes on empty circuits because `circuit.all_qubits()` returns an empty list.

This fix aligns with existing behavior in the main `QuMat` class (lines 92-97 in `qumat.py`) which already handles 0-qubit circuits. Moving this logic to the backend makes it self-contained and robust.

### Breaking Changes
None. This is a bug fix that maintains existing behavior.

### Checklist
- [x] Code follows project style guidelines
- [x] Changes are minimal and focused on the issue
- [x] Implementation aligns with maintainer feedback
- [x] No breaking changes introduced

### DIFF
--- a/qumat/cirq_backend.py
+++ b/qumat/cirq_backend.py
@@ -96,6 +96,11 @@ def apply_pauli_z_gate(circuit, qubit_index):
 
 
 def execute_circuit(circuit, backend, backend_config):
+    # handle 0-qubit circuits before adding measurements
+    if not circuit.all_qubits():
+        shots = backend_config["backend_options"].get("shots", 1)
+        return [{0: shots}]
+
     # Ensure measurement is added to capture the results
     if not circuit.has_measurements():
         circuit.append(cirq.measure(*circuit.all_qubits(), key="result"))


guan404ming (approved):
LGTM, thanks!

Issue #657, opened 2025-11-29
[Roadmap] [QDP] GPU/Cuda kernel implementation


ryankert01:
This is the mega roadmap for GPU related progress, feel free to create an issue and link it to the roadmap:

1. Implement kernel #677 
2. Optimizations: #706
    -  Parallel normalization kernel
    -  Coalesced memory access patterns
    -  Warp-level optimizations
    -  Stream support for async execution
3. Implement cuda optional test (x)
    -  If no cuda device, the test skips
4. Implement suitable benchmark
    - The Scaling Test (Latency vs. Qubits) #740 
    - The DataLoader Test (Batch Throughput) #687 
5. gracefully handles OOM #688 
6. Future encoding methods: (we can do it next Quarter
    - launch_angle_encode (angle encoding)
    - launch_basis_encode (basis encoding)
    - launch_iqp_encode (IQP encoding)
7. apache license pre-commit(cuda) #684 
8. (after poc) move pre-processing from cpu to gpu

guan404ming:
Looks nice, just assigned!

rich7420:
Yes, please go ahead!

ryankert01:
Hi @guan404ming, can you help on 7.?

400Ping:
Hi, @ryankert01 may I help with 4 & 5?

ryankert01:
@400ping for 4, I just add 2 task, tell me which one you want to starts with.
for 5, go for it.

400Ping:
for 4 I want to do the Batch Throughput

400Ping:
May I take 8 as well?

ryankert01:
You can take it until poc phase are done.

guan404ming:
> Hi @guan404ming, can you help on 7.?

Sure!

PR #668, opened 2025-11-30
[QDP] improve memory initialization 

rich7420:
### Purpose of PR

This PR changes `alloc_zero` to `unsafe_alloc`.
All the tests passed.
### Related Issues or PRs

Closes #667 

### Changes Made

- [ ] Bug fix
- [x] Improving
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

- [ ] Added or updated unit tests for all changes
- [ ] Added or updated documentation for all changes
- [x] Successfully built and ran all unit tests or manual tests locally
- [ ] PR title follows "MAHOUT-XXX: Brief Description" format (if related to an issue)
- [ ] Code follows ASF guidelines

PR #677, opened 2025-12-04
[QDP] add vanilla gpu kernel

### LINKED ISSUE #685: [QDP][Feature] docker image for safe build
### Summary
Add a docker image to save me from the environment hazard
### Use Case
Explain why this feature is useful.

### Proposed Implementation
How do you propose implementing this feature?

### Alternatives Considered
Have you thought of other ways to solve the same problem?

### Additional Context
Any other relevant information or resources.

ryankert01:
[###](url) Purpose of PR

### Related Issues or PRs
Related to #657 
Closes #685

### Changes Made

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [ ] No

### Checklist

- [ ] Successfully built and ran all unit tests or manual tests locally
- [ ] Code follows ASF guidelines

### DIFF
--- a/qdp/qdp-kernels/src/amplitude.cu
+++ b/qdp/qdp-kernels/src/amplitude.cu
@@ -1,3 +1,19 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
 // Amplitude Encoding CUDA Kernel
 
 #include <cuda_runtime.h>
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -42,3 +42,11 @@ repos:
           - testing/utils/.license-header.txt
           - --comment-style
           - "//"
+      - id: insert-license
+        name: check license headers (CUDA)
+        files: \.(cu|cuh)$
+        args:
+          - --license-filepath
+          - testing/utils/.license-header.txt
+          - --comment-style
+          - "//"


qdp/qdp-kernels/src/amplitude.cu
```diff
@@ -14,17 +14,27 @@
...
+    size_t input_len,
+    size_t state_len,
+    double norm
+) {
+    size_t idx = blockIdx.x * blockDim.x + threadIdx.x;
+    if (idx >= state_len) return;
+
+    double real_part = 0.0;
+    if (idx < input_len) {
+        real_part = input[idx] / norm;
```
rich7420: I think maybe we could replace slow division (/) with fast multiplication (*).
The logic looks solid. To maximize bandwidth on RTX cards, I suggest we optimize it by replacing the division with multiplication and using double2 for vectorized memory access.

rich7420:
Thanks @ryankert01 for the patch!!! some comments left.

qdp/qdp-kernels/src/amplitude.cu
400Ping: +1 agree, Also should we explicitly guard against norm <= 0.0 on the host side and return cudaErrorInvalidValue instead of relying on undefined 0 / 0 behavior in the kernel?

ryankert01:
- I added tests
- I also added devcontainer for streamlining setup of environment. In vscode, we can use cmd+shift+P and type "dev containers: rebuild and reopen in container". Then, it will automatically opened a vscode instance that ssh-ed into that docker container.
PTAL @rich7420 @400Ping @guan404ming

400Ping (approved):
Overall LGTM

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,29 @@
...
+        "--gpus",
+        "all"
+    ],
+    // Auto mount your GitHub repo as workspace
+    "workspaceFolder": "/workspace",
+    "workspaceMount": "source=${localWorkspaceFolder},target=/workspace,type=bind",
+    // Install Python via devcontainers-features
+    "features": {
+        "ghcr.io/devcontainers/features/python:1": {
+            "version": "3.10"
```
rich7420, line 15: According to qdp/qdp-python/pyproject.toml, it seems we should set python 3.11.

rich7420:
@ryankert01 Thanks for the updates! Left some comments.

rich7420:
I think we should set 0 when `state_idx_base >= input_len ` because we don't have `Device::alloc_zeros` action in rust core. #668

rich7420:
Tests all passed.

guan404ming:
Not really familiar with this but manually test looks good. Feel free to ping me as a manual ci 😊.

.devcontainer/setup.sh
```diff
@@ -0,0 +1,32 @@
...
+# Common dev tools
+apt-get update
+apt-get install -y \
+    build-essential \
+    pkg-config \
+    git \
+    vim
+
+
+# uv
```
guan404ming, line 19: Here comes 3 pkg management tools lol, how about migrate to use uv only?

.devcontainer/devcontainer.json
ryankert01: Mahout is on 3.10, < 3.11

.devcontainer/setup.sh
ryankert01: In short, other two used by mahout main project, we have to cooperate  to solve this

.devcontainer/devcontainer.json
ryankert01: I figure I'm "uv venv python=3.11"

ryankert01:
> I think we should set 0 when state_idx_base >= input_len  because we don't have Device::alloc_zeros action in rust core. https://github.com/apache/mahout/pull/668

@rich7420 What's the motivation behind this? If `idx >= input_len`, it's not in our range.

rich7420:
> 
@ryankert01 thanks for asking!
Since we lack Device::alloc_zeros, the memory is uninitialized. We must explicitly zero out indices beyond input_len to prevent garbage data in the gpu buffer.

ryankert01:
Agree, it's best practice to init them to make sure of the deterministic for downstream event!

rich7420:
I think we could merge this PR. cc @guan404ming

ryankert01:
For possible misunderstandings, we have a discussion offline and think the concern is resolved.

guan404ming:
lg in my local test, thanks!

PR #680, opened 2025-12-04
[QDP] Integrate Apache Arrow and Parquet for data processing

guan404ming:
### Purpose of PR

Integrates Apache Arrow to enable efficient columnar data processing for quantum encoding operations. This addition provides a standardized path for handling structured data inputs alongside the existing raw &[f64] interface.

### Changes Made

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/qdp/qdp-core/src/io.rs
+++ b/qdp/qdp-core/src/io.rs
@@ -14,10 +14,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! I/O module for reading and writing quantum data using Apache Arrow and Parquet.
+//! I/O module for reading and writing quantum data
 //!
 //! This module provides efficient columnar data exchange with the data science ecosystem,
-//! enabling zero-copy interoperability with pandas, polars, and other Arrow-compatible tools.
 
 use std::fs::File;
 use std::path::Path;
@@ -63,7 +62,7 @@ pub fn arrow_to_vec_chunked(arrays: &[Float64Array]) -> Vec<f64> {
 /// Reads quantum data from a Parquet file.
 ///
 /// Expects a single column named "data" containing Float64 values.
-/// This function performs one copy from Arrow to Vec. For zero-copy access,
+/// This function performs one copy from Arrow to Vec.
 /// use `read_parquet_to_arrow` instead.
 ///
 /// # Arguments
@@ -151,7 +150,7 @@ pub fn write_parquet<P: AsRef<Path>>(
 
 /// Reads quantum data from a Parquet file as Arrow arrays.
 ///
-/// **TRUE ZERO-COPY**: Returns Arrow arrays directly from Parquet batches without any data copying.
+/// Returns Arrow arrays directly from Parquet batches.
 /// Each element in the returned Vec corresponds to one Parquet batch.
 ///
 /// Directly constructs the Arrow array from Parquet batches
@@ -160,7 +159,7 @@ pub fn write_parquet<P: AsRef<Path>>(
 /// * `path` - Path to the Parquet file
 ///
 /// # Returns
-/// Vector of Float64Arrays, one per Parquet batch (zero-copy references to Parquet data)
+/// Vector of Float64Arrays, one per Parquet batch
 pub fn read_parquet_to_arrow<P: AsRef<Path>>(path: P) -> Result<Vec<Float64Array>> {
     let file = File::open(path.as_ref()).map_err(|e| {
         MahoutError::Io(format!("Failed to open Parquet file: {}", e))
@@ -218,7 +217,7 @@ pub fn read_parquet_to_arrow<P: AsRef<Path>>(path: P) -> Result<Vec<Float64Array
 
 /// Writes an Arrow Float64Array to a Parquet file.
 ///
-/// Zero-copy write from Arrow format to Parquet.
+/// Writes from Arrow format to Parquet.
 ///
 /// # Arguments
 /// * `path` - Path to write the Parquet file
--- a/qdp/qdp-core/tests/parquet_io.rs
+++ b/qdp/qdp-core/tests/parquet_io.rs
@@ -15,8 +15,7 @@
 // limitations under the License.
 
 use qdp_core::io::{
-    arrow_to_vec_chunked, read_parquet, read_parquet_to_arrow, write_arrow_to_parquet,
-    write_parquet,
+    read_parquet, read_parquet_to_arrow, write_arrow_to_parquet, write_parquet,
 };
 use arrow::array::Float64Array;
 use std::fs;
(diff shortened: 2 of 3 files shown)


rich7420:
Thanks for the patch @guan404ming !!!
I think we could check about zero-copy parts like `arrow_to_vec` and  `Vec::push`. 
They will cause extra memory-copy and memory-reallocation.

guan404ming:
> Thanks for the patch @guan404ming !!!
I think we could check about zero-copy parts like arrow_to_vec and Vec::push.
They will cause extra memory-copy and memory-reallocation.

Nice suggestion, I've updated with much more optimized version. Thanks!

qdp/qdp-core/src/io.rs
```diff
@@ -0,0 +1,250 @@
...
+
+    // Create Arrow schema
+    let schema = Arc::new(Schema::new(vec![Field::new(
+        col_name,
+        DataType::Float64,
+        false,
+    )]));
+
+    // Create Float64Array from slice
+    let array = Float64Array::from(Vec::from(data));
```
400Ping: You could use `Float64Array::from_iter_values(data.iter().copied())` to avoid the extra allocation.

qdp/qdp-core/src/io.rs
```diff
@@ -0,0 +1,250 @@
...
+///
+/// # Arguments
+/// * `path` - Path to the Parquet file
+///
+/// # Returns
+/// Float64Array containing the data from the first column
+pub fn read_parquet_to_arrow<P: AsRef<Path>>(path: P) -> Result<Float64Array> {
+    let data = read_parquet(path)?;
+    Ok(Float64Array::from(data))
+}
```
400Ping: Directly constructing an Arrow array via `ParquetRecordBatchReader` would avoid an extra copy.

rich7420:
Looks good, but the current implementation forces a memory copy (Vec allocation) even when we want to use Arrow directly. We should refactor io.rs so that read_parquet_to_arrow is the base implementation, ensuring true zero-copy performance for the pipeline.
origin: Disk -> Arrow -> Vec (copy) -> Arrow (copy) -> GPU
we need: Disk -> Arrow -> Arrow (Zero-copy Reference) -> GPU (through Pointer)
I think so, plz correct me if I'm wrong.

400Ping:
> Looks good, but the current implementation forces a memory copy (Vec allocation) even when we want to use Arrow directly. We should refactor io.rs so that read_parquet_to_arrow is the base implementation, ensuring true zero-copy performance for the pipeline.
> 
> origin: Disk -> Arrow -> Vec (copy) -> Arrow (copy) -> GPU
> 
> we need: Disk -> Arrow -> Arrow (Zero-copy Reference) -> GPU (through Pointer)
> 
> I think so, plz correct me if I'm wrong.

I agree with @rich7420 as well.

guan404ming:
> Looks good, but the current implementation forces a memory copy (Vec allocation) even when we want to use Arrow directly. We should refactor io.rs so that read_parquet_to_arrow is the base implementation, ensuring true zero-copy performance for the pipeline.
origin: Disk -> Arrow -> Vec (copy) -> Arrow (copy) -> GPU
we need: Disk -> Arrow -> Arrow (Zero-copy Reference) -> GPU (through Pointer)
I think so, plz correct me if I'm wrong.


I think you're right, thanks for pointing out. I've updated the implementation to only.
Disk  ---->  Arrow Buffers (pointer only)  -----> GPU

qdp/qdp-core/tests/parquet_io.rs
```diff
@@ -0,0 +1,174 @@
...
+    assert_eq!(data.len(), read_data.len());
+    assert!((data[0] - read_data[0]).abs() < 1e-10);
+    assert!((data[size - 1] - read_data[size - 1]).abs() < 1e-10);
+
+    // Cleanup
+    fs::remove_file(temp_path).unwrap();
+}
+
+#[test]
+fn test_chunked_zero_copy_api() {
```
ryankert01: I know it's hard but should we verify if the mem address are the same? Looks like it only do value check.

Might be something like this: (which only ensure the pointer's in mmap)
```rs
let chunk_ptr = chunks[0].values().as_ptr() as *const u8;

assert!(chunk_ptr >= mmap_ptr);
assert!(chunk_ptr < unsafe { mmap_ptr.add(mmap_len) });
```

rich7420: @ryankert01  you're right. It seems current test verifies data integrity but not the zero-copy behavior. I think that verifying memory addresses strictly requires `unsafe` access to the underlying Arrow buffer pointers. We should add a test case that inspects `array.values().as_ptr()` to ensure it aligns with expectations, though fully verifying mmap alignment is tricky in tests.

rich7420:
@guan404ming Great work on the io refactor! Thanks!
We need to upgrade our `amplitude.rs` because:
`encode_from_parquet` currently falls back to `encode_chunked,` which merges all chunks into a huge Vec<f64> on CPU. This breaks zero-copy for large files.
We should override `encode_chunked` in AmplitudeEncoder to stream chunks directly to the GPU without merging.
What do you think? 
I could send a PR for this part.

guan404ming:
> @guan404ming Great work on the io refactor! Thanks!
We need to upgrade our amplitude.rs because:
encode_from_parquet currently falls back to encode_chunked, which merges all chunks into a huge Vec on CPU. This breaks zero-copy for large files.
We should override encode_chunked in AmplitudeEncoder to stream chunks directly to the GPU without merging.
What do you think?
I could send a PR for this part.

I previously plan to send following PR for this but forget to add in PR description. You definitely could help with this, thanks!

qdp/qdp-core/tests/parquet_io.rs
ryankert01: this can be a followup too

guan404ming:
After our offline discussion and investigation, we confirmed that Parquet cannot achieve true zero-copy from disk to memory because its data is stored in compressed and encoded form and must be decoded before use. We’ll continue keeping Parquet for now given its convenience and practicality. Also I’ve updated the TODO to migrate the encoder to a chunk-based API. 

I would follow up on Arrow IPC as a potential path toward a real zero-copy data pipeline.

guan404ming:
cc @rich7420 @ryankert01 @400Ping

rich7420:
Sure! I will improve the path of RAM -> GPU  on #689 .

guan404ming:
I'm going to merge this, please feel free to open PR to refine this. Thanks for all review and discussion!

guan404ming:
LOL

rich7420:
Speed as Rocket.

PR #751, opened 2025-12-22
[QDP] Double-buffered pinned I/O pipeline and faster Parquet decode

### LINKED ISSUE #703: [QDP] Iterator-based paged state vector and double-buffered I/O to prevent OOM
### Summary

- Single Large Memory Allocation -> adopt a paged state vector requiring kernel changes
- read_parquet_batch blocks -> introduce double-buffered async I/O to overlap CPU read with GPU compute and improve throughput.

400Ping:
### Purpose of PR

- Add an pinned host buffer pool and wire it into the dual-stream pipeline so each chunk uses double-buffered pinned staging before H2D copies (reduces malloc/free and GPU idle).

### Related Issues or PRs

Closes #703 
### Changes Made

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

- [ ] Successfully built and ran all unit tests or manual tests locally
- [ ] Code follows ASF guidelines

### DIFF
--- a/qdp/qdp-core/src/lib.rs
+++ b/qdp/qdp-core/src/lib.rs
@@ -223,6 +223,23 @@ impl QdpEngine {
                     "Sample size cannot be zero".into(),
                 ));
             }
+            if sample_size > STAGE_SIZE_ELEMENTS {
+                return Err(MahoutError::InvalidInput(format!(
+                    "Sample size {} exceeds staging buffer capacity {}",
+                    sample_size, STAGE_SIZE_ELEMENTS
+                )));
+            }
+
+            let max_samples_in_chunk = STAGE_SIZE_ELEMENTS / sample_size;
+            let mut norm_buffer = self
+                .device
+                .alloc_zeros::<f64>(max_samples_in_chunk)
+                .map_err(|e| {
+                    MahoutError::MemoryAllocation(format!(
+                        "Failed to allocate norm buffer: {:?}",
+                        e
+                    ))
+                })?;
 
             full_buf_tx
                 .send(Ok((host_buf_first, first_len)))
@@ -327,16 +344,6 @@ impl QdpEngine {
                                 .cast::<u8>()
                                 .add(offset_bytes)
                                 .cast::<std::ffi::c_void>();
-
-                            let mut norm_buffer = self
-                                .device
-                                .alloc_zeros::<f64>(samples_in_chunk)
-                                .map_err(|e| {
-                                    MahoutError::MemoryAllocation(format!(
-                                        "Failed to allocate norm buffer: {:?}",
-                                        e
-                                    ))
-                                })?;
 
                             {
                                 crate::profile_scope!("GPU::NormBatch");


400Ping:
cc @guan404ming @rich7420 @ryankert01

rich7420:
Thanks @400Ping for the patch!
1. What's the reason you define a ffi function again?
2. some tests failed locally due to tensor shape problem or you should fix test for excepted output.

400Ping:
> Thanks @400Ping for the patch!
> 
> 1. What's the reason you define a ffi function again?
> 2. some tests failed locally due to tensor shape problem or you should fix test for excepted output.

My bad just fixed it.

qdp/qdp-core/src/gpu/memory.rs
```diff
@@ -134,6 +136,78 @@ pub(crate) fn map_allocation_error(
     }
 }
 
+/// Page-locked host buffer used to enable true async DMA for H2D copies.
+#[cfg(target_os = "linux")]
+pub struct PinnedBuffer<'a, T> {
```
rich7420: nits: It seems not been used in any other files.

qdp/qdp-core/src/gpu/pipeline.rs
```diff
@@ -194,39 +216,40 @@ where
             e,
         ))?;
 
+        // Acquire pinned staging buffer and populate it with the current chunk
+        let mut pinned_buf = pinned_pool.acquire();
+        pinned_buf.as_slice_mut()[..chunk.len()].copy_from_slice(chunk);
```
rich7420, line 282: maybe we could add a check here.
like: `chunk.len() > CHUNK_SIZE_ELEMENTS`

rich7420 (commented):
@400Ping thanks for the patch!
left some comments

qdp/qdp-core/src/gpu/pipeline.rs
400Ping: done.

qdp/qdp-core/src/gpu/pipeline.rs
```diff
@@ -24,90 +24,116 @@ use std::ffi::c_void;
...
     }
 
-    /// Make compute stream wait for copy completion
-    pub unsafe fn wait_for_copy(&self) {
+    /// Make compute stream wait for the copy completion event.
+    pub unsafe fn wait_for_copy(&self, slot: usize) -> Result<()> {
         crate::profile_scope!("GPU::StreamWait");
-        unsafe {
-            cudaStreamWaitEvent(self.stream_compute.stream as *mut c_void, self.event_copy_done, 0);
+        let ret = cudaStreamWaitEvent(self.stream_compute.stream as *mut c_void, self.events_copy_done[slot], 0);
```
rich7420: should we add a check here?
like:
```
if slot >= self.events_copy_done.len() {
        return Err(MahoutError::InvalidInput(format!(
            "Event slot {} out of range (max: {})",
            slot, self.events_copy_done.len() - 1
        )));
    }
```
or something

rich7420:
I think maybe we could add some unit tests for this.

ryankert01:
We have 2 improvement in this PR. Based on the benchmark result, I'm speculating if there's one of them are not contributing to the speed improvement. What's your experience?

400Ping:
> We have 2 improvement in this PR. Based on the benchmark result, I'm speculating if there's one of them are not contributing to the speed improvement. What's your experience?

I think both have improvements, for the second one is what @rich7420 and @guan404ming suggested to change a different decompression technique to improve its performance. But I think overall it is because of the first one improving the speed improvements.

qdp/qdp-core/src/gpu/pipeline.rs
400Ping: done

400Ping:
Just tested, the second one doesn't improve much performance, going to remove it.

rich7420:
plz fix pre-commit error

400Ping:
Done, @rich7420 PTAL

400Ping:
cc @guan404ming @ryankert01

ryankert01:
please fix pre-commit. I tested locally and get a 2.8% speedup on arrow ipc case.
Will look into it next week.

guan404ming:
I agree that the comment regarding `.unwrap()` is valid.
Do we want to handle this more gracefully, or is the current panic-on-poison behavior expected? If so, we can document it.

400Ping:
> I agree that the comment regarding `.unwrap()` is valid. Do we want to handle this more gracefully, or is the current panic-on-poison behavior expected? If so, we can document it.

I think I will change the code to handle it more gracefully and add some comments to document this behavior.

guan404ming:
Need resolve conflicts, and overall looks good to me!

qdp/qdp-core/src/gpu/buffer_pool.rs
```diff
@@ -54,9 +54,11 @@ impl std::ops::DerefMut for PinnedBufferHandle {
...
     fn drop(&mut self) {
         if let Some(buf) = self.buffer.take() {
-            let mut free = self.pool.free.lock().unwrap();
-            free.push(buf);
-            self.pool.available_cv.notify_one();
+            // If the mutex is poisoned, avoid touching shared state during drop.
+            if let Ok(mut free) = self.pool.free.lock() {
+                free.push(buf);
+                self.pool.available_cv.notify_one();
+            }
```
ryankert01: If the mutex is poisoned, we can get the lock anyways during drop? Like `lock_free()` defined in L103.

qdp/qdp-core/src/gpu/pipeline.rs
```diff
@@ -191,32 +252,40 @@ where
...
-    // 4. Pipeline loop: alternate between streams for maximum overlap
+    // 4. Pipeline loop: copy on copy stream, compute on compute stream with event handoff
     for (chunk_idx, chunk) in host_data.chunks(CHUNK_SIZE_ELEMENTS).enumerate() {
-        let current_stream = streams[chunk_idx % 2];
+        let chunk_offset = global_offset;
+        let event_slot = chunk_idx % PINNED_POOL_SIZE;
 
         crate::profile_scope!("GPU::ChunkProcess");
 
+        if chunk.len() > CHUNK_SIZE_ELEMENTS {
```
rich7420: I think this function would be never used. should be removed

qdp/qdp-core/src/lib.rs
```diff
@@ -221,38 +228,13 @@ impl QdpEngine {
...
-                    "Sample size {} exceeds staging buffer capacity {} (elements)",
-                    sample_size, STAGE_SIZE_ELEMENTS
-                )));
-            }
-
-            // Reuse a single norm buffer across chunks to avoid per-chunk allocations.
-            //
-            // Important: the norm buffer must outlive the async kernels that consume it.
-            // Per-chunk allocation + drop can lead to use-after-free when the next chunk
-            // reuses the same device memory while the previous chunk is still running.
```
rich7420, line 235: I think here has a potential problem in `qdp/qdp-core/src/lib.rs`. In the `encode_from_parquet()` function in `qdp/qdp-core/src/lib.rs`, there is a critical use-after-free bug in the lifetime management of `norm_buffer`. The code allocates `norm_buffer` inside the `BatchEncode` scope at line 331-339, and asynchronously launches `launch_l2_norm_batch` and `launch_amplitude_encode_batch` kernels at line 343-375, which execute asynchronously on `ctx.stream_compute`. However, when the `BatchEncode` scope ends at line 376, `norm_buffer` is immediately dropped, and according to the comment in `pipeline.rs:336`, when `CudaSlice` is dropped, it immediately calls `cudaFree` to free GPU memory. The problem is that `ctx.sync_copy_stream()` at line 378 only synchronizes the copy stream, not the compute stream, so when `norm_buffer` is freed, kernels on the compute stream may still be executing, causing kernels to access GPU memory that has already been freed. Even though kernels execute sequentially on the same stream, if their execution time is long, `norm_buffer` may still be freed before the kernels complete. In a loop processing multiple chunks, the first chunk's `norm_buffer` is dropped before its kernel completes, and while the second chunk's kernel will wait for the first to complete, the first kernel is already accessing freed memory. The old code's comment explicitly warned about this issue, stating that the norm buffer must outlive the async kernels that consume it, and that per-chunk allocation plus drop can lead to use-after-free. Plz correct me if I'm wrong.

rich7420:
nits: in `qdp/qdp-core/src/lib.rs` 38-41 lines seem not to be used and could be removed.

qdp/qdp-core/src/lib.rs
400Ping: hmm, I agree that it is a concern. Will update this into the pr

400Ping: updated in [a99baf7](https://github.com/apache/mahout/pull/751/commits/a99baf7ea8992aad605e5ad3bfeec59ea7979d56)

rich7420: thanks for the update

rich7420:
thanks for the update!
There are many return Err() functions in this PR. but we don't actually sync compute stream. 
I think we should prevent this in other PR.

rich7420:
except that overall LGTM

400Ping:
> thanks for the update! There are many return Err() functions in this PR. but we don't actually sync compute stream. I think we should prevent this in other PR.

Should I do this in this pr or open up a follow up

rich7420:
In follow-up

guan404ming:
Let's refine it after we get into main.

PR #758, opened 2025-12-25
[QDP] add Rust Linter to pre-commit hook

### LINKED ISSUE #736: [QDP] [Feature] add clippy & rustfmt as linter.
### Summary
A brief description of the feature.

### Use Case
Explain why this feature is useful.

### Proposed Implementation
How do you propose implementing this feature?

### Alternatives Considered
Have you thought of other ways to solve the same problem?

### Additional Context
Any other relevant information or resources.

ryankert01:
### Related Issues or PRs

Closes #736

### DIFF
--- a/qdp/qdp-core/examples/dataloader_throughput.rs
+++ b/qdp/qdp-core/examples/dataloader_throughput.rs
@@ -108,9 +108,7 @@ fn main() {
                 Err(e) => {
                     eprintln!(
                         "Encode failed on batch {} (vector {}): {:?}",
-                        batch_idx,
-                        total_vectors,
-                        e
+                        batch_idx, total_vectors, e
                     );
                     return;
                 }
--- a/qdp/qdp-core/examples/nvtx_profile.rs
+++ b/qdp/qdp-core/examples/nvtx_profile.rs
@@ -14,7 +14,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-
 // NVTX profiling example
 // Run: cargo run -p qdp-core --example nvtx_profile --features observability --release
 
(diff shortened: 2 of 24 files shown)


rich7420:
I think it's pretty good. thanks for the patch @ryankert01 !
could we add into .github/workflows/pre-commit.yml ?

ryankert01:
@rich7420 I think it already ran formatting in ci! https://github.com/apache/mahout/actions/runs/20506289497/job/58938406396#step:5:1

rich7420 (approved):

rich7420:
All passed locally.
Thanks!

.pre-commit-config.yaml
```diff
@@ -50,3 +50,19 @@ repos:
           - testing/utils/.license-header.txt
           - --comment-style
           - "//"
+
+# Rust Linter
+  - repo: local
+    hooks:
+      - id: rust-fmt
```
dentiny, line 57: There's native rust format precommit hook to use: https://github.com/Mooncake-Labs/moonlink/blob/97f9bd064c2ca2d633afea7c4534e2c662b55c3f/.pre-commit-config.yaml#L6-L9
The benefit of which is we don't need pre-setup in CI (though `cargo fmt` is built in by default)

ryankert01: @dentiny We probably cannot use it because our cargo.toml is not at root, it's at `qdp/cargo.toml`. https://github.com/apache/mahout/pull/766 WDYT?

ryankert01: My bad, we actually be able to config it. Thanks for the suggestion!

PR #803, opened 2026-01-12
[QDP] Simplify Python API with unified encode() method

guan404ming:
#### Why

- API had 6 different encoding methods (encode, encode_batch, encode_tensor, encode_from_parquet, encode_from_arrow_ipc, encode_from_numpy)
- Users needed to know which method to call for each input type

#### How

- Unified encode() method - auto-detects lists, NumPy arrays, PyTorch tensors, and file paths
- Default encoding method - encoding_method="amplitude" is now the default

### Changes Made

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [x] Yes
- [ ] No

### Checklist

### DIFF
--- a/qdp/qdp-python/README.md
+++ b/qdp/qdp-python/README.md
@@ -15,11 +15,11 @@ engine = QdpEngine(0)
 
 # Encode data from Python list
 data = [0.5, 0.5, 0.5, 0.5]
-dlpack_ptr = engine.encode(data, num_qubits=2, encoding_method="amplitude")
+qtensor = engine.encode(data, num_qubits=2, encoding_method="amplitude")
 
 # Or encode from file formats
-tensor_parquet = engine.encode_from_parquet("data.parquet", 10, "amplitude")
-tensor_arrow = engine.encode_from_arrow_ipc("data.arrow", 10, "amplitude")
+tensor_parquet = engine.encode("data.parquet", 10, "amplitude")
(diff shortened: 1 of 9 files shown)


guan404ming:
cc @ryankert01 @400Ping

rich7420:
I think we need to change encode() functions in all benchmark files as well. (the rest qdp/qdp-python/benchmark/benchmark_numpy_io.py and qdp/qdp-python/benchmark/benchmark_latency.py)
And it would be better to check the tensor's shape. like, if it's 1D, use the single-sample encode path and then if it's 2D, use the batch encoding path and for anything else, throw a clear error explaining what shapes are supported.
On the other hand , maybe we could support for pathlib.Path objects by using os.fspath() to convert them to strings before processing.

guan404ming:
I've updated with three additions. Please take another look, thanks!
  1. Updated benchmark files to use unified encode() API:
  2. Added tensor shape validation with clear error messages (lib.rs):
  3. Added pathlib.Path support

rich7420:
Thanks for the update!
I think some comments been removed unnecessarily.
maybe shape validation in pytorch hasn't done yet. but it could be follow-up.

guan404ming:
Thanks, I'll send a follow up for it.

PR #805, opened 2026-01-12
[QDP] Enhance numerical safety checks in preprocessor

### LINKED ISSUE #804: [Bug] `NaN` and `Infinity` Propagation may cause numerical safety issue
### Description

The preprocessing pipeline in `qdp-core` did not validate input data for NaN (Not a Number) and Infinity values, allowing invalid numerical values to propagate through the quantum encoding pipeline.

### Steps to Reproduce

Writing a script to reproduce it:

```rust
use qdp_core::preprocessing::Preprocessor;

// This would pass validation and return NaN
let data_with_nan = vec![1.0, f64::NAN, 3.0];
let result = Preprocessor::calculate_l2_norm(&data_with_nan);
// Result: Ok(NaN)
```

### Expected Behavior

All preprocessing functions should return `Err(MahoutError::InvalidInput)` with clear error messages when NaN or Infinity values are detected.

### Actual Behavior

When input data contains NaN or Infinity values:
- `validate_input()` and `validate_batch()` would pass validation
- `calculate_l2_norm()` and `calculate_batch_l2_norms()` would return NaN/Infinity
- These invalid values would propagate to quantum simulation, causing:
  - Invalid quantum state vectors
  - GPU computation errors
  - Silent failures or crashes

### Environment
- In every environment

### Additional Context
I'm willing to help to solve the issue! 🙌

viiccwen:
### Purpose of PR
1. Added validation for NaN and Infinity values in input data and L2 norm calculations.
2. Updated correlated tests to cover these scenarios.

### Related Issues or PRs
resolve #804 

### Changes Made

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/qdp/qdp-core/tests/preprocessing.rs
+++ b/qdp/qdp-core/tests/preprocessing.rs
@@ -14,6 +14,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use approx::assert_relative_eq;
 use qdp_core::MahoutError;
 use qdp_core::preprocessing::Preprocessor;
 
@@ -53,8 +54,8 @@ fn test_validate_input_empty_data() {
 
 #[test]
 fn test_validate_input_data_too_large() {
-    let data = vec![1.0, 0.0, 0.0];
-    let result = Preprocessor::validate_input(&data, 1);
+    let data = vec![1.0, 0.0, 0.0]; // 3 elements
+    let result = Preprocessor::validate_input(&data, 1); // max size 2^1 = 2
     assert!(
         matches!(result, Err(MahoutError::InvalidInput(msg)) if msg.contains("exceeds state vector size"))
     );
@@ -63,7 +64,7 @@ fn test_validate_input_data_too_large() {
 #[test]
 fn test_validate_input_allows_partial_state() {
     let data = vec![0.5, -0.5, 0.25];
-    assert!(Preprocessor::validate_input(&data, 3).is_ok());
+    assert!(Preprocessor::validate_input(&data, 3).is_ok()); // state vector can hold up to 8 elements
 }
 
 #[test]
@@ -76,11 +77,11 @@ fn test_validate_input_max_qubits_boundary() {
 fn test_calculate_l2_norm_success() {
     let data = vec![3.0, 4.0];
     let norm = Preprocessor::calculate_l2_norm(&data).unwrap();
-    assert!((norm - 5.0).abs() < 1e-10);
+    assert_relative_eq!(norm, 5.0);
 
     let data = vec![1.0, 1.0];
     let norm = Preprocessor::calculate_l2_norm(&data).unwrap();
-    assert!((norm - 2.0_f64.sqrt()).abs() < 1e-10);
+    assert_relative_eq!(norm, 2.0_f64.sqrt());
 }
 
 #[test]
@@ -94,7 +95,7 @@ fn test_calculate_l2_norm_zero() {
 fn test_calculate_l2_norm_mixed_signs() {
     let data = vec![-3.0, 4.0];
     let norm = Preprocessor::calculate_l2_norm(&data).unwrap();
-    assert!((norm - 5.0).abs() < 1e-10);
+    assert_relative_eq!(norm, 5.0);
 }
 
 #[test]
@@ -103,7 +104,7 @@ fn test_calculate_l2_norm_matches_sequential_sum() {
     let norm_parallel = Preprocessor::calculate_l2_norm(&data).unwrap();
 
     let norm_sequential = data.iter().map(|x| x * x).sum::<f64>().sqrt();
-    assert!((norm_parallel - norm_sequential).abs() < 1e-10);
+    assert_relative_eq!(norm_parallel, norm_sequential);
 }
 
 #[test]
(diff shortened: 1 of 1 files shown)


ryankert01:
To my knowledge, pennylane do not validate, although it's a nice-to-have. It's a O(N), so I will suggest we do it later than benchmarking.

viiccwen:
Sure, what if we enable it only in dev environment? 🤔
we get strict validation during development, but it compiles away to a zero-cost op in release builds.

ryankert01:
Not sure if we have a release build, but validate norm is O(1), so I think we can have that! Not sure how it should be done though. 
Let me cc a quantum guy @rich7420

viiccwen:
Thx! 🙌

rich7420:
thanks for the patch @viiccwen !
Doing validate_input and then calculate_l2_norm in debug mode is basically two full O(N) passes. Once you get into ~20+ qubits, that can definitely make local runs feel sluggish.
I think we could do with the IEEE 754 behavior approach: NaNs/Infs will naturally "poison" the sum during the norm calculation, so you're already getting most of that validation for free:

- Any NaN in the input → sum of squares becomes NaN
- Any Inf in the input → sum becomes Inf

So yeah, in most cases we can drop the separate check_numerical_safety loop and just rely on the end-of-calculate_l2_norm guard (lines 86–96) to catch:

1. Invalid inputs (NaN/Inf)
2. Overflow during accumulation

That gives you a single O(N) pass, keeps things cleaner, and speeds up debug builds without affecting release.
WDYT?

viiccwen:
> thanks for the patch @viiccwen ! Doing validate_input and then calculate_l2_norm in debug mode is basically two full O(N) passes. Once you get into ~20+ qubits, that can definitely make local runs feel sluggish. I think we could do with the IEEE 754 behavior approach: NaNs/Infs will naturally "poison" the sum during the norm calculation, so you're already getting most of that validation for free:
> 
> * Any NaN in the input → sum of squares becomes NaN
> * Any Inf in the input → sum becomes Inf
> 
> So yeah, in most cases we can drop the separate check_numerical_safety loop and just rely on the end-of-calculate_l2_norm guard (lines 86–96) to catch:
> 
> 1. Invalid inputs (NaN/Inf)
> 2. Overflow during accumulation
> 
> That gives you a single O(N) pass, keeps things cleaner, and speeds up debug builds without affecting release. WDYT?

Agree. I hadn’t considered IEEE 754 propagation behavior. It'll be removed later.
And also agree to keep the nan/inf guard, although we've checked nan/inf situation later stages:

https://github.com/apache/mahout/blob/bff0b88ef38808263a65a24a8d2ee2352bd0e1b4/qdp/qdp-kernels/src/amplitude.cu#L147-L149

It should help catch invalid values early in the pipeline as well : ) 
Thanks for the suggestion!

viiccwen:
@ryankert01 Thx for pointing that out, and apologies for not discussing this first before opening the PR.

Sure, even if we don’t have a clear release/debug distinction here, an O(1) validation step could still make sense.
I should have raised this for discussion earlier instead of jumping straight into an implementation.

I’ll make sure to bring these kinds of changes up for discussion first and avoid wasting others time : ) 
Thanks again for the feedback!

qdp/qdp-core/tests/preprocessing.rs
```diff
@@ -53,8 +53,8 @@ fn test_validate_input_empty_data() {
 
 #[test]
 fn test_validate_input_data_too_large() {
-    let data = vec![1.0, 0.0, 0.0]; // 3 elements
-    let result = Preprocessor::validate_input(&data, 1); // max size 2^1 = 2
+    let data = vec![1.0, 0.0, 0.0];
+    let result = Preprocessor::validate_input(&data, 1);
```
ryankert01: whats the reason of removing comments?

qdp/qdp-core/tests/preprocessing.rs
```diff
@@ -63,7 +63,7 @@ fn test_validate_input_data_too_large() {
 #[test]
 fn test_validate_input_allows_partial_state() {
     let data = vec![0.5, -0.5, 0.25];
-    assert!(Preprocessor::validate_input(&data, 3).is_ok()); // state vector can hold up to 8 elements
+    assert!(Preprocessor::validate_input(&data, 3).is_ok());
```
ryankert01: ditto.

qdp/qdp-core/tests/preprocessing.rs
```diff
@@ -105,3 +105,48 @@ fn test_calculate_l2_norm_matches_sequential_sum() {
...
+    }
+}
+
+#[test]
+fn test_calculate_batch_l2_norms_success() {
+    let batch_data = vec![3.0, 4.0, 5.0, 12.0];
+    let norms = Preprocessor::calculate_batch_l2_norms(&batch_data, 2, 2).unwrap();
+    assert_eq!(norms.len(), 2);
+    assert!((norms[0] - 5.0).abs() < 1e-10); // sqrt(3^2 + 4^2) = 5
+    assert!((norms[1] - 13.0).abs() < 1e-10); // sqrt(5^2 + 12^2) = 13
```
ryankert01: nit: although completely understand what its doing, we can also use `approx::assert_relative_eq` to enhance its readability.

ryankert01 (commented):
some nits

qdp/qdp-core/tests/preprocessing.rs
viiccwen: Thx for the suggestion!

I followed the existing pattern used earlier in the code, but I agree that using `approx::assert_relative_eq` would improve readability. 
I’ll add `approx` as a dev-dependency and update the tests accordingly.

qdp/qdp-core/tests/preprocessing.rs
viiccwen: thx for pointing out. I'll undo it later.

qdp/qdp-core/tests/preprocessing.rs
viiccwen: thx for pointing out. I'll undo it later.

PR #844, opened 2026-01-16
MAHOUT-743: Add regression test for async amplitude chunk bounds

### LINKED ISSUE #743: [QDP] Fix async amplitude pipeline out-of-bounds write
### Summary

When processing large data in chunks, launch_amplitude_encode() is called with state_len (full state vector size) instead of chunk_len (actual chunk size). This causes the kernel to write state_len elements starting from the chunk offset, which exceeds bounds for chunks after the first one.

Rutuja123-dos:
### Purpose of PR
This PR adds a regression test for the async amplitude encoding chunk bounds issue (#743).
The Problem: When performing amplitude encoding in chunks, the kernel must respect the specific chunk_len for the offset provided, rather than the full state_len. Out-of-bounds writes can occur if the bounds are not correctly constrained.
### Related Issues or PRs
Closes #743

### Changes Made

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/testing/test_amplitude_encoding.py
+++ b/testing/test_amplitude_encoding.py
@@ -1,0 +1,52 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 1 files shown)


ryankert01 (changes requested):
We don't test external package

Rutuja123-dos:
Thanks for the feedback @ryankert01. You’re right, I see now that this test is just checking NumPy and not the actual project logic.
I’m going to look back at the source code to find the exact function that’s causing the out-of-bounds issue. I'll rewrite the test to call that function instead and update my PR. Thanks for pointing that out

400Ping:
Please actually take a good look of the codebase before you submit a PR for it, you also added some unnecessary comments and your reply seems very GPT like answer. I do encourage using AI agents stuff to write code, but you need to understand the context and what it is producing. Don't just upload whatever it says like a no-brainer.

Rutuja123-dos:
Closing this for now. I've realized my local build environment is a bit too limited for these GPU/Rust changes and it's becoming a major blocker. I’m going to step back from this one and look for some other areas to contribute instead. Thanks for the help.

PR #839, opened 2026-01-16
[QDP] Add basis encoding

guan404ming:
### Purpose of PR

Implement basis encoding to map integer inputs directly to computational basis states (e.g., 3 $\to$ |011>) on the GPU.
I would follow up another one to add this in benchmark after this one get merged.

### Related Issues or PRs

- Related to #799

### Changes Made

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

- [x] Added or updated unit tests for all changes
- [x] Added or updated documentation for all changes
- [x] Successfully built and ran all unit tests or manual tests locally
- [ ] PR title follows "MAHOUT-XXX: Brief Description" format (if related to an issue)
- [x] Code follows ASF guidelines

PR #851, opened 2026-01-17
[QDP] Add streaming basis encoding

guan404ming:
### Purpose of PR

- Adds ChunkEncoder trait for unified streaming encoding
- Implements AmplitudeEncoder and BasisEncoder using shared pipeline
- Adds Float64 column support in Parquet reader for basis encoding
- Updates e2e benchmark to support both encoding methods

### Related Issues or PRs

- Related to #839 

### Changes Made

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [x] Yes
- [ ] No

### Checklist

### DIFF
--- a/qdp/qdp-core/src/encoding/basis.rs
+++ b/qdp/qdp-core/src/encoding/basis.rs
@@ -138,10 +138,11 @@ impl ChunkEncoder for BasisEncoder {
                 state.indices_cpu.push(index);
             }
 
-            // Copy indices to pre-allocated GPU buffer
+            // Copy indices to pre-allocated GPU buffer (slice to match actual chunk size)
+            let mut gpu_slice = state.indices_gpu.slice_mut(0..samples_in_chunk);
             engine
                 .device
-                .htod_sync_copy_into(&state.indices_cpu, &mut state.indices_gpu)
+                .htod_sync_copy_into(&state.indices_cpu, &mut gpu_slice)
                 .map_err(|e| {
                     MahoutError::MemoryAllocation(format!(
                         "Failed to upload basis indices to GPU: {:?}",
--- a/qdp/qdp-core/src/encoding/mod.rs
+++ b/qdp/qdp-core/src/encoding/mod.rs
@@ -148,11 +148,19 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
     const PIPELINE_EVENT_SLOTS: usize = 2;
     let ctx = PipelineContext::new(&engine.device, PIPELINE_EVENT_SLOTS)?;
 
-    // Double-buffered device staging
-    let dev_in_a = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
-        .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
-    let dev_in_b = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
-        .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
+    // Check if encoder needs staging buffers before allocating
+    let needs_staging_copy = encoder.needs_staging_copy();
+
+    // Double-buffered device staging (only allocated if needed)
+    let dev_staging = if needs_staging_copy {
+        let dev_in_a = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
+            .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
+        let dev_in_b = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
+            .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
+        Some((dev_in_a, dev_in_b))
+    } else {
+        None
+    };
 
     // Channel setup for async IO
     let (full_buf_tx, full_buf_rx): FullBufferChannel = sync_channel(2);
@@ -173,7 +181,6 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
     let mut encoder_state = encoder.init_state(engine, sample_size, num_qubits)?;
 
     let state_len = 1 << num_qubits;
-    let needs_staging_copy = encoder.needs_staging_copy();
 
     // Send first buffer to processing
     full_buf_tx
@@ -241,17 +248,23 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
         let samples_in_chunk = current_len / sample_size;
         if samples_in_chunk > 0 {
             let event_slot = if use_dev_a { 0 } else { 1 };
-            let dev_ptr = if use_dev_a {
-                *dev_in_a.device_ptr()
-            } else {
-                *dev_in_b.device_ptr()
-            };
+            // Get device pointer from staging buffers (0 if not allocated)
+            let dev_ptr = dev_staging
+                .as_ref()
+                .map(|(a, b)| {
+                    if use_dev_a {
+                        *a.device_ptr()
+                    } else {
+                        *b.device_ptr()
+                    }
+                })
+                .unwrap_or(0);
 
             unsafe {
                 crate::profile_scope!("GPU::Dispatch");
 
-                // Async copy to device
-                if needs_staging_copy {
+                // Async copy to device (only if staging buffers are allocated)
+                if dev_staging.is_some() {
                     ctx.async_copy_to_device(
                         host_buffer.ptr() as *const c_void,
                         dev_ptr as *mut c_void,
@@ -301,7 +314,7 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
                     global_sample_offset,
                 )?;
 
-                if needs_staging_copy {
+                if dev_staging.is_some() {
                     ctx.sync_copy_stream()?;
                 }
             }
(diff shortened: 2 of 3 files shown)


guan404ming:
cc @400Ping

qdp/qdp-core/src/encoding/mod.rs
```diff
@@ -0,0 +1,309 @@
...
+                crate::profile_scope!("GPU::Dispatch");
+
+                // Async copy to device
+                ctx.async_copy_to_device(
+                    host_buffer.ptr() as *const c_void,
+                    dev_ptr as *mut c_void,
+                    current_len,
+                )?;
+                ctx.record_copy_done(event_slot)?;
+                ctx.wait_for_copy(event_slot)?;
```
rich7420: It seems in `stream_encode()` always performs an H2D staging copy (`async_copy_to_device` + wait) for every chunk, but `BasisEncoder::encode_chunk()` does not use `dev_ptr` at all. It costs basis streaming pays an unnecessary H2D memcpy + synchronization per chunk.

qdp/qdp-core/src/encoding/basis.rs
```diff
@@ -0,0 +1,144 @@
...
+        _dev_ptr: u64,
+        samples_in_chunk: usize,
+        _sample_size: usize,
+        state_ptr_offset: *mut c_void,
+        state_len: usize,
+        num_qubits: usize,
+        global_sample_offset: usize,
+    ) -> Result<()> {
+        unsafe {
+            crate::profile_scope!("GPU::BatchEncode");
```
rich7420, line 101: It seems in this part per-chunk cpu allocations + per-chunk gpu allocations + sync H2D copies add significant overhead and reduce throughput. Maybe we could try reuse cpu/gpu buffer in this part ?

qdp/qdp-python/benchmark/benchmark_e2e.py
```diff
@@ -330,13 +344,7 @@ def run_mahout_arrow(engine, n_qubits, n_samples, encoding_method: str = "amplit
     start_time = time.perf_counter()
 
     arrow_encode_start = time.perf_counter()
-    try:
-        qtensor = engine.encode(ARROW_FILE, n_qubits, encoding_method)
-    except RuntimeError as e:
-        if "Only amplitude encoding supported" in str(e):
-            print("  Basis encoding not supported for streaming from Arrow, skipping.")
-            return 0.0, None
-        raise
```
rich7420, line 339: In basis mode the Arrow IPC file is written as a scalar `Float64` column, but Mahout’s Arrow reader path only supports `List<Float64>` / `FixedSizeList<Float64>`. 
After removing this old skip guard, this may become a crash here.
Maybe we could try to write basis Arrow IPC as `FixedSizeList<Float64>(len=1)`

rich7420 (commented):
some comments left

guan404ming:
Thanks, I've applied the review comments. I would try run all benchmark tomorrow.

qdp/qdp-core/src/encoding/mod.rs
```diff
@@ -0,0 +1,321 @@
...
+    engine
+        .device
+        .synchronize()
+        .map_err(|e| MahoutError::Cuda(format!("{:?}", e)))?;
+    io_handle
+        .join()
+        .map_err(|e| MahoutError::Io(format!("IO thread panicked: {:?}", e)))?;
+
+    let dlpack_ptr = total_state_vector.to_dlpack();
+    Ok(dlpack_ptr)
```
viiccwen: I found `stream_encode` has multiple early `return Err(...)`, that skip the cleanup code at the end of the function, may lead to undefined behavior if GPU operations continue after resources are dropped.

qdp/qdp-core/src/encoding/basis.rs
```diff
@@ -0,0 +1,173 @@
...
+            // Copy indices to pre-allocated GPU buffer
+            engine
+                .device
+                .htod_sync_copy_into(&state.indices_cpu, &mut state.indices_gpu)
+                .map_err(|e| {
+                    MahoutError::MemoryAllocation(format!(
+                        "Failed to upload basis indices to GPU: {:?}",
+                        e
+                    ))
+                })?;
```
viiccwen: Since basis encoding doesn't need staging buffer copy, we could use async copy

guan404ming: I think even the "async" staging path waits immediately after copying:                                                                                               

```
ctx.async_copy_to_device(...)?;                                                                                                                              
ctx.record_copy_done(event_slot)?;                                                                                                                           
ctx.wait_for_copy(event_slot)?;  // waits anyway                                                                                                             
```

Since the kernel needs the indices uploaded first, sync copy has the same semantics and is simpler. Thus, there is no benefit to changing it based on my understanding. WDYT?

qdp/qdp-core/src/encoding/mod.rs
guan404ming: Early returns do skip cleanup. However, I think "undefined behavior" is overstated. In Rust, the IO thread detaches (not joins) and GPU resources get dropped. It's bad practice but not UB. I've added a scope guard as a proper fix. Please take another look, thanks!

viiccwen: thx for correcting! LGTM here.

qdp/qdp-core/src/encoding/basis.rs
viiccwen: Yes, ur right. In the current implementation, the async version is more complex for no clear benefit.
If we refactor this to a true dual‑stream pipeline (prefetch copy(i+1) while compute(i) runs, similar to run_dual_stream_pipeline), we can reintroduce the async/event machinery with the wait moved out of the immediate hot path so that we actually get measurable overlap. WDYT?

guan404ming: Make sense to me. Let's do it in another follow up. I want to make each PR's scope smaller to ensure a faster and more effective code review process~

viiccwen: Sure, I think so.

400Ping: I can do the follow up if needed

qdp/qdp-core/src/encoding/mod.rs
```diff
@@ -0,0 +1,359 @@
...
+    // Allocate output state vector
+    let total_state_vector = GpuStateVector::new_batch(&engine.device, num_samples, num_qubits)?;
+    const PIPELINE_EVENT_SLOTS: usize = 2;
+    let ctx = PipelineContext::new(&engine.device, PIPELINE_EVENT_SLOTS)?;
+
+    // Double-buffered device staging
+    let dev_in_a = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
+        .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
+    let dev_in_b = unsafe { engine.device.alloc::<f64>(STAGE_SIZE_ELEMENTS) }
+        .map_err(|e| MahoutError::MemoryAllocation(format!("{:?}", e)))?;
```
400Ping: These 2x512MB device staging buffers are allocated unconditionally; please gate allocation on needs_staging_copy==true (e.g., Option<CudaSlice>) to avoid ~1GB VRAM usage for basis encoding.

guan404ming: Nice catch, I've updated

guan404ming:
```
======================================================================
E2E BENCHMARK: 16 Qubits, 200 Samples
======================================================================

[PennyLane] Full Pipeline (Disk -> GPU) - basis encoding...
  IO Time: 0.0092 s
/home/gmchiu/Documents/GitHub/mahout/qdp/qdp-python/benchmark/benchmark_e2e.py:263: UserWarning: Casting complex values to real discards the imaginary part (Triggered internally at /pytorch/aten/src/ATen/native/Copy.cpp:309.)
  state_gpu = state_cpu.to("cuda", dtype=torch.float32)
  Total Time: 0.3249 s

[Mahout-Parquet] Full Pipeline (Parquet -> GPU)...
  Parquet->GPU (IO+Encode): 0.4275 s
  DLPack conversion: 0.0001 s
  Convert to float32: 0.0180 s
  Total Time: 0.4596 s

[Mahout-Arrow] Full Pipeline (Arrow IPC -> GPU)...
  Arrow->GPU (IO+Encode): 0.0014 s
  DLPack conversion: 0.0001 s
  Convert to float32: 0.0011 s
  Total Time: 0.0068 s

======================================================================
E2E LATENCY (Lower is Better)
Samples: 200, Qubits: 16
======================================================================
Mahout-Arrow         0.0068 s
PennyLane            0.3249 s
Mahout-Parquet       0.4596 s
----------------------------------------------------------------------
Speedup vs PennyLane:      47.53x

======================================================================
VERIFICATION (Mahout-Arrow vs Mahout-Parquet)
======================================================================
Max Probability Difference: 0.00e+00
Max Amplitude Difference:   0.00e+00
>> SUCCESS: Quantum States Match!

======================================================================
VERIFICATION (Mahout-Arrow vs PennyLane)
======================================================================
Max Probability Difference: 0.00e+00
Max Amplitude Difference:   0.00e+00
>> SUCCESS: Quantum States Match!

======================================================================
VERIFICATION (Mahout-Parquet vs PennyLane)
======================================================================
Max Probability Difference: 0.00e+00
Max Amplitude Difference:   0.00e+00
>> SUCCESS: Quantum States Match!
```

guan404ming:
Thanks for all the reviews!

PR #866, opened 2026-01-18
MAHOUT-865: Add release.md

guan404ming:
### Purpose of PR
Add detailed release step documentation

ref:
- https://github.com/apache/airflow/blob/main/dev/README_RELEASE_AIRFLOW.md
- https://release-test.apache.org/docs/

### Related Issues or PRs

- Closes #865

### Changes Made

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [x] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

- [ ] Added or updated unit tests for all changes
- [x] Added or updated documentation for all changes
- [x] Successfully built and ran all unit tests or manual tests locally
- [x] PR title follows "MAHOUT-XXX: Brief Description" format (if related to an issue)
- [x] Code follows ASF guidelines

PR #875, opened 2026-01-19
Add qumat dependency to qumat-qdp package

guan404ming:
### Purpose of PR

- Add `qumat>0.4.0` as a dependency of `qumat-qdp` to ensure users always get the main package when installing the extension refer to https://github.com/apache/airflow/tree/main/providers/airbyte                          
- Bump qumat version to `0.5.0.dev0` for development                                                                                                  
- Update lock files

### Related Issues or PRs

- Related to #866

### Changes Made

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [x] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "qumat"
-version = "0.4"
+version = "0.5.0.dev0"
 description = "A library for composing quantum machine learning."
 authors = [{ name = "Apache Mahout", email = "dev@mahout.apache.org" }]
 license = "Apache-2.0"
--- a/qdp/qdp-python/pyproject.toml
+++ b/qdp/qdp-python/pyproject.toml
@@ -5,6 +5,7 @@ build-backend = "maturin"
 [project]
 name = "qumat-qdp"
 requires-python = ">=3.10,<3.13"
+dependencies = ["qumat>0.4.0"]
 classifiers = [
     "Programming Language :: Rust",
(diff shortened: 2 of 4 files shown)


guan404ming:
cc @400Ping @viiccwen

rawkintrevo:
Is qumat-qdp becoming a distinct package from qumat? I saw this in the release.md too. If so, let's get a discussion thread open about merging the packages. 

Edit: Responded here, then saw comment in release.md PR. But same q, why not make it an extra such that the end user would `pip install qumat[qdp]` totally OK if there's a reason, but it should probably be captured in an ADR or something.

guan404ming:
Yes, pip install qumat[qdp] is exactly what we're going for. However, the extra still requires a separate package because extras are just dependencies on other packages, not embedded optional code. Also the extra still requires a separate package since 
                                                                                                                                                         
- qumat uses hatchling (pure Python)                                                                                                                  
- qumat-qdp uses maturin (Rust extension)                                                                                                             
- A single pyproject.toml can only have one build backend 

This follows the same pattern as pydantic + pydantic-core.                                                                                            
Happy to document this in an ADR if you think it would be helpful.

guan404ming:
Here is the discussion thread https://github.com/apache/mahout/discussions/876

guan404ming:
Feel free to open pr to refine this, thanks!

Issue #726, opened 2025-12-12
[QDP] gpu tensor zero-copy via dlpack protocol


rich7420:
### Summary
This issue asks gpu tensor support using dlpack protocol for zero-copy data transfer. This is critical for training loop performance where tensors are already on gpu.
- extend encode_tensor() to handle gpu tensors
- add dlpack extraction
- add gpu pointer validation
- add stream synchronization
- add direct gpu processing

If anything important is missing, please leave a comment.

400Ping:
take.

PR #881, opened 2026-01-20
MAHOUT-878 Add CUDA Torch Tensor Support for QDP Python Binding

### LINKED ISSUE #878: [QDP] Support torch CUDA tensor inputs directly avoid (GPU→CPU→GPU copies)
QDP’s Python binding currently rejects torch CUDA tensors and only supports CPU tensors, forcing users to copy GPU→CPU→GPU (or failing). Add a CUDA-tensor path that keeps data on GPU and runs encoding without host staging.

Current behavior
- Passing a torch CUDA tensor to `QdpEngine.encode(...)` raises an error like:
  - “Only CPU tensors are currently supported for this path…”
- Users must `.cpu()` first, which adds large overhead and breaks GPU-only pipelines.

Expected behavior
- `QdpEngine.encode(torch_tensor_cuda, ...)` works for:
  - 1D tensors (single sample)
  - 2D tensors (batch)
- No GPU→CPU copy is required.
- Clear errors for unsupported cases (dtype, contiguity, device mismatch).

ryankert01:
### Purpose of PR
Add zero-copy encoding support for PyTorch CUDA tensors in `QdpEngine.encode()`. Previously, CUDA tensors were rejected with "Only CPU tensors are currently supported", forcing users to call `.cpu()` first and incurring a GPU→CPU→GPU copy overhead.

### Motivation

When working with PyTorch CUDA tensors, the previous workflow required:
```python
# Before: GPU → CPU → GPU (slow)
result = engine.encode(cuda_tensor.cpu(), num_qubits, "amplitude")
```

With this change:
```python
# After: GPU → GPU (zero-copy)
result = engine.encode(cuda_tensor, num_qubits, "amplitude")
```

This eliminates unnecessary memory transfers and improves performance for GPU-based ML pipelines.

### Changes

| File | Changes |
|------|---------|
| `qdp-core/src/lib.rs` | Added `unsafe` methods `encode_from_gpu_ptr()` and `encode_batch_from_gpu_ptr()` |
| `qdp-python/src/lib.rs` | Added CUDA tensor detection, validation, DLPack extraction; modified `encode()` to route CUDA tensors |
| `qdp-core/.../amplitude.rs` | Made `calculate_inv_norm_gpu()` public |
| `testing/.../test_bindings.py` | Added 14 new CUDA tensor tests |

### Error Messages

| Condition | Message |
|-----------|---------|
| Non-float64 dtype | `CUDA tensor must have dtype float64, got {dtype}. Use tensor.to(torch.float64)` |
| Non-contiguous | `CUDA tensor must be contiguous. Use tensor.contiguous()` |
| Empty tensor | `CUDA tensor cannot be empty` |
| Device mismatch | `Device mismatch: tensor is on cuda:{X}, but engine is on cuda:{Y}. Move tensor with tensor.to('cuda:{Y}')` |
| Zero/NaN/Inf values | `Input data has zero or non-finite norm (contains NaN, Inf, or all zeros)` |
| Unsupported encoding | `CUDA tensor encoding currently only supports 'amplitude' method, got '{method}'...` |

### Follow-ups
1. support float32
2. supports more than amplitude encoding (eg. `"angle"`, `"basis"`)
3. (maybe) refactoring to extend `QuantumEncoder` trait

### Testing

```bash
pytest testing/qdp/test_bindings.py -v
```

### Related Issues or PRs

Closes #878

### Changes Made

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/qdp/qdp-core/src/lib.rs
+++ b/qdp/qdp-core/src/lib.rs
@@ -28,7 +28,7 @@ pub mod tf_proto;
 #[macro_use]
 mod profiling;
 
-pub use error::{cuda_error_to_string, MahoutError, Result};
+pub use error::{MahoutError, Result, cuda_error_to_string};
 pub use gpu::memory::Precision;
 
 use std::sync::Arc;


ryankert01:
PTAL @shiavm006 @400Ping @viiccwen

ryankert01:
```
  ⎿  ============================================================                                                                         
     CUDA Tensor Encoding Speed Comparison                                                                                                
     ============================================================                                                                         
     Batch=  100, Features=  64 | CPU:    0.30ms | CUDA:    0.12ms | Speedup: 2.45x                                                       
     Batch=  100, Features= 256 | CPU:    0.21ms | CUDA:    0.46ms | Speedup: 0.45x                                                       
     Batch=  100, Features=1024 | CPU:    0.67ms | CUDA:    0.17ms | Speedup: 3.94x                                                       
     Batch= 1000, Features=  64 | CPU:    3.82ms | CUDA:    0.48ms | Speedup: 7.93x                                                       
     Batch= 1000, Features= 256 | CPU:    2.90ms | CUDA:    0.26ms | Speedup: 11.18x                                                      
     Batch= 1000, Features=1024 | CPU:    6.11ms | CUDA:    0.57ms | Speedup: 10.78x                                                      
     Batch=10000, Features=  64 | CPU:    4.26ms | CUDA:    2.03ms | Speedup: 2.10x                                                       
     Batch=10000, Features= 256 | CPU:   10.09ms | CUDA:    4.78ms | Speedup: 2.11x                                                       
     Batch=10000, Features=1024 | CPU:   27.10ms | CUDA:   10.42ms | Speedup: 2.60x                                                       
     ============================================================   
```

400Ping:
I left this for @CheyuWu to do, but sure.

400Ping:
We should probably add a if you checked known issues when you create an issue

ryankert01:
Sorry @CheyuWu not noticing it. Would you perhaps want to take on some of the followups? There are three of them.

400Ping:
I will finish the float32 along with what's left of https://github.com/apache/mahout/issues/726 if no one takes it

CheyuWu:
> Sorry @CheyuWu not noticing it. Would you perhaps want to take on some of the followups? There are three of them.

No worries at all! I can help review the code and take care of the remaining follow-ups.

qdp/qdp-python/src/lib.rs
```diff
@@ -152,7 +152,7 @@ fn is_pytorch_tensor(obj: &Bound<'_, PyAny>) -> PyResult<bool> {
     Ok(module_name == "torch")
 }
 
-/// Helper to validate tensor
+/// Helper to validate CPU tensor
 fn validate_tensor(tensor: &Bound<'_, PyAny>) -> PyResult<()> {
```
CheyuWu, line 156: I think we can remove this function, since we already support GPUs.

qdp/qdp-python/src/lib.rs
```diff
@@ -321,6 +460,78 @@ impl QdpEngine {
...
+                            "Unsupported CUDA tensor shape: {}D. Expected 1D tensor for single \
+                             sample encoding or 2D tensor (batch_size, features) for batch encoding.",
+                            ndim
+                        )));
+                    }
+                }
+            }
+
+            // CPU tensor path (existing code)
             validate_tensor(data)?;
```
CheyuWu, line 496: I think we don't need this anymore

qdp/qdp-python/src/lib.rs
ryankert01: Do you mean we can identify tensor on gpu or cpu, if cpu, we can:`tensor.to("cuda")` and go to gpu tensor path? It can be another followup to refactor make it simple.

CheyuWu: Sorry, I didn’t explain this very clearly on my side.

Currently, `validate_tensor` is only called at this [location](https://github.com/apache/mahout/pull/881/files#diff-2127e48c371dd5082474a4e975ac908065dc7b4a5736dfaa9f0efbfc95c8981aR535).
The only thing `validate_tensor` does is call `is_pytorch_tensor`.
Therefore, I don’t think it’s necessary to keep this function, since `is_pytorch_tensor` has already been called earlier in [here](https://github.com/apache/mahout/pull/881/files#diff-2127e48c371dd5082474a4e975ac908065dc7b4a5736dfaa9f0efbfc95c8981aR462).

ryankert01: Good catch, got it!

 In `validate_tensor` has a part of it redundant as it rechecks `is_pytorch_tensor(tensor)`, but it can also catch other hardware tensor like MPS, XLA, HPU, etc. if any. 
We can refactor it and remove `is_pytorch_tensor(tensor)` as a followup. and the function name probably should be renamed to `validate_tensor_cpu` or something.

qdp/qdp-core/src/gpu/encodings/amplitude.rs
```diff
@@ -95,11 +95,15 @@ impl QuantumEncoder for AmplitudeEncoder {
...
                 // GPU-accelerated norm for medium+ inputs, CPU fallback for tiny payloads
                 let inv_norm = if host_data.len() >= GPU_NORM_THRESHOLD {
-                    Self::calculate_inv_norm_gpu(
-                        _device,
-                        *input_slice.device_ptr() as *const f64,
-                        host_data.len(),
-                    )?
+                    // SAFETY: input_slice was just allocated and copied from host_data,
+                    // so the pointer is valid and contains host_data.len() elements
+                    unsafe {
```
shiavm006, line 100: We’re calling AmplitudeEncoder::calculate_inv_norm_gpu twice here: once directly and once within the unsafe block. Since the function is now pub unsafe fn, the first call should no longer compile, and even if it did, we’d be doing redundant work. We should keep only the unsafe call with the safety justification comment

qdp/qdp-python/src/lib.rs
```diff
@@ -171,6 +171,145 @@ fn validate_tensor(tensor: &Bound<'_, PyAny>) -> PyResult<()> {
...
+
+/// Extract GPU pointer from PyTorch tensor's __dlpack__() capsule
+///
+/// # Safety
+/// The returned `data_ptr` points to GPU memory owned by the source tensor.
+/// The caller must ensure the source tensor remains alive and unmodified
+/// for the entire duration that `data_ptr` is in use. Python's GIL ensures
+/// the tensor won't be garbage collected during `encode()`, but the caller
+/// must not deallocate or resize the tensor while encoding is in progress.
+fn extract_dlpack_tensor(_py: Python<'_>, tensor: &Bound<'_, PyAny>) -> PyResult<DLPackTensorInfo> {
```
shiavm006: Using __dlpack__ to get the device pointer is convenient, but we’re currently not calling the DLManagedTensor.deleter, and we rename the capsule to used_dltensor, so the deleter won’t run at GC either. That likely means a small leak per call on PyTorch’s side.

shiavm006: use tensor.data_ptr() + tensor.shape directly and avoid DLPack for this direction

viiccwen:
I'll dive into refactor `encode` (`qdp/qdp-python/src/lib.rs`) after this PR merged.

qdp/qdp-core/src/gpu/encodings/amplitude.rs
```diff
@@ -411,8 +415,20 @@ impl AmplitudeEncoder {
...
+    ///
+    /// # Returns
+    /// The inverse L2 norm (1/||x||_2) of the input data
+    ///
+    /// # Safety
+    /// The caller must ensure `input_ptr` points to valid GPU memory containing
+    /// at least `len` f64 elements on the same device as `device`.
     #[cfg(target_os = "linux")]
-    fn calculate_inv_norm_gpu(
+    pub unsafe fn calculate_inv_norm_gpu(
```
viiccwen: Given this is an unsafe primitive that’s only used inside the core GPU pipeline, I’d lean towards making it `pub(crate)` for now and exposing only safe encoding APIs at the crate boundary.

viiccwen: If we promote GPU norm computation later as a public feature (e.g. via a QuantumEncoder extension trait), maybe we can do a follow-up about design a dedicated safe wrapper and keep this function as an internal building block. WDYT?

viiccwen (commented):
overall LGTM

qdp/qdp-python/src/lib.rs
viiccwen: I raise an issue for #888. @ryankert01 would u like to fix it in current PR?

ryankert01: yes, but I'm not sure if I've solved it all, I only solve related ones.

ryankert01:
fixed, PTAL @viiccwen @shiavm006 @CheyuWu @400Ping

qdp/qdp-core/src/lib.rs
```diff
@@ -300,6 +300,269 @@ impl QdpEngine {
...
+    /// # Returns
+    /// DLPack pointer for zero-copy PyTorch integration
+    ///
+    /// # Safety
+    /// The input pointer must:
+    /// - Point to valid GPU memory on the same device as the engine
+    /// - Contain at least `input_len` f64 elements
+    /// - Remain valid for the duration of this call
+    #[cfg(target_os = "linux")]
+    pub unsafe fn encode_from_gpu_ptr(
```
rich7420, line 329: The `encode_from_gpu_ptr` and `encode_batch_from_gpu_ptr` methods in `QdpEngine` have some code duplication with `AmplitudeEncoder::encode()` and `AmplitudeEncoder::encode_batch()`.
I think maybe we could add `encode_from_gpu_ptr` and `encode_batch_from_gpu_ptr` methods to the `QuantumEncoder` trait (in `qdp/qdp-core/src/gpu/encodings/mod.rs`). And then implement these methods in `AmplitudeEncoder` (move the logic from `QdpEngine`) and simplify `QdpEngine::encode_from_gpu_ptr` to just get the encoder and call its method. WDYT?

qdp/qdp-core/src/lib.rs
```diff
@@ -300,6 +300,269 @@ impl QdpEngine {
...
+                    state_ptr as *mut std::ffi::c_void,
+                    input_len,
+                    state_len,
+                    inv_norm,
+                    std::ptr::null_mut(), // default stream
+                )
+            };
+
+            if ret != 0 {
+                return Err(MahoutError::KernelLaunch(format!(
```
rich7420, line 396: I think here we could use the `cuda_error_to_string` function that's already in the codebase to convert error codes into readable names. This is already done correctly in `AmplitudeEncoder`, so just copy that pattern.

qdp/qdp-core/src/lib.rs
```diff
@@ -300,6 +300,269 @@ impl QdpEngine {
...
+                    input_batch_d,
+                    num_samples,
+                    sample_size,
+                    *buffer.device_ptr_mut() as *mut f64,
+                    std::ptr::null_mut(), // default stream
+                )
+            };
+
+            if ret != 0 {
+                return Err(MahoutError::KernelLaunch(format!(
```
rich7420, line 501: same here

qdp/qdp-core/src/lib.rs
```diff
@@ -300,6 +300,269 @@ impl QdpEngine {
...
+                    *inv_norms_gpu.device_ptr() as *const f64,
+                    num_samples,
+                    sample_size,
+                    state_len,
+                    std::ptr::null_mut(), // default stream
+                )
+            };
+
+            if ret != 0 {
+                return Err(MahoutError::KernelLaunch(format!(
```
rich7420, line 550: and here

rich7420 (commented):
@ryankert01 thanks for the patch! left some comments:

qdp/qdp-core/src/lib.rs
ryankert01: done

qdp/qdp-core/src/lib.rs
ryankert01: This is actually one of my followups! @lib.rs L310

ryankert01:
cc @rich7420 @guan404ming

qdp/qdp-core/src/lib.rs
rich7420: no problem!

rich7420:
LGTM , @ryankert01  thanks for the update !

ryankert01:
opened issues! @CheyuWu @400Ping, feel free to comment take
https://github.com/apache/mahout/issues/912
https://github.com/apache/mahout/issues/911
https://github.com/apache/mahout/issues/907

PR #941, opened 2026-01-25
[QDP] Fix misaligned memory access in L2 norm batch kernel

### LINKED ISSUE #940: [Bug] Misaligned double2 vector loads in batched L2norm kernel when the length of sample is odd
### Description

The current batched L2 norm CUDA kernel performs vectorized loads starting from input_batch + base, where:

https://github.com/apache/mahout/blob/d5fc18bd01bd4fbaa6c947252b142f38c9d8b720/qdp/qdp-kernels/src/amplitude.cu#L367-L368

When sample_len is odd and sample_idx is also odd, base becomes odd, meaning input_batch + base is only 8-byte aligned. This violates the alignment requirements of vectorized loads: double2 requires 16-byte alignment.

Although this does not cause incorrect results, it introduces avoidable performance penalties and makes the kernel’s memory access behavior architecture-dependent.

reference: [comment](https://github.com/apache/mahout/pull/918#discussion_r2724245973)

viiccwen:
### Purpose of PR

The `l2_norm_batch_kernel` uses vectorized `double2` loads (16-byte alignment required) to compute L2 norms for batched samples. However, when `sample_len` is odd, the base offset `base = sample_idx * sample_len` may be odd, causing `input_batch + base` to be misaligned (offset by 8 bytes instead of 16 bytes).

Added alignment handling logic to ensure vectorized loads always start from properly aligned addresses:

1. **Alignment check**: Check if `base` is even (16-byte aligned for `double2`)
2. **Peel first element**: If misaligned, handle the first element separately using a single `double` load
3. **Vectorized loads**: Start vectorized `double2` loads from the second element (`base + 1`), which is guaranteed to be 16-byte aligned
4. **Index separation**: Strictly separate "aligned sub-array index" from "original sample index" to avoid confusion and bugs

### Related Issues or PRs
closes #940 

### Changes Made

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/qdp/qdp-kernels/tests/amplitude_encode.rs
+++ b/qdp/qdp-kernels/tests/amplitude_encode.rs
@@ -623,6 +623,66 @@ fn test_l2_norm_batch_kernel_stream() {
 }
 
 #[test]
+#[cfg(target_os = "linux")]
+fn test_l2_norm_batch_kernel_odd_length_alignment() {
+    println!(
+        "Testing batched L2 norm reduction with odd-length samples to exercise alignment handling..."
+    );
+
+    let device = match CudaDevice::new(0) {
+        Ok(d) => d,
+        Err(_) => {
+            println!("SKIP: No CUDA device available");
+            return;
+        }
+    };
+
+    let batch: Vec<f64> = vec![
+        3.0, 4.0, 5.0, // norm = sqrt(9+16+25) = sqrt(50) ≈ 7.071
+        1.0, 1.0, 1.0, // norm = sqrt(3) ≈ 1.732 (misaligned base)
+        5.0, 12.0, 13.0, // norm = sqrt(25+144+169) = sqrt(338) ≈ 18.385
+    ];
+    let num_samples = 3;
+    let sample_len = 3;
+
+    let expected: Vec<f64> = vec![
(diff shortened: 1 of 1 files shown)


qdp/qdp-kernels/src/amplitude.cu
```diff
@@ -371,17 +371,32 @@ __global__ void l2_norm_batch_kernel(
 
     double local_sum = 0.0;
 
+    // Alignment peel for double2 (16B) loads
```
ryankert01, line 374: the comments should include more details
```suggestion
    // Alignment peel for double2 (16B) loads
    // 1. **Alignment check**: Check if `base` is even (16-byte aligned for `double2`)
    // 2. **Peel first element**: If misaligned, handle the first element separately using a single `double` load
    // 3. **Vectorized loads**: Start vectorized `double2` loads from the second element (`base + 1`), which is guaranteed to be 16-byte aligned
```

ryankert01:
Overall make sense, just not the standard way to implement this.

Issue #969, opened 2026-01-29
[QDP] User Python API interface features and refactoring — Roadmap


rich7420:
### Summary
This roadmap defines the implementation path for QDP pipeline tuning (PR3): instrumenting the pipeline with observability, collecting baselines, introducing hardware-aware chunk and pool sizing, and replacing periodic sync with event-based buffer reuse—so that throughput and host-to-device overlap improve in a measurable, data-driven way.

#### Phase 2:  Baseline and Data Collection #972 
Capture reproducible pre-optimization baselines for later comparison.
#### Phase 5:  refactor Reader and add ChunkedReader
Remove periodic `sync_copy_stream()` and reuse pinned buffers only after per-slot copy-complete events.

### Phase Order and Dependencies

- **Phase 2** depends only on Phase 1; run first and publish baseline.
- **Phase 3** can be developed in parallel with Phase 4; recommend merging C3 before C4 so C4 uses configurable pool/chunk.
- **Phase 4** should be validated by re-running Phase 2 and comparing overlap, throughput, and latency.

### Alternatives Considered
Have you thought of other ways to solve the same problem?

### Additional Context
Any other relevant information or resources.

400Ping:
I am interested in phase2 & 3.

rich7420:
Sorry, I’m working on it. But if you could review those PRs, I’d be grateful.

400Ping:
This seems to completed, closing this

PR #1000, opened 2026-01-31
[QDP] Add a Quantum Data Loader and API refactor

rich7420:
### Purpose of PR

Introduces a **stateful Rust iterator** (Quantum Data Loader) that yields one encoded batch (DLPack) per step, so Python can drive encoding with `for qt in loader:` instead of a closed benchmark loop. The same Rust core supports both **benchmark** (full pipeline, stats only) and **data loader** (batch-by-batch iteration). The public Python API is moved out of `benchmark/` into a package **`qumat_qdp`**; benchmark scripts import from `qumat_qdp` and remain the primary consumers.

### qdp-core (Rust)

- **`pipeline_runner.rs`** (new): `PipelineConfig`, `PipelineRunResult`, `run_throughput_pipeline`, `run_latency_pipeline`; shared `vector_len`, `generate_batch`, `fill_sample`. Adds **`DataSource`** enum (Synthetic only; File reserved for Phase 2) and **`PipelineIterator`** with `new_synthetic(engine, config)` and `next_batch(&mut self) -> Result<Option<*mut DLManagedTensor>>` reusing `encode_batch`.
- **`lib.rs`**: `QdpEngine` implements **`Clone`**; re-exports `DataSource`, `PipelineIterator`, `PipelineConfig`, `PipelineRunResult`, `run_throughput_pipeline`, `run_latency_pipeline` (Linux only).
- **`gpu/encodings/amplitude.rs`**: `run_amplitude_dual_stream_pipeline` and exposure for dual-stream encode path used by benchmarks.

### qdp-python (Rust bindings)

- **`SendPtr`** wrapper: wraps `*mut DLManagedTensor` so the raw pointer can cross `py.allow_threads` (closure return must be `Send`). Used only to release GIL during encode.
- **`PyQuantumLoader`** (`#[pyclass]`): holds `Option<PipelineIterator>`. Implements Python iterator protocol: `__iter__` returns `self`; `__next__` takes iterator out with `take()`, runs `next_batch()` inside `py.allow_threads`, then restores the iterator or returns a `QuantumTensor`. Raises `StopIteration` when exhausted. Stub on non-Linux.
- **`QdpEngine::create_synthetic_loader(...)`**: builds `PipelineConfig`, calls `PipelineIterator::new_synthetic(engine.clone(), config)`, returns `PyQuantumLoader`. Stub on non-Linux.
- **`run_throughput_pipeline_py`**: pipeline run is executed inside **`py.allow_threads`** so the full benchmark loop runs with GIL released (replacing previous detach/allow_threads usage as appropriate).

### qumat_qdp (Python package)

- **`qumat_qdp/`** at project root: `__init__.py` (re-exports `QdpEngine`, `QuantumTensor`, `QdpBenchmark`, `ThroughputResult`, `LatencyResult`, `QuantumDataLoader`, `run_throughput_pipeline_py`), **`api.py`** (QdpBenchmark, ThroughputResult, LatencyResult; calls `_qdp.run_throughput_pipeline_py`), **`loader.py`** (QuantumDataLoader builder; `__iter__` calls `engine.create_synthetic_loader` and returns the Rust iterator).
- **`pyproject.toml`**: **`python-source = "."`** so the root `qumat_qdp` package is included in the wheel.

### Benchmark scripts

- **Imports**: All benchmark scripts now use `from qumat_qdp import ...` (or re-export). **`benchmark/api.py`** and **`benchmark/loader.py`** only re-export from `qumat_qdp` for backward compatibility.
- **Import path**: Each script inserts the project root into `sys.path` before importing `qumat_qdp`, so `uv run python benchmark/run_pipeline_baseline.py` (and similar) works without extra `PYTHONPATH`.
- **`benchmark_loader_throughput.py`** (new): runs throughput by iterating `QuantumDataLoader` (`for qt in loader`) and compares with `QdpBenchmark.run_throughput()` (full Rust pipeline).

## Behaviour

- **Benchmark path**: Unchanged from caller’s perspective. `QdpBenchmark(device_id=0).qubits(16).batches(100, 64).run_throughput()` still runs the full pipeline in Rust with GIL released via `run_throughput_pipeline_py`.
- **Loader path**: `QuantumDataLoader(device_id=0).qubits(16).batches(100, 64).source_synthetic()` then `for qt in loader:` yields one `QuantumTensor` (batch) per iteration; GIL is released during each `next_batch()` in Rust.

## Testing

- Existing benchmark flow: `uv run python benchmark/run_pipeline_baseline.py`, `benchmark_throughput.py`, `benchmark_latency.py`.
- Loader throughput: `uv run python benchmark/benchmark_loader_throughput.py` (compares loader iteration vs full pipeline).

### Related Issues or PRs

Related to #969 

### Changes Made

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [x] Yes
- [ ] No

### Checklist

### DIFF
--- a/qdp/qdp-python/qumat_qdp/loader.py
+++ b/qdp/qdp-python/qumat_qdp/loader.py
@@ -35,6 +35,10 @@ from typing import TYPE_CHECKING, Iterator, Optional
 if TYPE_CHECKING:
     import _qdp  # noqa: F401 -- for type checkers only
 
+# Rust interface expects seed as Option<u64>: non-negative and <= 2^64 - 1.
+# Ref: qdp-core PipelineConfig seed: Option<u64>
+_U64_MAX = 2**64 - 1
+
 # Lazy import _qdp at runtime until __iter__ is used; TYPE_CHECKING import above
 # is for type checkers only so they can resolve "_qdp.*" annotations if needed.
 
@@ -44,6 +48,41 @@ def _get_qdp():
     import _qdp as m
 
     return m
+
+
+def _validate_loader_args(
+    *,
+    device_id: int,
+    num_qubits: int,
+    batch_size: int,
+    total_batches: int,
+    encoding_method: str,
+    seed: Optional[int],
+) -> None:
+    """Validate arguments before passing to Rust (PipelineConfig / create_synthetic_loader)."""
+    if device_id < 0:
+        raise ValueError(f"device_id must be non-negative, got {device_id!r}")
+    if not isinstance(num_qubits, int) or num_qubits < 1:
+        raise ValueError(f"num_qubits must be a positive integer, got {num_qubits!r}")
+    if not isinstance(batch_size, int) or batch_size < 1:
+        raise ValueError(f"batch_size must be a positive integer, got {batch_size!r}")
+    if not isinstance(total_batches, int) or total_batches < 1:
+        raise ValueError(
+            f"total_batches must be a positive integer, got {total_batches!r}"
+        )
+    if not encoding_method or not isinstance(encoding_method, str):
+        raise ValueError(
+            f"encoding_method must be a non-empty string, got {encoding_method!r}"
+        )
+    if seed is not None:
+        if not isinstance(seed, int):
+            raise ValueError(
+                f"seed must be None or an integer, got {type(seed).__name__!r}"
+            )
+        if seed < 0 or seed > _U64_MAX:
+            raise ValueError(
+                f"seed must be in range [0, {_U64_MAX}] (Rust u64), got {seed!r}"
+            )
 
 
 class QuantumDataLoader:
@@ -63,6 +102,14 @@ class QuantumDataLoader:
         encoding_method: str = "amplitude",
         seed: Optional[int] = None,
     ) -> None:
+        _validate_loader_args(
+            device_id=device_id,
+            num_qubits=num_qubits,
+            batch_size=batch_size,
+            total_batches=total_batches,
+            encoding_method=encoding_method,
+            seed=seed,
+        )
         self._device_id = device_id
         self._num_qubits = num_qubits
         self._batch_size = batch_size
@@ -72,16 +119,26 @@ class QuantumDataLoader:
 
     def qubits(self, n: int) -> QuantumDataLoader:
         """Set number of qubits. Returns self for chaining."""
+        if not isinstance(n, int) or n < 1:
+            raise ValueError(f"num_qubits must be a positive integer, got {n!r}")
         self._num_qubits = n
         return self
 
     def encoding(self, method: str) -> QuantumDataLoader:
         """Set encoding method (e.g. 'amplitude', 'angle', 'basis'). Returns self."""
+        if not method or not isinstance(method, str):
+            raise ValueError(
+                f"encoding_method must be a non-empty string, got {method!r}"
+            )
         self._encoding_method = method
         return self
 
     def batches(self, total: int, size: int = 64) -> QuantumDataLoader:
         """Set total number of batches and batch size. Returns self."""
+        if not isinstance(total, int) or total < 1:
+            raise ValueError(f"total_batches must be a positive integer, got {total!r}")
+        if not isinstance(size, int) or size < 1:
+            raise ValueError(f"batch_size must be a positive integer, got {size!r}")
         self._total_batches = total
         self._batch_size = size
         return self
@@ -92,16 +149,37 @@ class QuantumDataLoader:
     ) -> QuantumDataLoader:
         """Use synthetic data source (default). Optionally override total_batches. Returns self."""
         if total_batches is not None:
+            if not isinstance(total_batches, int) or total_batches < 1:
+                raise ValueError(
+                    f"total_batches must be a positive integer, got {total_batches!r}"
+                )
             self._total_batches = total_batches
         return self
 
     def seed(self, s: Optional[int] = None) -> QuantumDataLoader:
-        """Set RNG seed for reproducible synthetic data. Returns self."""
+        """Set RNG seed for reproducible synthetic data (must fit Rust u64: 0 <= seed <= 2^64-1). Returns self."""
+        if s is not None:
+            if not isinstance(s, int):
+                raise ValueError(
+                    f"seed must be None or an integer, got {type(s).__name__!r}"
+                )
+            if s < 0 or s > _U64_MAX:
+                raise ValueError(
+                    f"seed must be in range [0, {_U64_MAX}] (Rust u64), got {s!r}"
+                )
         self._seed = s
         return self
 
     def __iter__(self) -> Iterator[object]:
         """Return Rust-backed iterator that yields one QuantumTensor per batch."""
+        _validate_loader_args(
+            device_id=self._device_id,
+            num_qubits=self._num_qubits,
+            batch_size=self._batch_size,
+            total_batches=self._total_batches,
+            encoding_method=self._encoding_method,
+            seed=self._seed,
+        )
         qdp = _get_qdp()
         QdpEngine = getattr(qdp, "QdpEngine", None)
         if QdpEngine is None:
--- a/qdp/qdp-python/src/lib.rs
+++ b/qdp/qdp-python/src/lib.rs
@@ -340,12 +340,8 @@ fn get_torch_cuda_stream_ptr(tensor: &Bound<'_, PyAny>) -> PyResult<*mut c_void>
         )));
     }
 
+    // PyTorch default stream can report cuda_stream as 0; treat as valid (Rust sync is no-op for null).
     let stream_ptr: u64 = stream.getattr("cuda_stream")?.extract()?;
-    if stream_ptr == 0 {
-        return Err(PyRuntimeError::new_err(
-            "PyTorch returned a null CUDA stream pointer",
-        ));
-    }
     Ok(stream_ptr as *mut c_void)
 }
 
@@ -1152,7 +1148,7 @@ impl QdpEngine {
         seed: Option<u64>,
     ) -> PyResult<PyQuantumLoader> {
         let config = PipelineConfig {
-            device_id: 0,
+            device_id: self.engine.device().ordinal(),
             num_qubits,
             batch_size,
             total_batches,
--- a/testing/qdp/test_benchmark_api.py
+++ b/testing/qdp/test_benchmark_api.py
@@ -86,9 +86,30 @@ def test_qdp_benchmark_run_latency():
 
 
 @requires_qdp
-def test_qdp_benchmark_validation():
-    """QdpBenchmark.run_throughput() raises if qubits/batches not set."""
+@pytest.mark.parametrize("method", ["run_throughput", "run_latency"])
+def test_qdp_benchmark_validation(method):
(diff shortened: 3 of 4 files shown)


rich7420:
Before
Running throughput: 5 trials (qubits=16, batch_size=64, prefetch=16, batches=200)
Throughput: median=1153.5 vec/s, p95=1279.0 vec/s
Running latency: 5 trials
Latency: median=0.853 ms/vec, p95=0.874 ms/vec

After
Running throughput: 5 trials (qubits=16, batch_size=64, prefetch=16, batches=200)
Throughput: median=2541.9 vec/s, p95=3000.6 vec/s
Running latency: 5 trials
Latency: median=0.373 ms/vec, p95=0.401 ms/vec

rich7420:
cc @viiccwen

ryankert01:
Love the idea! When will 0.6.0 release dates be? I think we should land it faster.

> Btw, can you share the benchmark result? Do it increase speed for decrease speed? I might be able to help with that.
edit: sorry I didn't saw it at first place!

guan404ming:
We could prepare it recently I think. We could iterate it fast.

qdp/qdp-python/benchmark/benchmark_latency.py
```diff
@@ -26,12 +26,20 @@
...
 import argparse
+import sys
 import time
+from pathlib import Path
 
 import torch
 
-from _qdp import QdpEngine
-from utils import normalize_batch, prefetched_batches
+# Add project root to path so qumat_qdp is importable when run as script
```
guan404ming: I think we could maybe handle this more better by using the internal qdp export here?

qdp/qdp-python/benchmark/benchmark_loader_throughput.py
```diff
@@ -0,0 +1,114 @@
...
+"""
+
+from __future__ import annotations
+
+import argparse
+import sys
+import time
+from pathlib import Path
+
+# Add project root to path so qumat_qdp is importable when run as script
```
guan404ming: ditto

qdp/qdp-python/benchmark/benchmark_throughput.py
```diff
@@ -28,13 +28,21 @@
...
+import sys
 import time
+from pathlib import Path
 
 import numpy as np
 import torch
 
-from _qdp import QdpEngine
-from utils import normalize_batch, prefetched_batches
+# Add project root to path so qumat_qdp is importable when run as script
```
guan404ming: ditto

qdp/qdp-python/qumat_qdp/__init__.py
```diff
@@ -0,0 +1,55 @@
...
+    from qumat_qdp import QdpBenchmark, ThroughputResult, LatencyResult
+    from qumat_qdp import QuantumDataLoader
+"""
+
+from __future__ import annotations
+
+# Rust extension (built by maturin)
+import _qdp
+
+from qumat_qdp.api import (
```
guan404ming, line 36: Do we want to export these apis to our users?

qdp/qdp-python/qumat_qdp/loader.py
```diff
@@ -0,0 +1,123 @@
...
+
+from __future__ import annotations
+
+from typing import Iterator, Optional
+
+# Lazy import _qdp until __iter__ is used
+_qdp: Optional[object] = None
+
+
+def _get_qdp():
```
guan404ming, line 47: I'm not really sure if we need this. Could you explain more?

.pre-commit-config.yaml
```diff
@@ -71,7 +71,7 @@ repos:
         name: cargo clippy
         language: system
         pass_filenames: false
-        entry: bash -c 'LIBTORCH_USE_PYTORCH=1 cargo clippy "$@"' --
+        entry: bash -c 'LIBTORCH_USE_PYTORCH=1 LIBTORCH_BYPASS_VERSION_CHECK=1 cargo clippy "$@"' --
```
guan404ming: I think we use local torch instead of libtorch here. Could you help share why we need this change?

rich7420: oh , cuz my local pre-commit errors.
I could remove this part.

qdp/qdp-python/benchmark/benchmark_latency.py
rich7420: Done. I removed the `sys.path` hack and rely on the package being importable when run from the qdp-python directory.

qdp/qdp-python/qumat_qdp/__init__.py
rich7420: I think we should intend to expose `QdpEngine` and `QuantumTensor` (and optionally `run_throughput_pipeline_py`) for advanced users who need direct control; `QdpBenchmark` and `QuantumDataLoader` remain the recommended high-level API. WDYT?

qdp/qdp-python/qumat_qdp/loader.py
rich7420: yeah , you're right, we don't need it in this part

qdp/qdp-python/benchmark/benchmark_loader_throughput.py
rich7420: done

qdp/qdp-python/benchmark/benchmark_throughput.py
rich7420: done

qdp/qdp-python/qumat_qdp/loader.py
rich7420: I had intended the lazy import to defer loading the Rust extension until first use, but with the current design it doesn’t actually achieve that

qdp/qdp-python/qumat_qdp/__init__.py
guan404ming: Make sense, thanks!

qdp/qdp-python/benchmark/benchmark_latency.py
guan404ming: Thanks!

guan404ming:
Please help resolve ci error thanks!

qdp/qdp-python/qumat_qdp/loader.py
guan404ming: After some investigation, I found this piece of code seems work for this issue. Could you help test it?

```
from functools import lru_cache

@lru_cache(maxsize=1)
def get_qdp():
    import _qdp
    return _qdp
```

guan404ming: I tested it seems not work. Maybe we could also try to move the import `_qdp` to `if TYPE_CHECKING:` and rewrite the type def to `def func(x: "_qdp.Type"):` and see if it works

rich7420: got it! thanks for the advice!

viiccwen:
It's failed with testing (CI didn't trigger these tests due to lack of CUDA environment).

```
testing/qdp/test_bindings.py::test_encode_cuda_tensor[data_shape0-expected_shape0-1] FAILED                                                                [  3%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor[data_shape1-expected_shape1-3] FAILED                                                                [  3%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_preserves_input[data_shape0-False] FAILED                                                            [  3%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_preserves_input[data_shape1-True] FAILED                                                             [  4%]
testing/qdp/test_bindings.py::test_encode_3d_rejected[cuda_tensor-Unsupported CUDA tensor shape: 3D] FAILED                                                [  4%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_non_finite_values[<lambda>-zeros] FAILED                                                             [  4%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_non_finite_values[<lambda>-NaN] FAILED                                                               [  5%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_non_finite_values[<lambda>-Inf] FAILED                                                               [  5%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_output_dtype[float32-expected_dtype0] FAILED                                                         [  5%]
testing/qdp/test_bindings.py::test_encode_cuda_tensor_output_dtype[float64-expected_dtype1] FAILED                                                         [  5%]
```

qdp/qdp-python/src/lib.rs
viiccwen: We don't need this since PyTorch default stream can report cuda_stream as 0.
I don't know whether ur change cause this or not. Just found.
Then we can pass the testing. : D

qdp/qdp-python/src/lib.rs
```diff
@@ -1038,20 +1112,193 @@ impl QdpEngine {
...
+    ) -> PyResult<PyQuantumLoader> {
+        let config = PipelineConfig {
+            device_id: 0,
+            num_qubits,
+            batch_size,
+            total_batches,
+            encoding_method: encoding_method.to_string(),
+            seed,
+            warmup_batches: 0,
+        };
```
viiccwen, line 1158: device_id uses hard-coded 0, config should equals to use actually.

testing/qdp/test_benchmark_api.py
```diff
@@ -0,0 +1,94 @@
...
+    assert result.latency_ms_per_vector > 0
+
+
+@requires_qdp
+def test_qdp_benchmark_validation():
+    """QdpBenchmark.run_throughput() raises if qubits/batches not set."""
+    import api
+
+    with pytest.raises(ValueError, match="qubits and batches"):
+        api.QdpBenchmark(device_id=0).run_throughput()
```
viiccwen: lacks of testing of `run_latency()` if qubits/batches not set.

qdp/qdp-python/qumat_qdp/loader.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+
+    def __init__(
+        self,
+        device_id: int = 0,
+        num_qubits: int = 16,
+        batch_size: int = 64,
+        total_batches: int = 100,
+        encoding_method: str = "amplitude",
+        seed: Optional[int] = None,
```
viiccwen, line 103: seed in Rust interface is: `seed: Option<u64>`, but in Python is: `seed: Optional[int] = None`.

maybe we can add validate guard to raise `ValueError` with clear description.

testing/qdp/test_benchmark_api.py
ryankert01: In test we also don't have `device_id`propagation.

qdp/qdp-python/qumat_qdp/loader.py
ryankert01: Validate every args user enters and raise proper error might be a good one.

rich7420: updated! plz take a another look

ryankert01:
Overall lgtm

rich7420:
thanks for the review @guan404ming , @ryankert01 and @viiccwen !

PR #1026, opened 2026-02-07
[QDP] I/O errors: expose underlying cause via Error::source()

rich7420:
### Purpose of PR

Adopt Rust’s standard error convention for I/O: add a new `MahoutError::IoWithSource` variant that wraps `std::io::Error` and implements `Error::source()`. Callers can now downcast or walk the cause chain for file/read errors instead of losing the underlying error in a string.
#### References
- **Rust:** [std::error::Error](https://doc.rust-lang.org/std/error/trait.Error.html) ([Error::source](https://doc.rust-lang.org/std/error/trait.Error.html#method.source))
- **thiserror:** [docs.rs thiserror](https://docs.rs/thiserror/latest/thiserror/) (derive, `#[source]`, struct variant for custom message + source)

### Changes Made

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Breaking Changes

- [ ] Yes
- [x] No

### Checklist

### DIFF
--- a/qdp/qdp-core/src/readers/arrow_ipc.rs
+++ b/qdp/qdp-core/src/readers/arrow_ipc.rs
@@ -49,11 +49,14 @@ impl ArrowIPCReader {
                 )));
             }
             Err(e) => {
-                return Err(MahoutError::Io(format!(
-                    "Failed to check if Arrow IPC file exists at {}: {}",
-                    path.display(),
-                    e
-                )));
+                return Err(MahoutError::IoWithSource {
+                    message: format!(
+                        "Failed to check if Arrow IPC file exists at {}: {}",
+                        path.display(),
+                        e
+                    ),
+                    source: e,
+                });
             }
             Ok(true) => {}
         }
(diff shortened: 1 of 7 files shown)


ryankert01:
good idea

qdp/qdp-core/src/readers/arrow_ipc.rs
```diff
@@ -49,11 +49,14 @@ impl ArrowIPCReader {
...
-                    e
-                )));
+                return Err(MahoutError::IoWithSource {
+                    message: format!(
+                        "Failed to check if Arrow IPC file exists at {}: {}",
+                        path.display(),
+                        e
+                    ),
+                    source: e,
+                });
```
ryankert01, line 59: Will this have double error message?

ryankert01: (this is not a blocker)

rich7420: Good catch. With thiserror, only the message field is used for Display / to_string(), so the underlying source is not automatically appended. That means there’s no double message when we just print the error once.

rich7420:
@ryankert01 and @viiccwen thanks for the review!

PR #1042, opened 2026-02-11
Replace mypy with ty for static type checking

guan404ming:
### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

  
Switch to Astral's ty type checker for faster, more consistent tooling alongside ruff and uv. 

### How

- Replace mypy pre-commit hook with ty, add ty to dev dependencies
- Fix type errors: np.array annotation, dict.get nullability, reduce overload, stale type: ignore comments
- Configure ty to exclude CI/benchmark dirs and allow _qdp native extension imports

## Checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -23,11 +23,14 @@ repos:
         files: \.py$
 
   # Static type checking
-  - repo: https://github.com/pre-commit/mirrors-mypy
-    rev: v1.19.0
+  - repo: local
     hooks:
-      - id: mypy
-        args: ["--ignore-missing-imports"]
+      - id: ty
+        name: ty check
+        entry: uv run ty check .
+        language: system
+        pass_filenames: false
+        types: [python]
 
   # Check Apache license headers
   - repo: https://github.com/lucas-c/pre-commit-hooks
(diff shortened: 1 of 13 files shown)


400Ping:
Not familiar with it as well

guan404ming:
FYI: https://docs.astral.sh/ty/, which is widely adopted by open source projects like `tvm-ffi`

viiccwen (approved):

guan404ming:
Thanks!

rich7420:
A quick question: pre-commit runs `ty check`, but `.github/workflows/python-testing.yml` does not.

rich7420:
should we add it as well?

guan404ming:
Oops

guan404ming:
Lol, I could add it in a follow up. Thanks for the nice catch.

rich7420:
got it, thanks

guan404ming:
After confirmation, we already have ci running for pre-commit, which contains the `ty` check. I think we could skip that. WDYT?

rich7420:
sure!

PR #1076, opened 2026-02-22
avoid mutating user backend config

### LINKED ISSUE #1070: [Bug] Backend config mutation causes state leakage between QuMat instances
### What
The `QuMat` class mutates the user-provided `backend_config` dictionary, causing state leakage between instances that share the same config object. This violates encapsulation principles and can lead to incorrect results in production code.

### Location

- **Primary location:** `mahout/qumat/qumat.py:330`  
- **Secondary location:** `mahout/qumat/qumat.py:386` (same issue in `get_final_state_vector()`)

shiavm006:
Stop QuMat from changing the caller’s backend_config dict. Before the fix, calling execute_circuit() or get_final_state_vector() added a parameter_values key to that dict, so the same config could be mutated when reused for multiple QuMat instances. The fix copies the config in __init__ so only QuMat’s internal copy is updated and the user’s dict stays unchanged.

### Related Issues

Closes #1070

### DIFF
--- a/qumat/qumat.py
+++ b/qumat/qumat.py
@@ -60,12 +60,12 @@ class QuMat:
                 "Please provide a dictionary with backend-specific options "
             )
 
-        self.backend_config = backend_config
-        self.backend_name = backend_config["backend_name"]
+        self.backend_config = backend_config.copy()
+        self.backend_name = self.backend_config["backend_name"]
         self.backend_module = import_module(
             f".{self.backend_name}_backend", package="qumat"
         )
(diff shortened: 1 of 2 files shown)


shiavm006:
cc @ryankert01 @rawkintrevo @guan404ming

ryankert01 (approved):

ryankert01:
Ty @shiavm006 for the contributions

guan404ming (approved):

PR #1093, opened 2026-02-27
Enable CUDA tensor validation for iqp and iqp-z encodings

SuyashParmar:
QdpEngine.encode() supports iqp and iqp-z in qdp-core, but the CUDA tensor validation path in pytorch.rs currently rejects these methods and only allows amplitude, angle, and basis.

This creates inconsistent behavior:

CPU/list/NumPy paths can use iqp/iqp-z
CUDA tensor path fails early with an unsupported-method error
Expected behavior
CUDA tensor inputs should allow iqp and iqp-z when dtype/shape/device constraints are satisfied (same style as other supported encodings).

### DIFF
--- a/qdp/qdp-python/src/pytorch.rs
+++ b/qdp/qdp-python/src/pytorch.rs
@@ -182,9 +182,18 @@ pub fn validate_cuda_tensor_for_encoding(
                 )));
             }
         }
+        "iqp" | "iqp-z" => {
+            if !dtype_str_lower.contains("float64") {
+                return Err(PyRuntimeError::new_err(format!(
+                    "CUDA tensor must have dtype float64 for {} encoding, got {}. \
+                     Use tensor.to(torch.float64)",
+                    method, dtype_str
+                )));
+            }
+        }
         _ => {
             return Err(PyRuntimeError::new_err(format!(
-                "CUDA tensor encoding currently only supports 'amplitude', 'angle', or 'basis' methods, got '{}'. \
+                "CUDA tensor encoding currently only supports 'amplitude', 'angle', 'basis', 'iqp', or 'iqp-z' methods, got '{}'. \
                  Use tensor.cpu() to convert to CPU tensor for other encoding methods.",
                 encoding_method
             )));
--- a/testing/qdp/test_bindings.py
+++ b/testing/qdp/test_bindings.py
@@ -422,9 +422,34 @@ def test_encode_cuda_tensor_preserves_input(data_shape, is_batch):
 
 @requires_qdp
 @pytest.mark.gpu
-@pytest.mark.parametrize("encoding_method", ["iqp"])
-def test_encode_cuda_tensor_unsupported_encoding(encoding_method):
-    """Test error when using CUDA tensor with an encoding not supported on GPU (only amplitude, angle, basis)."""
+@pytest.mark.parametrize(
+    "encoding_method,data",
+    [
+        ("iqp-z", [0.1, -0.2]),
+        ("iqp", [0.1, -0.2, 0.3]),
+    ],
+)
+def test_encode_cuda_tensor_iqp_methods(encoding_method, data):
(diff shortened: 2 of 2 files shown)


qdp/qdp-python/src/pytorch.rs
```diff
@@ -182,9 +182,18 @@ pub fn validate_cuda_tensor_for_encoding(
...
         }
+        "iqp" | "iqp-z" => {
+            if !dtype_str_lower.contains("float64") {
+                return Err(PyRuntimeError::new_err(format!(
+                    "CUDA tensor must have dtype float64 for {} encoding, got {}. \
+                     Use tensor.to(torch.float64)",
+                    method, dtype_str
+                )));
+            }
+        }
```
viiccwen, line 193: It's duplicated with angle method block, could be merged together.

testing/qdp/test_bindings.py
viiccwen, line 1: It's happy path, I'll take a follow-up PR to cover edge case.

viiccwen (commented):
overall lg, leave some comments. : )
Edit: wait, the testing even not passed, pls MAKE SURE ur patches all correct.

ryankert01:
not sure if he's on a gpu machine yet (:

rich7420:
locally passed tests.
A little thought:
but I think the unknown-encoding error now lists 'amplitude', 'angle', 'basis', 'iqp', or 'iqp-z' as the only CUDA-supported methods; if future encodings are added, this message in validate_cuda_tensor_for_encoding will need manual updates again—maybe we consider centralizing the supported method list to avoid drift.

SuyashParmar:
@ryankert01 @viiccwen @rich7420 Thanks everyone for the review and validation. CI is green, local validation was confirmed, and the CUDA iqp/iqp-z validation inconsistency is resolved.

qdp/qdp-python/src/pytorch.rs
SuyashParmar: yeah , agreed it can be merged with the angle path. Since this PR is already green and scoped to the behavior fix, I’m okay to keep this minimal and do that small refactor in a follow-up (unless you prefer I include it here). Also sounds good on edge-case coverage follow-up, thank you.

qdp/qdp-python/src/pytorch.rs
```diff
@@ -182,9 +182,18 @@ pub fn validate_cuda_tensor_for_encoding(
...
         _ => {
             return Err(PyRuntimeError::new_err(format!(
-                "CUDA tensor encoding currently only supports 'amplitude', 'angle', or 'basis' methods, got '{}'. \
+                "CUDA tensor encoding currently only supports 'amplitude', 'angle', 'basis', 'iqp', or 'iqp-z' methods, got '{}'. \
```
ryankert01, line 196: +1 with rich. should be dynamic.

ryankert01 (approved):
lgtm

qdp/qdp-python/src/pytorch.rs
ryankert01: opened an issue for it. cc @SuyashParmar

SuyashParmar: @ryankert01 ok working on it

qdp/qdp-python/src/pytorch.rs
viiccwen: IMO, it should be refactor in this PR.

viiccwen:
> @ryankert01 @viiccwen @rich7420 Thanks everyone for the review and validation. CI is green, local validation was confirmed, and the CUDA iqp/iqp-z validation inconsistency is resolved.

did u run CI in ur local machine? It's red.
Edit: I found u didn't implement GPU-pointer path, so CUDA tensor go into `NotImplemented ` from `encode_from_gpu_ptr `, I'll open a feature PR to fix it.

Issue #1096, opened 2026-02-27
Make unknown-encoding error dynamically list existing methods


ryankert01:
locally passed tests.
A little thought:
but I think the unknown-encoding error now lists 'amplitude', 'angle', 'basis', 'iqp', or 'iqp-z' as the only CUDA-supported methods; if future encodings are added, this message in validate_cuda_tensor_for_encoding will need manual updates again—maybe we consider centralizing the supported method list to avoid drift.

_Originally posted by @rich7420 in https://github.com/apache/mahout/issues/1093#issuecomment-3973029007_

alisha-1000:
Thanks for raising this — I agree that hardcoding the CUDA-supported encodings in the error message can easily lead to drift as new encodings are introduced.

I’d be happy to take this up. My plan is to centralize the CUDA-supported encoding list (e.g., via the encoding registry or a supports_cuda flag) and have validate_cuda_tensor_for_encoding derive the message dynamically from that source. This should keep the error messaging fully aligned with the implementation and avoid future maintenance overhead.

guan404ming:
Closed by #1100

PR #1100, opened 2026-02-27
Fix/dynamic cuda encoding method list

SuyashParmar:
Summary
This PR addresses [#1096] by making the CUDA unknown-encoding error message in QDP dynamic and maintainable, instead of hard-coding method names in [pytorch.rs].

This is a follow-up refactor to the CUDA iqp/iqp-z support work from [#1093].

Problem
Previously, the CUDA validation path returned an error string with a manually written list of supported methods.
When new encodings are added, this message can drift from actual support unless manually updated in multiple places.

What changed - 
1) Centralized CUDA method list and formatter
Added [constants.rs]:
CUDA_ENCODING_METHODS
format_supported_cuda_encoding_methods()
2) Updated CUDA validation error path
In [pytorch.rs], unknown-method errors now use:
format_supported_cuda_encoding_methods()
This keeps the message synchronized with the centralized method list.
3) Small cleanup included
Merged duplicated float64 dtype validation logic for:
angle
iqp
iqp-z
No behavior change intended, only reduced duplication.

Behavior impact -
No API change in method signatures.
No functional change to which methods are supported.
Only improves maintainability and reduces risk of stale/incorrect error text.

Why this is better - 
Single source of truth for CUDA-supported encoding names.
Lower maintenance cost when adding new encodings.
Clearer, less brittle validation code.

### DIFF
--- a/qdp/qdp-python/src/pytorch.rs
+++ b/qdp/qdp-python/src/pytorch.rs
@@ -182,9 +182,18 @@ pub fn validate_cuda_tensor_for_encoding(
                 )));
             }
         }
+        "iqp" | "iqp-z" => {
+            if !dtype_str_lower.contains("float64") {
+                return Err(PyRuntimeError::new_err(format!(
+                    "CUDA tensor must have dtype float64 for {} encoding, got {}. \
+                     Use tensor.to(torch.float64)",
+                    method, dtype_str
+                )));
+            }
+        }
         _ => {
             return Err(PyRuntimeError::new_err(format!(
-                "CUDA tensor encoding currently only supports 'amplitude', 'angle', or 'basis' methods, got '{}'. \
+                "CUDA tensor encoding currently only supports 'amplitude', 'angle', 'basis', 'iqp', or 'iqp-z' methods, got '{}'. \
                  Use tensor.cpu() to convert to CPU tensor for other encoding methods.",
                 encoding_method
             )));
--- a/qdp/qdp-python/src/pytorch.rs
+++ b/qdp/qdp-python/src/pytorch.rs
@@ -56,6 +56,20 @@ pub fn is_cuda_tensor(tensor: &Bound<'_, PyAny>) -> PyResult<bool> {
     Ok(device_type == "cuda")
 }
 
+const CUDA_ENCODING_METHODS: &[&str] = &["amplitude", "angle", "basis", "iqp", "iqp-z"];
+
+fn format_supported_cuda_encoding_methods() -> String {
+    let quoted: Vec<String> = CUDA_ENCODING_METHODS
+        .iter()
+        .map(|method| format!("'{}'", method))
+        .collect();
+    let len = quoted.len();
+    if len == 1 {
+        return quoted[0].clone();
+    }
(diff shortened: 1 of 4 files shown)


SuyashParmar:
@ryankert01 can you pls check the updates ?

qdp/qdp-python/src/pytorch.rs
```diff
@@ -56,6 +56,20 @@ pub fn is_cuda_tensor(tensor: &Bound<'_, PyAny>) -> PyResult<bool> {
...
+        .iter()
+        .map(|method| format!("'{}'", method))
+        .collect();
+    let len = quoted.len();
+    if len == 1 {
+        return quoted[0].clone();
+    }
+    format!("{}, or {}", quoted[..len - 1].join(", "), quoted[len - 1])
+}
+
```
viiccwen: The intent of this helper function can be more clean, manual length handling and clone is unnecessary.

viiccwen (commented):
IMO, it can be resolved in #1093, also left comment.

qdp/qdp-python/src/pytorch.rs
SuyashParmar: @viiccwen updated this helper to a cleaner implementation (no manual length indexing or clone)

qdp/qdp-python/src/pytorch.rs
```diff
@@ -56,6 +56,23 @@ pub fn is_cuda_tensor(tensor: &Bound<'_, PyAny>) -> PyResult<bool> {
     Ok(device_type == "cuda")
 }
 
+const CUDA_ENCODING_METHODS: &[&str] = &["amplitude", "angle", "basis", "iqp", "iqp-z"];
```
ryankert01: I think we have a better place to store consts?

ryankert01 (approved):
overall lg, left comments

qdp/qdp-python/src/pytorch.rs
SuyashParmar: @ryankert01 Makes sense — for this follow-up I kept the const local to keep scope small, but I agree it should live in a shared constants/module location.

I can move it in a quick follow-up refactor if preferred.

ryankert01: I think you can fix it in this PR.

ryankert01 (changes requested):
please help resolve conflicts

400Ping:
Can you add a clearer PR description? It is best to leave good records so we can go back and look into it.

SuyashParmar:
@ryankert01 Conflict resolved

qdp/qdp-python/src/pytorch.rs
SuyashParmar: @ryankert01 done

SuyashParmar:
@400Ping updated the description

SuyashParmar:
@ryankert01 updated this PR to use CUDA_ENCODING_METHODS.contains(...) as the single supported/unsupported gate, so the error message list and validation path can’t drift.

ryankert01 (approved):

PR #1101, opened 2026-02-27
Add unified and configurable null handling

### LINKED ISSUE #765: [QDP] Unify Null Value Handling Strategy between Batch and Streaming Modes
### **Description**

There is currently an inconsistency in how our system handles `null` values depending on the execution mode. To ensure predictable behavior and data integrity, we should align these strategies.

### **Current Behavior**

* **Batch Mode:** Null values are automatically coerced/set to `0`.
* **Streaming Mode:** Null values trigger a runtime error.

### **Proposed Change**

We need to decide on a unified strategy for handling `null` values across both modes. Potential directions include:

1. **Strict Error Handling:** Both modes return an error to prevent silent data corruption.
2. **Default Coercion:** Both modes default to `0` (or a type-appropriate default).
3. **Configurable Policy:** Allow users to define the `on_null` behavior (e.g., `ignore`, `fill_zero`, or `raise_error`).

### **Context**

https://github.com/apache/mahout/pull/753#discussion_r2650061909

### **To-Do / Discussion Points**

* [ ] Research performance implications of adding null-checks in the streaming hot path.
* [ ] Determine if "Default to 0" is safe for all supported data types.
* [ ] Reach a consensus on the unified default behavior.

ryankert01:
### Related Issues

Closes #765 

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### How

- Unify null value handling between batch and streaming Parquet/Arrow readers by introducing a configurable `NullHandling` enum (`FillZero` | `Reject`)
- Previously batch mode silently coerced nulls to `0.0` while streaming mode threw a runtime error — both paths now use the same `handle_float64_nulls()` helper
- Defaults to `FillZero` for backward compatibility; `Reject` returns a clear error with guidance
- Threads the policy through `PipelineConfig`, PyO3 bindings, and the Python `QuantumDataLoader` builder (`.null_handling("fill_zero" | "reject")`)

## Checklist

### DIFF
--- a/qdp/qdp-core/src/encoding/mod.rs
+++ b/qdp/qdp-core/src/encoding/mod.rs
@@ -141,7 +141,8 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
     encoder: E,
 ) -> Result<*mut DLManagedTensor> {
     // Initialize reader
-    let mut reader_core = crate::io::ParquetBlockReader::new(path, None)?;
+    let mut reader_core =
+        crate::io::ParquetBlockReader::new(path, None, crate::reader::NullHandling::FillZero)?;
     let num_samples = reader_core.total_rows;
 
     // Allocate output state vector
--- a/qdp/qdp-core/src/io.rs
+++ b/qdp/qdp-core/src/io.rs
@@ -26,7 +26,7 @@ use std::fs::File;
 use std::path::Path;
 use std::sync::Arc;
 
-use arrow::array::{Array, ArrayRef, Float64Array, RecordBatch};
+use arrow::array::{ArrayRef, Float64Array, RecordBatch};
 use arrow::datatypes::{DataType, Field, Schema};
 use parquet::arrow::ArrowWriter;
 use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
--- a/qdp/qdp-core/src/io.rs
+++ b/qdp/qdp-core/src/io.rs
@@ -33,30 +33,28 @@ use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
 use parquet::file::properties::WriterProperties;
 
 use crate::error::{MahoutError, Result};
+use crate::reader::{NullHandling, handle_float64_nulls};
 
 /// Converts an Arrow Float64Array to Vec<f64>.
-pub fn arrow_to_vec(array: &Float64Array) -> Vec<f64> {
-    if array.null_count() == 0 {
-        array.values().to_vec()
-    } else {
-        array.iter().map(|opt| opt.unwrap_or(0.0)).collect()
-    }
+pub fn arrow_to_vec(array: &Float64Array, null_handling: NullHandling) -> Result<Vec<f64>> {
+    let mut result = Vec::with_capacity(array.len());
+    handle_float64_nulls(&mut result, array, null_handling)?;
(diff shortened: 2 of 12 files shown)


qdp/qdp-core/src/reader.rs
```diff
@@ -45,8 +45,48 @@
 //! }
 //! ```
 
+use arrow::array::{Array, Float64Array};
```
viiccwen, line 48: `Array` is not used in this file.

qdp/qdp-core/src/io.rs
```diff
@@ -33,38 +33,36 @@ use parquet::arrow::arrow_reader::ParquetRecordBatchReaderBuilder;
...
     }
 
-    result
+    Ok(result)
 }
 
 /// Reads Float64 data from a Parquet file.
 ///
 /// Expects a single Float64 column. For zero-copy access, use [`read_parquet_to_arrow`].
 pub fn read_parquet<P: AsRef<Path>>(path: P) -> Result<Vec<f64>> {
```
viiccwen, line 63: currently we specify `read_*` ( `read_parquet(...)` ) to use `NullHandling::FillZero`, should we adopt same strategy like `*Readers` (`ParquetReader`)?

like this?
```rust
pub fn read_parquet<P: AsRef<Path>>(path: P) -> Result<Vec<f64>>;
// to
pub fn read_parquet<P: AsRef<Path>>(
    path: P,
    null_handling: Option<NullHandling>,
) -> Result<Vec<f64>>;
```

viiccwen (commented):
overall lg, but need to update docs, and left some comments.

qdp/qdp-core/src/reader.rs
ryankert01: nice catch

qdp/qdp-core/src/io.rs
ryankert01: Looks like we can use `ParquetReader::new(path, None, NullHandling::Reject)` when we want to configure that. We see `read_parquet` as a convenient function? (not sure about that, but we can left it to the future)

qdp/qdp-core/src/encoding/mod.rs
```diff
@@ -141,7 +141,8 @@ pub(crate) fn stream_encode<E: ChunkEncoder>(
     encoder: E,
 ) -> Result<*mut DLManagedTensor> {
     // Initialize reader
-    let mut reader_core = crate::io::ParquetBlockReader::new(path, None)?;
+    let mut reader_core =
+        crate::io::ParquetBlockReader::new(path, None, crate::reader::NullHandling::FillZero)?;
```
rich7420, line 145: It seems stream_encode() uses hardcoded NullHandling::FillZero for ParquetBlockReader. Batch and streaming reader paths are unified; this call site has no config. I think maybe we could have a follow-up here if stream_encode is ever given a config or options, expose null_handling there. otherwise document that streaming encode keeps FillZero for backward compatibility.

rich7420 (approved):
overall LGTM!

qdp/qdp-core/src/encoding/mod.rs
ryankert01: yes

PR #1104, opened 2026-02-28
Fix incorrect path of workspace in devcontainer setup.sh

### LINKED ISSUE #1103: [Bug] DevContainer setup fails due to incorrect workspace path in setup.sh
### What

When building the DevContainer for the first time, the `postCreateCommand` fails with the following error:
`/workspace/.devcontainer/setup.sh: line 24: cd: /workspaces/mahout: No such file or directory`

### Environment

- OS: Ubuntu 22.04.4 LTS
- Python version: 3.10.19
- Qumat version:

Error Screenshot: 
Image

Howardisme:
### Related Issues

Fix #1103 

### Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

This PR fixes an issue where the DevContainer fails to initialize properly during a fresh build.

### How

Modified `.devcontainer/setup.sh` line 24: `cd /workspaces/mahout` -> `cd /workspace`

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: test_rust test_python tests pre-commit setup-test-python
+.PHONY: test_rust test_python tests pre-commit setup-test-python install-llvm-cov
 
 # Detect NVIDIA GPU
 HAS_NVIDIA := $(shell command -v nvidia-smi >/dev/null 2>&1 && nvidia-smi -L >/dev/null 2>&1 && echo yes || echo no)
@@ -22,7 +22,10 @@ HAS_NVIDIA := $(shell command -v nvidia-smi >/dev/null 2>&1 && nvidia-smi -L >/d
 setup-test-python:
 	uv sync --group dev
 
-test_rust:
+install-llvm-cov:
+	@cargo llvm-cov --version >/dev/null 2>&1 || (echo "[INFO] Installing cargo-llvm-cov..." && cargo install cargo-llvm-cov)
+
+test_rust: install-llvm-cov
 ifeq ($(HAS_NVIDIA),yes)
 	cd qdp && cargo llvm-cov test --workspace --exclude qdp-python --html --output-dir target/llvm-cov/html
 	cd qdp && cargo llvm-cov report --summary-only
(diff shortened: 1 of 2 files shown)


.devcontainer/setup.sh
```diff
@@ -21,6 +21,6 @@ curl -LsSf https://astral.sh/uv/install.sh | sh
 source $HOME/.local/bin/env
 
 # setup pre-commit hooks
-cd /workspaces/mahout
+cd /workspace
```
ryankert01, line 24: Could you help me understand? I think it should be in mahout directory?

viiccwen: +1

Howardisme: Hi @ryankert01  and @viiccwen , Thank you for the quick response. I completely understand why it feels like it should be in mahout directly. However, I found devcontainer.json cause this error happened. In devcontainer.json, the workspace is explicitly mounted to `/workspace`: 

image

Because devcontainer.json forces the target to /workspace, when setup.sh tries to run cd /workspaces/mahout on line 24, it throws the No such file or directory error and break the initialization of setup.

We have two ways to fix it:
1. (My current PR): Update setup.sh to use cd /workspace to match the devcontainer.json. This is often safer because we don't know what the user names their local clone directory.
2. Update devcontainer.json to mount to /workspaces/mahout instead.

Let me know which way you think is better to this issue or any follow up questions!

viiccwen: Pls update ur own setting file.

viiccwen (commented):
Thx for contributing, but there is no need to change.

viiccwen:
If u still have any issue in building the env, pls let us know. : )

.devcontainer/setup.sh
Howardisme: Got it! I wil update my local file to align with the script.

Howardisme: Closing this PR. Looking forward to contributing to other parts of this project!

viiccwen: nice one!

PR #1136, opened 2026-03-06
[QDP] Optimize run_throughput_pipeline to avoid per-iteration Vec allocations

### LINKED ISSUE #1125: [QDP] Optimize run_throughput_pipeline to avoid per-iteration Vec allocations
### What
`run_throughput_pipeline` currently allocates a new `Vec<f64>` for every batch via generate_batch(...) inside the hot loops (warmup + timed). In throughput benchmarks this adds significant allocator overhead and variance, making results reflect CPU allocation time rather than GPU encode_batch throughput. We should allocate a single batch buffer once and refill it in-place each iteration.

### Location
[qdp/qdp-core/src/pipeline_runner.rs](https://github.com/apache/mahout/blob/main/qdp/qdp-core/src/pipeline_runner.rs)
Function: `run_throughput_pipeline` 

### Proposed Change
- Allocate batch_buf: Vec<f64> once outside the loops.
- Replace per-iteration generate_batch(...) with an in-place fill function (e.g. fill_batch_inplace(...)) that writes into &mut [f64].
- Reuse the same batch_buf in both warmup and timed loops.

0lai0:
### Related Issues

Close [#1125](https://github.com/apache/mahout/issues/1125) 

### Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

In `run_throughput_pipeline`, each iteration of the warmup and timed loops called `generate_batch(...)`, which allocates a new `Vec<f64>` (`batch_size * vector_len` elements). For high-frequency throughput benchmarks, this creates significant memory allocation pressure and adds allocator overhead to the timing results, making them reflect CPU allocation time rather than pure GPU `encode_batch` throughput.

### How

- Pre-allocate a single batch buffer before entering the loops
- Reuse the buffer in each iteration via a new `fill_batch_inplace(...)` function that writes data in-place
- Refactor `generate_batch` to delegate to `fill_batch_inplace`, eliminating code duplication

## Checklist

### DIFF
--- a/qdp/qdp-core/src/pipeline_runner.rs
+++ b/qdp/qdp-core/src/pipeline_runner.rs
@@ -589,11 +589,13 @@ mod tests {
     use super::*;
 
     fn assert_generate_and_inplace_match(encoding_method: &str) {
-        let mut config = PipelineConfig::default();
-        config.num_qubits = 5;
-        config.batch_size = 8;
-        config.encoding_method = encoding_method.to_string();
-        config.seed = Some(123);
+        let config = PipelineConfig {
+            num_qubits: 5,
+            batch_size: 8,
+            encoding_method: encoding_method.to_string(),
+            seed: Some(123),
+            ..Default::default()
+        };
 
         let vector_len = vector_len(config.num_qubits, &config.encoding_method);
         let batch_idx = 7;
--- a/qdp/qdp-core/src/pipeline_runner.rs
+++ b/qdp/qdp-core/src/pipeline_runner.rs
@@ -488,20 +488,31 @@ fn fill_sample(seed: u64, out: &mut [f64], encoding_method: &str) -> Result<()> 
 
 /// Generate one batch (batch_size * vector_len elements, or batch_size * 1 for basis).
 fn generate_batch(config: &PipelineConfig, batch_idx: usize, vector_len: usize) -> Vec<f64> {
+    let mut batch = vec![0.0f64; config.batch_size * vector_len];
+    fill_batch_inplace(config, batch_idx, vector_len, &mut batch);
+    batch
+}
+
+/// Fill an existing batch buffer in-place (avoids per-iteration allocations in benchmarks).
+fn fill_batch_inplace(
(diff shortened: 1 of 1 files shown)


qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -569,3 +583,42 @@ pub fn run_throughput_pipeline(config: &PipelineConfig) -> Result<PipelineRunRes
...
+        let config = PipelineConfig {
+            num_qubits: 5,
+            batch_size: 8,
+            encoding_method: encoding_method.to_string(),
+            seed: Some(123),
+            ..Default::default()
+        };
+
+        let vector_len = vector_len(config.num_qubits, &config.encoding_method);
+        let batch_idx = 7;
```
ryankert01: we add edge cases like: 0, batch_size-1

ryankert01: We can also test if adjacent batch is same of not.
```rust
let batch0 = generate_batch(&config, 0, vector_len);
let batch1 = generate_batch(&config, 1, vector_len);
assert_ne!(batch0, batch1);
```

ryankert01 (commented):
Implementation looks good, a nit with the tests

qdp/qdp-core/src/pipeline_runner.rs
0lai0: Thanks @ryankert01 for review. I'll update in next commit.

viiccwen (commented):
lg!
but we can add more tests like `Seed=None`, `batch_size=1`, `amplitude==AMPLITUDE`, `amplitude sample in [0, 1]`

ryankert01 (approved):

rich7420 (approved):

rich7420:
@0lai0 thanks for the patch, @ryankert01 and @viiccwen thanks for the review

PR #1154, opened 2026-03-08
[QDP] Enhance benchmark's installation journey

### LINKED ISSUE #1141: Enhace benchmark's installation journey
### Why

Now, I don't see `make benchmark` working. Btw, because we unifies the python environment right now, I think we should now `make benchmark` on the root directory? (not sure)

### How

After, we should be able to...
1. setup benchmark env on root python env
2. be able to successfully run benchmarks
3. Update benchmark readme
4. (optional, can be seperate pr) I want to run benchmark on the root, and probably modulize it to `python -m benchmark.numpy_io` etc.

Root refers to mahout project root (mahout/)

0lai0:
### Related Issues

Closes #1141 

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [x] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

Currently, `make benchmark` doesn't exist, and setting up benchmarks requires multiple manual steps. Since the Python environment has been unified to use the root `.venv`, we should provide a simple `make benchmark` command from the root directory to streamline the benchmark setup and execution process.
### How

- Added `benchmark` and `setup-benchmark` targets to the root `Makefile`
  - `setup-benchmark`: Installs benchmark dependencies and builds QDP extension
  - `benchmark`: Sets up environment and displays instructions for running benchmarks
- Updated `qdp/qdp-python/benchmark/README.md` to document `make benchmark` as the recommended way
- Updated `qdp/DEVELOPMENT.md` to include `make benchmark` in the benchmarks section

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,6 +48,7 @@ pre-commit: setup-test-python
 setup-benchmark: setup-test-python
 ifeq ($(HAS_NVIDIA),yes)
 	@echo "[INFO] Setting up benchmark environment..."
+	uv sync --group dev --extra qdp
 	uv sync --project qdp/qdp-python --group benchmark --active
 	unset CONDA_PREFIX && uv run --active maturin develop --manifest-path qdp/qdp-python/Cargo.toml
 else
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-.PHONY: test_rust test_python tests pre-commit setup-test-python install-llvm-cov benchmark setup-benchmark
+.PHONY: test_rust test_python tests pre-commit setup-test-python install-llvm-cov setup-benchmark
 
 # Detect NVIDIA GPU
 HAS_NVIDIA := $(shell command -v nvidia-smi >/dev/null 2>&1 && nvidia-smi -L >/dev/null 2>&1 && echo yes || echo no)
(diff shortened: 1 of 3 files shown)


qdp/qdp-python/benchmark/README.md
```diff
@@ -11,12 +11,20 @@ scripts:
...
+
+This will:
+1. Set up the benchmark environment in the unified root venv (`mahout/.venv`)
+2. Build the QDP extension (if GPU available)
+3. Display instructions for running specific benchmarks
+
+Then run individual benchmarks:
 
 ```bash
-uv sync --group dev --extra qdp
```
guan404ming, line 17: Why do we need to remove these lines?

guan404ming (commented):
Thanks

qdp/qdp-python/benchmark/README.md
0lai0: Good catch! Thanks @guan404ming.
I found that `setup-benchmark` should include `uv sync --group dev --extra qdp` to ensure the `qumat-qdp` package path is properly registered in the environment, consistent with the development environment setup in DEVELOPMENT.md.
I will fix the Makefile to include this, and also restore the manual setup steps in the README as an alternative option. Will push a fix shortly!

0lai0: I moved this line because it will be handled in `make benchmark`.

Makefile
```diff
@@ -39,8 +39,39 @@ ifeq ($(HAS_NVIDIA),yes)
...
+	uv sync --group dev --extra qdp
+	uv sync --project qdp/qdp-python --group benchmark --active
+	unset CONDA_PREFIX && uv run --active maturin develop --manifest-path qdp/qdp-python/Cargo.toml
+else
+	@echo "[SKIP] No NVIDIA GPU detected, skipping maturin develop"
+	@echo "[INFO] Setting up benchmark environment (CPU-only)..."
+	uv sync --project qdp/qdp-python --group benchmark --active
+endif
+
+benchmark: setup-benchmark
```
guan404ming: Should we rename to setup-benchmark or benchmark-info since it doesn't run any benchmarks?

guan404ming (approved):
Left one nit

viiccwen (approved):
overall lg!

Makefile
0lai0: I think setup-benchmark would be better.

ryankert01 (approved):

guan404ming:
Thanks for the update.

PR #1242, opened 2026-04-06
perf(qdp): Implement async prefetching and native f32 dispatch pipelines

rich7420:
### Changes

- [x] Bug fix
- [x] New feature
- [x] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

The legacy Quantum Data Pipeline (QDP) suffered from critical structural bottlenecks and technical debt that severely compromised maximum throughput:
- **GPU Starvation (Synchronous Blocking):** The DataLoader design relied on unoptimized, serial execution. Because CPU data generation and file I/O operations blocked the GPU encoding logic, expensive GPU computational cycles were left idle waiting for buffer preparations.
- **Precision Conversion Tax:** Downstream processes forcefully allocated and shuttled CPU data via `Vec<f64>`, even when the neural network explicitly required `f32` structures. This strictly doubled the required memory bandwidth, exacerbated host heap allocations, and mandated costly upcasting/downcasting conversions within CUDA contexts.
- **Nightly Toolchain Dependency:** Core implementation modules internally relied on experimental, nightly-only Rust semantic features (e.g., `is_multiple_of` numeric casting and incomplete `if let` chain scoping), breaking standard developer verification tools and preventing conventional `cargo test`/`cargo check` routines on a standard toolchain.

### How

- **Asynchronous Prefetch & Buffer Recycling Architecture:** Substituted the serial, monolithic reader structure with a generic, decoupled `BatchProducer` format (`Synthetic` / `InMemory` / `Streaming`). Background parallelization was achieved utilizing a bounded `std::sync::mpsc::sync_channel(16)`. Furthermore, we implemented a **zero-allocation buffer recycling loop**, which returns exhausted arrays back to the producer, completely eliminating dynamic CPU allocation overhead during steady-state execution.
- **Native `f32` Pipelines & Safety Gating:** Expanded both the `QuantumEncoder` and `QdpEngine` endpoints to actively dispatch `encode_batch_f32` APIs. Python consumers can now leverage an explicit `float32_pipeline=True` flag to drive input data straight to the optimized `f32` kernels, accelerating kernel launch duration by approximately ~300% (RTX 2080ti).
  - *Stability Gating:* We implemented an auto-gating mechanism (`PipelineConfig::normalize()`). If an encoding scheme does not yet support native `f32` dispatch (e.g., `angle` or `basis`), the pipeline safely logs and gracefully downgrades the pipeline to `f64`, entirely preventing GPU kernel crashes or unsupported `NotImplemented` panics.
  - *Accurate Benchmarking:* Repaired the pipeline's warmup phase to correctly honor the requested pipeline precision. The framework now successfully triggers the JIT compilation for `f32` kernels before the timers start, ensuring perfectly accurate benchmarks.
- **Rust Stability Teardown:** Scrubbed all non-compliant unstable conditionals and experimental primitives across `qdp-core` (`pipeline_runner.rs`, `dlpack.rs`, `tensorflow.rs`, `parquet.rs`), substituting them with stable logic equivalents to secure a 100% resilient `cargo build/test` developer experience. Added 4 new coverage tests explicitly for the `f32` pipeline fallback routines.

### test 
```bash
uv run --group benchmark python benchmark/benchmark_throughput.py --qubits 16 --batches 200 --batch-size 64 --frameworks mahout
```

## Checklist

### DIFF
--- a/qdp/qdp-core/tests/common/mod.rs
+++ b/qdp/qdp-core/tests/common/mod.rs
@@ -45,6 +45,7 @@ pub fn create_test_data_f32(size: usize) -> Vec<f32> {
 /// Writes a FixedSizeList<Float64, sample_size> Parquet file for streaming encoder tests.
 /// Each `sample_size` consecutive values in `data` form one row.
 #[allow(dead_code)]
+#[allow(clippy::manual_is_multiple_of)]
 pub fn write_fixed_size_list_parquet(path: &str, data: &[f64], sample_size: usize) {
     assert!(sample_size > 0, "sample_size must be > 0");
     assert!(
--- a/qdp/qdp-core/tests/common/mod.rs
+++ b/qdp/qdp-core/tests/common/mod.rs
@@ -48,8 +48,8 @@ pub fn create_test_data_f32(size: usize) -> Vec<f32> {
 pub fn write_fixed_size_list_parquet(path: &str, data: &[f64], sample_size: usize) {
     assert!(sample_size > 0, "sample_size must be > 0");
     assert!(
-        data.len().is_multiple_of(sample_size),
-        "data.len() ({}) must be a multiple of sample_size ({})",
+        data.len() % sample_size == 0,
+        "Data length ({}) must be a multiple of sample size ({})",
         data.len(),
         sample_size
     );
--- a/qdp/qdp-core/src/dlpack.rs
+++ b/qdp/qdp-core/src/dlpack.rs
@@ -252,6 +252,7 @@ impl GpuStateVector {
     /// # Safety
     /// Freed by DLPack deleter when PyTorch releases tensor.
     /// Do not free manually.
+    #[allow(clippy::manual_is_multiple_of)]
     pub fn to_dlpack(&self) -> *mut DLManagedTensor {
         // Always return 2D tensor: Batch [num_samples, state_len], Single [1, state_len]
         let (shape, strides) = if let Some(num_samples) = self.num_samples {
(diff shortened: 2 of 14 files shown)


qdp/qdp-core/tests/common/mod.rs
```diff
@@ -45,11 +45,12 @@ pub fn create_test_data_f32(size: usize) -> Vec<f32> {
...
 pub fn write_fixed_size_list_parquet(path: &str, data: &[f64], sample_size: usize) {
     assert!(sample_size > 0, "sample_size must be > 0");
     assert!(
-        data.len().is_multiple_of(sample_size),
-        "data.len() ({}) must be a multiple of sample_size ({})",
+        data.len() % sample_size == 0,
+        "Data length ({}) must be a multiple of sample size ({})",
         data.len(),
         sample_size
     );
```
viiccwen, line 55: why made this change? pre-commit?

viiccwen (commented):
Performance optimization looks nice! Some nits:
1. This PR enables `float32_pipeline` by default in the loader/pipeline entry points, but only amplitute imple `encode_batch_32()`. Other encodings still fallback to `NotImplemented` path, I think we should gate and only enable supporting method rn, or fallback to the existing f64 path.
2. Add some testing to cover changed.

qdp/qdp-core/tests/common/mod.rs
rich7420: This was a Clippy compatibility fix — usize::is_multiple_of() requires the unsigned_is_multiple_of feature (stabilized in Rust 1.87+), and CI's Clippy flags % == 0 as clippy::manual_is_multiple_of. We replaced the nightly API with % and added #[allow(clippy::manual_is_multiple_of)] where needed to pass both stable and nightly CI. The error message was also made slightly more readable.

ryankert01 (approved):
lgtm 

Image

rich7420:
btw that's my latest results
image

viiccwen (approved):
Looks good!

PR #1264, opened 2026-04-13
feat: Add float32 zero-copy angle encoding in Python bindings (single sample)

### LINKED ISSUE #1250: [Feature] Add float32 zero-copy angle encoding in Python bindings (single sample)
Reference: #1004 

### What

This issue depends on the core/kernel work from #1248. It should expose the new single-sample angle `f32` path to Python users without introducing batch support yet.

viiccwen:
### Related Issues

Closes #1250

### Changes

- [x] Bug fix
- [x] New feature
- [ ] Refactoring
- [x] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

QDP core already supports single-sample angle encoding from CUDA `float32`
buffers, but the Python binding did not expose that path. This PR adds the
Python-side support while keeping batch CUDA angle inputs on `float64`.

### How

- route 1D CUDA `float32` angle tensors through the dedicated zero-copy angle
  GPU-pointer path
- keep CUDA angle `float32` support limited to 1D single-sample tensors
- add targeted Python tests for success and rejection paths
- document the CUDA tensor dtype rules for `amplitude` and `angle`
- pin the IQP naive-reference comparison test to `float64` precision so the
  comparison matches the reference dtype

## Checklist

### DIFF
--- a/qdp/qdp-python/src/engine.rs
+++ b/qdp/qdp-python/src/engine.rs
@@ -477,7 +477,7 @@ impl QdpEngine {
 
     /// Encode directly from a PyTorch CUDA tensor. Internal helper.
     ///
-    /// Dispatches to the core f32 GPU pointer API for float32 amplitude encoding,
+    /// Dispatches to the core f32 GPU pointer APIs for supported float32 CUDA paths,
     /// or to the float64/basis GPU pointer APIs for other dtypes and methods.
     fn _encode_from_cuda_tensor(
         &self,
@@ -495,7 +495,7 @@ impl QdpEngine {
         let ndim: usize = data.call_method0("dim")?.extract()?;
         let tensor_info = extract_cuda_tensor_info(data)?;
 
-        if method.as_str() == "amplitude" && is_f32 {
+        if is_f32 && matches!(method.as_str(), "amplitude" | "angle") {
             match ndim {
                 1 => {
                     let input_len: usize = data.call_method0("numel")?.extract()?;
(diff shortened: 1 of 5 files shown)


qdp/qdp-python/src/engine.rs
```diff
@@ -522,6 +541,12 @@ impl QdpEngine {
                     })
                 }
                 2 => {
+                    if method.as_str() == "angle" {
+                        return Err(PyRuntimeError::new_err(
+                            "CUDA tensor float32 angle encoding currently supports only 1D single-sample tensors. \
+                             Use tensor.to(torch.float64) for batch angle encoding.",
+                        ));
+                    }
```
rich7420: validate_cuda_tensor_for_encoding in pytorch.rs already rejects 2D F32 angle tensors before                    
  _encode_from_cuda_tensor is ever called. So the check here , it seems unreachable                                                                                                        
plz correct me if i'm wrong.

qdp/qdp-python/src/engine.rs
```diff
@@ -504,16 +508,31 @@ impl QdpEngine {
...
+                                .encode_angle_from_gpu_ptr_f32_with_stream(
+                                    data_ptr, input_len, num_qubits, stream_ptr,
+                                )
+                                .map_err(|e| {
+                                    PyRuntimeError::new_err(format!(
+                                        "Encoding failed (float32 angle): {}",
+                                        e
+                                    ))
+                                })?,
+                            _ => unreachable!("matches!() should have limited f32 methods"),
```
rich7420: just a nits: The panic message references the internal matches!() macro. Something like "unreachable: unhandled f32 encodin method" is clearer.

rich7420 (commented):
thanks for the patch! sorry for the late review:

qdp/qdp-python/src/engine.rs
viiccwen: Sure, it'll get blocked in `validate_cuda_tensor_for_encoding(...)` in `qdp-python/src/pytorch.rs`.
IMO, just simply remove it.

qdp/qdp-python/src/engine.rs
viiccwen: Fine, I adjust it to `unreachable: unhandled f32 encoding method`, WDYT?

rich7420 (approved):
LGTM!

PR #1268, opened 2026-04-13
feat: Add batch float32 zero-copy angle encoding in qdp-core

viiccwen:
### Related Issues

- Closes #1267

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

---

- add batched float32 CUDA kernel support for angle encoding
- expose the new kernel through `qdp-kernels` FFI
- implement host-side angle batch float32 encoding in `qdp-core`
- add zero-copy batch float32 angle APIs for CUDA pointers, including stream-aware variants
- align float32 GPU-pointer validation with the existing float64 angle batch behavior
- add kernel and core tests for success paths and failure paths

### Why

QDP already had angle float32 zero-copy support for single samples, but not for batched angle features. That made batched angle workloads inconsistent with the existing single-sample API and with the amplitude float32 batch implementation pattern.

This change closes that gap in the Rust/CUDA core only. Python bindings remain out of scope for a follow-up PR.

### How

- implement `launch_angle_encode_batch_f32` in `qdp-kernels/src/angle.cu`
- add the corresponding extern declaration and non-CUDA stub in `qdp-kernels/src/lib.rs`
- extend `AngleEncoder` with host-side batch float32 encoding and batch GPU-pointer float32 encoding
- add public `QdpEngine` entry points for angle batch float32 zero-copy on the default stream and explicit streams
- validate sample shape and reject non-finite float32 GPU-resident batch inputs before launching the encoding kernel
- add tests covering kernel correctness, host-side batch encoding, zero-copy batch encoding, stream interop, error handling, and Float64-engine conversion

## Checklist

- [x] Added or updated unit tests for all changes
- [ ] Added or updated documentation for all changes

PR #1275, opened 2026-04-19
[Feature][QDP] F32 support for angle/basis encoders, fidelity metrics and pipeline improvements

rich7420:
### Changes

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [x] Documentation
- [x] Test
- [x] CI/CD pipeline
- [ ] Other

### Why

Angle and basis encoding always dispatched to F64 CUDA kernels even though F32
is sufficient for typical ML workloads. The hard-coded prefetch depth of 16
was also unsafe at high qubit counts (20-qubit amplitude encoding = 512 MB/batch
× 16 = 8 GB of buffered data). The `PipelineIterator` shutdown sequence also
had no ordered teardown, risking thread leaks on drop.

### How

**F32 batch kernels for angle and basis**
- Added `angle_encode_batch_kernel_f32` (grid-stride) and
  `launch_angle_encode_batch_f32` in `angle.cu`
- Added `basis_encode_kernel_f32`, `basis_encode_batch_kernel_f32`, and their
  launchers in `basis.cu`
- Extended `Encoding::supports_f32()` to return `true` for
  `Amplitude | Angle | Basis`

**F32 zero-copy CUDA tensor paths**
- New `encode_{,angle_,basis_}{,batch_}from_gpu_ptr_f32{,_with_stream}` methods
  on `QdpEngine` cover the six `(encoding, ndim)` combinations
  (`Amplitude | Angle | Basis` × `1D | 2D`)
- `_encode_from_cuda_tensor` (`qdp-python/src/engine.rs`) was refactored from
  six separately-wrapped branches into a single `match (encoding, ndim)` table,
  hoisting the duplicated `validate_shape` + `get_torch_cuda_stream_ptr` calls
  shared by both the f32 and f64 paths (-45 LoC, behavior unchanged, clippy clean)

**GPU input validation (new `gpu/validation.rs` + `kernels/validation.cu`)**
- Basis f32 zero-copy needs to reject inputs that violate the basis-index
  contract before they touch the state-vector write. Float32 inputs can be
  `NaN`/`Inf`, negative, non-integer, or `≥ 2^num_qubits` — any of which
  would silently corrupt the encoded state if not caught
- `check_basis_indices_kernel_{f32,f64}` uses an atomic error-flag bitmask
  (`NON_FINITE | NEGATIVE | NON_INTEGER | OUT_OF_RANGE`) so a single device
  pass catches all four failure modes
- `assert_all_finite_f32` is reused by the angle f32 path; integration test
  `test_angle_batch_f32_rejects_nan` / `_rejects_infinity` verifies behavior

### Benchmark results

**Original (different host)**, 16 qubits, batch_size = 64, 200 batches:

| Encoding  | main (F64)     | this PR (F32)   | Speedup |
|-----------|----------------|-----------------|---------|
| amplitude | 1,170 vec/s    | 1,184 vec/s     | ~1×     |
| angle     | 7,466 vec/s    | 29,790 vec/s    | **4×**  |
| basis     | 59,883 vec/s   | 314,031 vec/s   | **5×**  |

**Re-verified locally** post-rebase (RTX 2080 Ti, 16 qubits, batch_size = 64,
200 batches, prefetch = 16):

| Encoding  | this PR (F32) — sustained pipeline |
|-----------|------------------------------------|
| amplitude | 13,583 vec/s                       |
| angle     | **177,391 vec/s**                  |
| basis     | **622,483 vec/s**                  |

Amplitude tracks the pre-rebase baseline (≤2% delta on this hardware);
angle and basis show the expected 4–5× F32 speedup relative to F64.

**Auto prefetch depth**
- `compute_optimal_prefetch_depth(num_qubits, batch_size, encoding, dtype)`
  targets a 256 MB CPU buffer, clamped to `[1, 32]`; checked arithmetic
  guards overflow at very large qubit counts. Default `prefetch_depth` is
  now `0` (resolved in `normalize()` at iterator start)

**PipelineIterator cleanup**
- `rx` wrapped in `Mutex<Receiver>` (required for PyO3 `#[pyclass]` `Sync` bound)
- `recycle_tx` and `producer_handle` changed to `Option<_>` for owned teardown
- `Drop` follows correct shutdown order: drop sender → drain channel → join thread

**Fidelity metrics (`gpu/metrics.rs`)**
- CPU-side `|⟨ψ|φ⟩|²` fidelity and trace distance, including cross-precision
  (F32 vs F64) helpers
- Module + re-exports are `#[doc(hidden)]` and gated as test-only — not part
  of the supported runtime API. Integration tests `tests/gpu_fidelity.rs`
  verify ≥ 0.9999996 fidelity at 8 / 12 / 16 / 20 qubits

**Python / benchmark**
- `backend("auto")` falls back to PyTorch with a `RuntimeWarning`
- `as_torch_dataset()` wraps the loader as a `torch.utils.data.IterableDataset`
- `--warmup N` flag added to `benchmark_throughput.py` / `benchmark_latency.py`
  (default 5; same as the previously-hardcoded `WARMUP_BATCHES`)
- Fixed misleading profile label `GPU::H2D_Indices_f32` → `GPU::H2D_BasisIndices`
- Restored `_validate_loader_args()` call on the synthetic path

**CI / docs**
- `python-testing.yml` adds a fast `rust-check` gate (with and without the
  `QDP_NO_CUDA` stub-only build) so duplicate stub definitions or cfg
  mismatches fail in ~30 s instead of during the slower maturin build
- `DEVELOPMENT.md` documents the pre-push `QDP_NO_CUDA=1 cargo build`
  sanity check

### Test results (local, RTX 2080 Ti)

- **Rust unit + integration**: 284 / 284 passed (CPU)
- **GPU-required tests**: 93 / 93 passed
  - `tests/gpu_fidelity.rs`: 17 (cross-precision F32 vs F64 @ 8/12/16/20 q)
  - `tests/gpu_ptr_encoding.rs`: 64 (f32 zero-copy single + batch, all encodings)
  - `tests/gpu_angle_encoding.rs`: 12 (angle f32 batch + async pipeline + validation rejects)
- **Python pytest** (`qdp/qdp-python/tests`): 7 / 7 passed (22 ROCm skips on CI without AMD GPU)
- **Clippy / fmt / license-header**: clean
- **Baseline benchmark**: `run_pipeline_baseline.py` reports 13,583 vec/s median
  (within hardware noise of the original 14,301 baseline; previous host)

### Known scope limitations (intentional, follow-up tracked)

This PR delivers F32 on the **hot paths** — CUDA-tensor zero-copy and the
synthetic / file / streaming loader. The following entry points still go
through the F64 path; they are not regressions, they remain at parity with
upstream `main` before this PR:

- `engine.encode(np.array(..., dtype=np.float32))` — numpy extraction only
  accepts `PyReadonlyArray<f64>` (`engine.rs:143, 162`)
- `engine.encode(cpu_tensor.to(torch.float32))` — same numpy view path
- `engine.encode("data.parquet")` with a float32 column — `DataReader::read_batch`
  returns `Vec<f64>` (this is the pre-existing `reader.rs:106` structural
  debt, now annotated in-source as a comment)
- `loader.backend("pytorch")` reference fallback — always uses `torch.float64`
- Three loader factories hardcode `Dtype::Float32`; there is no user-facing
  `precision` argument yet (e.g. for forcing F64 to debug)

Resolving the above is the proper scope of follow-up work on generic
readers + `InputAdapter` + a user-facing `.dtype()` builder method, not
of this PR.

### Migration / compatibility

- `Encoding::supports_f32()` widened from `Amplitude` only to
  `Amplitude | Angle | Basis`. Callers gating on this method will now route
  angle / basis batches through the new f32 path. If a downstream caller
  was depending on the previous F64-only behavior for angle / basis batch
  pipelines, it must explicitly request `Dtype::Float64` in `PipelineConfig`
- `PipelineConfig::default()` now uses `prefetch_depth = 0` (auto). Callers
  reading the field directly should call `config.normalize()` before use
- `compute_optimal_prefetch_depth` signature changed:
  `(num_qubits, batch_size, &str, bool)` → `(num_qubits, batch_size, Encoding, Precision)`.
  This was a `pub(crate)` helper, but flagged for visibility

### Checklist

- [x] Added or updated documentation for all changes (`DEVELOPMENT.md`,
  `qumat_qdp/README.md`, in-source structural-debt notes)


ryankert01 (commented):
I read this pr high levelly, lg so far

400Ping (approved):
Overall LGTM

viiccwen:
F32 support for angle encoding is duplicated to #1268

rich7420:
@viiccwen yes! I think #1268 should be merged first. then I'll rebase it

rich7420:
@viiccwen , @400Ping , @ryankert01 and @guan404ming do you want to take another look?

ryankert01:
@rich7420 Main branch has been differs a lot because of a previous change. Could you help verify this change works or not?

rich7420:
@ryankert01 no problem

rich7420:
@ryankert01 I've updated the pr description

ryankert01 (approved):

rich7420:
@ryankert01 thanks!

PR #1278, opened 2026-04-22
[Feature] Add batch float32 zero-copy angle encoding in qdp-python bindings

### LINKED ISSUE #1277: [Feature] Add batch float32 zero-copy angle encoding in qdp-python bindings
### What

Add batch `float32` zero-copy angle encoding support to the Python bindings for CUDA `torch.Tensor` inputs.

Today, qdp-core already exposes batch `float32` angle encoding APIs, but the Python binding only wires the single-sample `float32` angle path. For 2D CUDA tensors shaped as `(batch_size, num_qubits)`, the binding rejects or misses the dedicated batch `float32` angle path instead of dispatching to the existing qdp-core batch encoder.

This feature should make `QdpEngine.encode(cuda_float32_batch, num_qubits, "angle")` work the same way the existing single-sample `float32` angle path does, while preserving engine-controlled output precision.

### Why

This closes a binding gap between qdp-core and qdp-python.

Without this support, users must either:

- upcast batched angle inputs to `float64`, which increases memory bandwidth and footprint, or
- loop over single samples, which defeats the zero-copy batch path.

Adding the missing batch binding keeps the scope tight while enabling the intended low-overhead CUDA workflow for angle-encoded batched inputs.

viiccwen:
### Related Issues

Closes #1277

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

qdp-core already supports batch `float32` angle encoding from CUDA pointers, but qdp-python only exposed the single-sample `float32` angle path. That left a mismatch between core capabilities and the Python binding surface, forcing batched CUDA users to upcast to `float64` or give up the zero-copy batch path.

This PR closes that binding gap with a narrowly scoped change in qdp-python only.

### How

- allow CUDA `torch.float32` tensors for `encoding_method="angle"` in the shared Python-side tensor validation path
- dispatch 2D CUDA `float32` angle tensors to `encode_angle_batch_from_gpu_ptr_f32_with_stream(...)`
- keep the existing 1D `float32` angle fast path unchanged
- add targeted GPU binding tests covering:
  - 1D `float32` angle output dtype behavior
  - 2D batch `float32` angle output dtype behavior
  - expected encoded states for a small deterministic batch
  - invalid batch feature-count handling
  - non-contiguous batch rejection
  - empty batch rejection
  - invalid `float16` rejection
  - existing `float64` batch angle regression coverage

## Checklist

### DIFF
--- a/qdp/qdp-python/src/engine.rs
+++ b/qdp/qdp-python/src/engine.rs
@@ -128,7 +128,7 @@ impl QdpEngine {
     ///     encoding_method: Encoding strategy ("amplitude" default, "angle", or "basis")
     ///         CUDA tensor note:
     ///         - amplitude accepts float64 and float32
-    ///         - angle accepts float64 generally, plus float32 for 1D single-sample tensors
+    ///         - angle accepts float64 generally, plus float32 for 1D and 2D CUDA tensors
     ///
     /// Returns:
     ///     QuantumTensor: DLPack-compatible tensor for zero-copy PyTorch integration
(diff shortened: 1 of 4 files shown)


ryankert01 (commented):
lg, need rebase

ryankert01 (approved):
lg, some nits from copilot

PR #1283, opened 2026-04-22
[Feature] Make QDP CUDA kernel build targets configurable and future-compatible

### LINKED ISSUE #1282: [Feature] Make QDP CUDA kernel build targets configurable and future-compatible
### What

Make QDP CUDA kernel build targets configurable and future-compatible instead of hardcoding a small fixed set of `nvcc -gencode` targets.

Today `qdp/qdp-kernels/build.rs` only emits a narrow cubin set, which makes it harder to support both current GPUs and newer architectures such as Blackwell without editing the build script again. It also does not provide a clean override for CI, wheel builds, or local development environments that need a different architecture mix.

### Why

QDP should be able to:

- keep working on currently used GPUs without regressing existing builds
- adopt newer NVIDIA architectures without repeatedly rewriting the build script
- let maintainers and CI choose a specific architecture list when packaging or testing
- preserve a PTX fallback path for forward compatibility where appropriate

Without this, architecture support remains tied to source edits and older `nvcc` defaults, which makes maintenance and release packaging more fragile than necessary.

### How

Update `qdp/qdp-kernels/build.rs` so that:

- the default target set is derived from a project-maintained shortlist filtered by what the local `nvcc` actually supports
- at least one PTX target is emitted for forward compatibility when supported by the toolchain
- callers can override the target set through an environment variable such as `QDP_CUDA_ARCH_LIST`
- the parser can represent newer suffixed architectures such as `90a` when supported by `nvcc`
- documentation explains how to override architecture targets for local builds and CI

This keeps the change localized to build configuration and documentation, without changing QDP runtime behavior.

viiccwen:
### Related Issues

Closes #1282

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

QDP's CUDA kernel build currently relies on a small hardcoded set of `nvcc -gencode` targets. That makes it awkward to keep existing GPUs working while also supporting newer architectures such as Blackwell, and it forces follow-up source edits whenever the desired target mix changes.

This change makes architecture targeting configurable and toolchain-aware so one build configuration can remain usable across current and newer NVIDIA GPU generations without changing QDP runtime behavior.

### How

- derive the default cubin/PTX target set from a project shortlist filtered by the local `nvcc` supported architecture lists
- add `QDP_CUDA_ARCH_LIST` as an explicit override for local builds, CI, and packaging workflows
- preserve a legacy fallback when `nvcc` does not expose architecture listing flags
- validate the change locally on both Ada and Blackwell GPUs and rerun targeted QDP GPU tests

## Checklist

### DIFF
--- a/docs/qdp/getting-started.md
+++ b/docs/qdp/getting-started.md
@@ -13,6 +13,23 @@ QDP (Quantum Data Plane) is a GPU-accelerated library for encoding classical dat
   - NVIDIA GPU with a CUDA-compatible PyTorch build (verify with `python -c "import torch; print(torch.cuda.is_available())"`)
   - AMD GPU with a ROCm-compatible PyTorch build (verify with `python -c "import torch; print(torch.version.hip)"`) plus Triton with HIP support
 
+## Supported GPU Backends
+
+QDP currently supports the following GPU backends:
+
+- NVIDIA CUDA backend
+  - CUDA builds target NVIDIA GPUs supported by the installed CUDA toolkit.
+  - The current default CUDA architecture shortlist spans common NVIDIA
+    generations from Turing through Blackwell: `sm_75`, `sm_80`, `sm_86`,
+    `sm_89`, `sm_90`, `sm_100`, and `sm_120`.
+  - The final generated CUDA targets depend on the local `nvcc` supported
+    architectures.
+
+- AMD ROCm backend
+  - ROCm builds target AMD GPUs supported by the installed ROCm stack and the
+    Triton backend used by QDP.
+  - The final supported devices depend on the local ROCm environment.
+
 ## Installation
 
 ```bash
--- a/docs/qdp/getting-started.md
+++ b/docs/qdp/getting-started.md
@@ -13,6 +13,11 @@ QDP (Quantum Data Plane) is a GPU-accelerated library for encoding classical dat
   - NVIDIA GPU with a CUDA-compatible PyTorch build (verify with `python -c "import torch; print(torch.cuda.is_available())"`)
   - AMD GPU with a ROCm-compatible PyTorch build (verify with `python -c "import torch; print(torch.version.hip)"`) plus Triton with HIP support
 
+For NVIDIA CUDA builds, QDP selects CUDA kernel architecture targets from the
+local `nvcc` supported target list and includes a PTX target when available.
+Set `QDP_CUDA_ARCH_LIST` to override the generated targets, for example
+`QDP_CUDA_ARCH_LIST="89,90,120+PTX"`.
+
 ## Installation
 
 ```bash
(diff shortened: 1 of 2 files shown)


viiccwen:
cc @ryankert01, @rich7420, need testing on ur local machine.
Already testing in both RTX 4090 and newer architectures Pro 6000 Blackwell. It's all work well.

viiccwen:
@ryankert01 sounds like **`getting-started.md`**. I'll woke on it.

ryankert01:
sorry @viiccwen, I thought we have a section say we only support some Nvidia gpus (30, 40), but we turns out don't have this section. So, I think it's not needed.

Also, just a heads up, the 30s gpu server that I have access to is currently compromised. So I might not be able to test it anywhere soon.

400Ping:
Agree on writing a section on the current supporting GPUs(Including Nvidia and AMD).

viiccwen:
Goti it, I think a backend-level section is more accurate than listing specific GPU models.

For NVIDIA, QDP does not maintain a fixed supported-SKU whitelist. The actual CUDA targets generated by the build depend on the installed CUDA toolkit and the local `nvcc` supported architectures, with QDP selecting from its default architecture shortlist.

For AMD, support similarly depends on the local ROCm environment and the Triton backend used by QDP, rather than a hardcoded model list in the repo.

Because of that, I plan to document this as a “Supported GPU Backends” section instead of list a set of supported GPU models.

ryankert01 (commented):
tested!

rich7420:
tested!

rich7420:
tested locally on a host with cuda 12.5 + rtx 2080 ti (sm_75). all green.

**default path (no `QDP_CUDA_ARCH_LIST`)**
`cuobjdump --list-elf` on the built amplitude object:
```
amplitude.sm_75.cubin
amplitude.sm_80.cubin
amplitude.sm_86.cubin
amplitude.sm_89.cubin
amplitude.sm_90.cubin
PTX: amplitude.sm_90.ptx
```
nvcc 12.5 correctly filtered `sm_100` and `sm_120` out of `DEFAULT_CUBIN_ARCHES` since they're not yet supported by this toolchain — exactly the intended behavior. ptx falls back to the highest supported (sm_90).

**explicit list `QDP_CUDA_ARCH_LIST=75,80`** → only `sm_75` + `sm_80` cubins, no ptx.

**`QDP_CUDA_ARCH_LIST=75+PTX,80`** → `sm_75` cubin + `sm_75` ptx + `sm_80` cubin. `+PTX` suffix parser works.

**invalid input `QDP_CUDA_ARCH_LIST=invalid!!`** → clean panic at build time:
```
Invalid CUDA architecture 'invalid!!' in QDP_CUDA_ARCH_LIST.
Expected entries like '89', '90a', or '120+PTX'.
```

**empty after parse `QDP_CUDA_ARCH_LIST=,,,`** → also a clean panic.

**`QDP_NO_CUDA=1`** → bypasses arch detection entirely; no-cuda stub build still succeeds even with an invalid arch list set.

**runtime smoke (after default rebuild)**: `cargo test --test gpu_fidelity --release` → 17/17 pass. f32 vs f64 fidelity at 8/12/16/20 qubits stays ≥0.9999996.

minor nit (optional, not blocking):
- the `parse_arch_name` regex allows trailing letters (so `90a` is accepted) — looks intentional for the `sm_90a` variant. worth a comment in the doc-comment to confirm that's deliberate vs accidental, since `120+PTX` is mentioned in the error but `90a` only shows up in the error text without a corresponding test.

happy to approve. nice work :)

ryankert01:
looks like its okay, i tested at 3090ti last week btw.

PR #1158, opened 2026-03-08
[Feature][QDP] Add AMD GPU support via Triton backend

400Ping:
### Related Issues

Closes #1155

### Changes

- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [x] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

QDP currently assumes the CUDA/native `_qdp` path, which makes the Python API effectively NVIDIA-only.
This change adds an AMD/ROCm execution path so QDP can run on AMD GPUs while keeping the existing CUDA route intact.

The goal of this PR is to:

- reduce NVIDIA vendor lock-in for QDP users
- keep a single `QdpEngine` entry point for backend selection
- document and test the AMD/ROCm route without regressing the existing CUDA/native path

### User-Facing Behavior

CUDA:

```python
from qumat.qdp import QdpEngine

engine = QdpEngine(device_id=0, backend="cuda")
qtensor = engine.encode(data, num_qubits=2, encoding_method="amplitude")
```

AMD/ROCm:

```python
from qumat.qdp import QdpEngine

engine = QdpEngine(device_id=0, backend="amd", precision="float32")
tensor = engine.encode(data, num_qubits=2, encoding_method="amplitude")
```

## Checklist

- [x] Added or updated unit tests for all changes
- [x] Added or updated documentation for all changes

PR #1291, opened 2026-04-25
[QDP] Add Mahout-AMD and PennyLane-AMDGPU frameworks to throughput benchmark

ryankert01:
### Related Issues

Follow-up to #1289 (AMD backend selection in encoding benchmarks).

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [x] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

## Summary

Extends the canonical throughput benchmark
(`benchmark_throughput.py` — Mahout vs PennyLane vs Qiskit) with three
new frameworks so users on AMD ROCm hosts can benchmark on equal
footing with the existing CUDA paths:

- **`mahout-amd`** — QDP AMD path via `QdpEngine(backend="amd")`
  (the TritonAmdEngine landed in #1158, exposed via the public router
  in #1289).
- **`pennylane-amdgpu`** — PennyLane `lightning.amdgpu`, the official
  ROCm simulator (Kokkos+HIP backend).
- **`pytorch-ref`** — Pure-PyTorch reference implementation
  (`qumat_qdp.torch_ref.amplitude_encode`). Same workload, no engine
  wrapper. Useful as a "what naive PyTorch on the same hardware can
  do" ceiling — gaps between this and `mahout-amd` quantify any
  per-call overhead in the AMD engine adapter.

Both names are added to `FRAMEWORK_CHOICES`, dispatched in `main()`,
and included in the speedup-ratio summary. Existing CUDA-only and
CPU-only hosts are unaffected; the new frameworks auto-skip with a
clear message when their runtimes aren't available.

While here, also fixed three pre-existing methodology issues that were
inflating Mahout's reported speedup vs PennyLane:

1. **`run_pennylane` was casting `complex64 → float32`** for the GPU
   transfer (line `state_cpu.to("cuda", dtype=torch.float32)`),
   discarding the imaginary part of the encoded state. PyTorch's
   `Casting complex values to real discards the imaginary part`
   warning fired on every batch. This silently produced wrong results
   AND inflated Mahout's win because the broken pennylane path was
   ~4× slower than a correct one.
2. **No warmup** anywhere — first-batch costs (Triton AMD JIT
   autotune, PennyLane QNode tracing, Kokkos device init, AerSim
   transpile cache) were inside every timer. Added `WARMUP_BATCHES = 3`
   used by all runners.
3. **Dtype mismatch** — Mahout-AMD ran float32, both PennyLane runners
   ran float64. All runners now use float32 input.

## Verified on AMD Instinct MI300X (ROCm 7.2 / torch 2.9.0+rocm6.4 / pennylane 0.44.1)

End-to-end DataLoader → encode → consumer pipeline, batch_size=64,
3 warmup batches not timed, all runners on float32 input:

| Config | Mahout-AMD | PyTorch-ref | PennyLane | PL-AMDGPU |
|---:|---:|---:|---:|---:|
| q=8, 12800 samples  | 112,650 vec/s | **119,815** | 48,367 | 1,492 |
| q=12, 12800 samples | **4,918**     | 4,163      | 1,755  |   658 |
| q=16, 6400 samples  |     890       |   **926**  |   704  |   347 |

Mahout-AMD speedup ratios:

| | vs PyTorch-ref | vs PennyLane | vs PennyLane-AMDGPU |
|---|---:|---:|---:|
| q=8  | 0.94× | 2.33× | 75.5× |
| q=12 | 1.18× | 2.80× |  7.48× |
| q=16 | 0.96× | 1.27× |  2.57× |

**Honest takeaways:**

* **Mahout-AMD essentially ties PyTorch-ref** (±6–18% across qubit
  sizes, within run-to-run noise). The AMD path's TritonAmdEngine is
  not adding kernel-level speedup over the same operation expressed
  in plain PyTorch ops. **The Mahout-AMD value-add on AMD is not
  speed; it's the unified Mahout API surface and DLPack-zero-copy
  integration story.** Closing or beating the PyTorch-ref gap
  meaningfully would require real `@triton.jit` kernels — flagged
  as follow-up.
* **Mahout-AMD wins 1.27–2.80× over PennyLane `default.qubit` GPU.**
  Both paths run real GPU compute on the same MI300X (verified via
  `torch.cuda.memory_allocated` growth and 40× CPU-vs-CUDA speedup);
  the win is from skipping PennyLane's `MottonenStatePreparation`
  decomposition (~2^(n+1) gate ops) and writing the state vector
  directly. Wins narrow at large states (compute-bound regime).
* **`lightning.amdgpu` is 2.6–75× behind everything else** because
  it doesn't broadcast over the batch dimension for
  `AmplitudeEmbedding`; the per-sample QNode dispatch (~1 ms each in
  PennyLane v0.44) dominates. This is the public-API reality today,
  not a kernel-level comparison.

Earlier draft of this PR claimed "5.6×" against PennyLane — that was
inflated by a `complex64 → float32` cast in `run_pennylane` that
silently dropped the imaginary part of the encoded state, making the
PennyLane baseline ~4× slower than a correct one. Fixed in this PR;
the corrected number is 1.6–2.8×.

## Runtime caveat: lightning.amdgpu loader

The `pennylane-lightning-amdgpu` wheel needs `libhsa-runtime64.so.1`
and `libamdhip64.so.7` matching the system ROCm install. Ubuntu 24.04
ships an older `libhsa-runtime64.so.1` (from ROCm 5.7) at
`/lib/x86_64-linux-gnu/`, which shadows newer symbols
(e.g. `hsa_amd_memory_get_preferred_copy_engine`) the plugin requires.

The script handles this by RTLD_GLOBAL-pre-loading the matching
ROCm 7.x libs at **module top**, gated on **`MAHOUT_PRELOAD_ROCM=1`**
(only — `ROCM_LIB_DIR` alone no longer auto-triggers, to avoid
surprising other code that imports this module). The preload **must**
happen before `torch` / `pennylane` import — doing it later deadlocks
because torch's HIP runtime has already mapped the older libhsa.

Usage on a ROCm host:

```bash
MAHOUT_PRELOAD_ROCM=1 ROCM_LIB_DIR=/opt/rocm-7.2.0/lib \
  uv run python benchmark/benchmark_throughput.py \
  --frameworks mahout-amd,pennylane,pennylane-amdgpu --qubits 12
```

If `pennylane-lightning-amdgpu` isn't installed or the libs aren't
found, the framework gracefully skips with a hint pointing at
`MAHOUT_PRELOAD_ROCM` + `ROCM_LIB_DIR`.

## Test plan

- [x] CUDA host: existing frameworks (`pennylane`, `qiskit`,
  `mahout`) unchanged — all still call `prefetched_batches` /
  `normalize_batch` the same way, plus they now warm up.
- [x] AMD host: `mahout-amd` runs end-to-end on MI300X (3 sizes
  verified above).
- [x] AMD host: `pennylane-amdgpu` runs end-to-end with
  `MAHOUT_PRELOAD_ROCM=1 ROCM_LIB_DIR=...`.
- [x] AMD host without `pennylane-lightning-amdgpu` installed:
  graceful skip.
- [x] Host without preload env: `pennylane-amdgpu` fails with clear
  hint pointing to the env vars.
- [x] Mixed framework selection (`mahout-amd,pennylane`) works
  without ROCm libs preloaded.
- [x] `ruff check` and `ruff format` clean.

## Checklist

- [ ] Added or updated unit tests for all changes — N/A (this is a
  benchmark script; runs are validated by the verification table
  above)

### DIFF
--- a/qdp/qdp-python/benchmark/benchmark_throughput.py
+++ b/qdp/qdp-python/benchmark/benchmark_throughput.py
@@ -54,6 +54,7 @@ import ctypes
 import os
 import time
 
+
 # IMPORTANT: preload system ROCm 7.x libs *before* importing torch / pennylane.
 # Once torch's HIP runtime maps the older libhsa-runtime from
 # /lib/x86_64-linux-gnu (Ubuntu 24.04 ships ROCm 5.7), a later RTLD_GLOBAL of
@@ -77,11 +78,11 @@ if os.environ.get("MAHOUT_PRELOAD_ROCM") == "1" or os.environ.get("ROCM_LIB_DIR"
     _preload_rocm_libs_at_import()
 
 
-import numpy as np  # noqa: E402
-import torch  # noqa: E402
-from qumat_qdp import QdpBenchmark  # noqa: E402
+import numpy as np
+import torch
+from qumat_qdp import QdpBenchmark
 
-from benchmark.utils import normalize_batch, prefetched_batches  # noqa: E402
+from benchmark.utils import normalize_batch, prefetched_batches
 
 BAR = "=" * 70
 SEP = "-" * 70
@@ -107,8 +108,6 @@ try:
     HAS_QISKIT = True
 except ImportError:
     HAS_QISKIT = False
-
-
 
 
 def parse_frameworks(raw: str) -> list[str]:
@@ -197,9 +196,7 @@ def run_pennylane(num_qubits: int, total_batches: int, batch_size: int, prefetch
     return duration, throughput
 
 
-def run_mahout_amd(
-    num_qubits: int, total_batches: int, batch_size: int, prefetch: int
-):
+def run_mahout_amd(num_qubits: int, total_batches: int, batch_size: int, prefetch: int):
     """Run Mahout AMD path (TritonAmdEngine) end-to-end through the prefetch pipeline."""
     try:
         from qumat_qdp import QdpEngine, is_triton_amd_available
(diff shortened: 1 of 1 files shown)


ryankert01:
lg, cc @400Ping

400Ping (approved):
Overall LGTM

400Ping:
I only did a quick scan but since it is kind of urgent, so I am merging this.

PR #1296, opened 2026-05-04
[Feature][QDP] Add MPI-ready distributed amplitude execution scaffolding

viiccwen:
### Related Issues

Closes #1295
Part of #1297

### Summary

This PR adds the first QDP-native distributed GPU state execution foundation, with amplitude state construction as the first workload.

The implementation is intentionally **MPI-ready** but does **not** require MPI or NCCL as mandatory dependencies. The executable path in this PR is a real single-process fallback that runs through the same rank-local planning, runtime, transport, and state boundaries that a future MPI-backed execution path should use.

In other words, this PR is not claiming full `mpirun -n N` support yet. It establishes the architecture and a testable fallback boundary that future MPI, CUDA-aware MPI, and NCCL backends can replace or extend.

The main architecture boundary is intentionally split into four layers:

- **Planning** decides where shards should live.
- **Runtime** decides what work the current rank performs.
- **Transport** owns scalar and future GPU-resident collectives.
- **State** owns local GPU shard buffers and exposes local zero-copy views.

## Current Capability

This PR currently supports:

- rank/world execution metadata through `DistributedExecutionContext`
- rank-local device meshes and placement planning
- shared distributed state planning through `DistributedStatePlan`
- amplitude-specific execution through `DistributedAmplitudePlan`
- local GPU shard materialization for shards owned by the current rank
- scalar rank-level reduction through `CollectiveCommunicator`
- single-process fallback through `LocalCollectiveCommunicator`
- zero-copy **local** shard metadata through `DistributedStateVector::local_shard_views()`
- planned `device_id` to CUDA handle resolution, so shard metadata, active CUDA device context, and allocated buffers stay aligned
- placeholder CUDA-aware MPI / NCCL device collective types that identify their backend but return `NotImplemented` for GPU-resident collectives

### Architecture Shape

```mermaid
sequenceDiagram
    participant Caller as QdpEngine caller
    participant Engine as QdpEngine
    participant Ctx as DistributedExecutionContext
    participant Planner as PlacementPlanner
    participant StatePlan as DistributedStatePlan
    participant AmpPlan as DistributedAmplitudePlan
    participant Runtime as distributed runtime
    participant Comm as CollectiveCommunicator
    participant State as DistributedStateVector

    Caller->>Engine: submit distributed amplitude request
    Engine->>Engine: validate input and resolve request
    Engine->>Ctx: construct rank-local execution context
    Engine->>Planner: build rank-local placement plan
    Planner-->>Engine: placement + shard ranges
[...]

PR #1298, opened 2026-05-05
[Feature][QDP] Add MPI-ready distributed amplitude execution scaffolding

viiccwen:
### Related Issues
- None yet.

### Changes
- add distributed amplitude planning, layout, and state scaffolding for multi-GPU execution
- introduce a distributed execution context and collective seam that are MPI-ready without requiring an MPI runtime yet
- fix placement-to-device binding so reordered placements materialize buffers on the correct CUDA devices
- add a distributed q34 probe and distributed runtime / planner / topology / communicator test coverage
- update CUDA arch targeting so the probe builds on the GPUs available on this host

### Why
- establish a QDP-native distributed amplitude foundation that can scale beyond a single GPU
- keep the execution model extensible toward MPI-backed collectives without hard-coding Lightning-specific assumptions
- verify that the current implementation can successfully materialize and encode a 34-qubit float32 distributed state on this host

### How
- keep planning, layout, runtime, and collective responsibilities separate
- drive shard materialization from placement device IDs rather than raw mesh iteration order
- package mesh plus collectives into a single distributed execution context
- validate the path with distributed tests and an executable q34 probe

## Checklist
- [x] Feature
- [x] Tests added or updated
- [x] No Python surface changes
- [x] Scope kept to QDP distributed amplitude groundwork
- [ ] MPI multi-rank launcher path implemented

### DIFF
--- a/qdp/qdp-core/examples/distributed_multigpu_q34_probe.rs
+++ b/qdp/qdp-core/examples/distributed_multigpu_q34_probe.rs
@@ -16,8 +16,11 @@
 
 use std::time::Instant;
 
-use qdp_core::{DistributionMode, MahoutError};
-use qdp_core::{HostCommunicator, PlacementRequest, Precision, QdpEngine, ShardPolicy};
+use qdp_core::gpu::LocalCollectiveCommunicator;
+use qdp_core::{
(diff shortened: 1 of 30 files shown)


viiccwen:
Closing this PR because the changes have been moved onto PR #1296 as intended.

PR #1311, opened 2026-05-12
feat(qdp): add Backend enum for compute-target classification

rich7420:
### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

Callers (Python bindings, benchmarks, future planning layers) currently have no first-class way to ask "what compute target is the Rust pipeline running against?". They either sniff engine state or rely on string-based feature detection.

This adds a small `Backend` enum so that question has a typed answer.

It is **not** a dispatch axis: there is currently only one Rust-side encoder implementation (CUDA / nvcc), so a 2-D `(Encoding, Backend) -> &'static dyn QuantumEncoder` table would have nothing extra to dispatch to. The AMD / Triton path that already exists lives entirely on the Python side (`qumat_qdp/backends/amd.py`) and does not have a matching Rust-level `QuantumEncoder` implementation — so it is **not** modelled here as a variant. Grow the enum and the dispatch table at the same time when a Rust-side AMD path is proposed, not before.

### How

**Variants** (`qdp-core/src/types.rs`)

- `CudaNvidia` — in-tree CUDA kernels (the historical default)
- `TorchRef`   — pure-PyTorch reference, used by CPU / non-CUDA CI lanes

**API**

- `Backend::as_str()`      — stable lower-case name for logging (`"cuda-nvidia"` / `"torch-ref"`)
- `Backend::from_str_ci()` — case-insensitive parse with convenience aliases (`cuda` / `nvidia` map to `CudaNvidia`; `torch` / `torchref` / `cpu` map to `TorchRef`). Names that look like AMD (`amd`, `triton`, `triton-amd`) get a dedicated error message pointing at the Python-side path so users don't think the Rust core silently dropped the flag.
- `Backend::detect()`      — compile-time best guess (Linux → `CudaNvidia`, else `TorchRef`). Does **not** probe runtime CUDA availability; that remains the engine's responsibility.

`Backend` is re-exported from `qdp_core::` next to `Encoding` / `Dtype` so callers can `use qdp_core::Backend;`.

**Tests** (`qdp-core/tests/types.rs`) — 5 new

- case-insensitive parse + aliases + whitespace tolerance
- AMD names rejected with a message mentioning both "AMD" and "Python"
- unknown names rejected
- `as_str` round-trips through `from_str_ci` for every variant
- `detect()` returns the platform-appropriate variant

### Local verification

- `cargo test --tests` on `qdp-core`: all passing (11 type tests including the 5 new ones)
- Clippy / fmt / license-header: clean
- No call sites touched — this is a new public type, not a dispatch change

## Checklist

### DIFF
--- a/qdp/qdp-core/src/lib.rs
+++ b/qdp/qdp-core/src/lib.rs
@@ -38,7 +38,7 @@ mod profiling;
 pub use error::{MahoutError, Result, cuda_error_to_string};
 pub use gpu::memory::Precision;
 pub use reader::{NullHandling, handle_float64_nulls};
-pub use types::{Dtype, Encoding};
+pub use types::{Backend, Dtype, Encoding};
 
 // Throughput/latency pipeline runner: single path using QdpEngine and encode_batch in Rust.
 #[cfg(target_os = "linux")]
--- a/qdp/qdp-core/src/types.rs
+++ b/qdp/qdp-core/src/types.rs
@@ -158,3 +158,99 @@ impl Encoding {
         }
     }
 }
+
+/// Compute backend used for the Rust-side encoding pipeline.
+///
+/// Today this is a thin, 1-D classification used by callers (Python bindings,
+/// benchmarks, future planning layers) to ask "what compute target is the
+/// Rust pipeline running against?". It is intentionally **not** plumbed through
+/// [`crate::pipeline_runner::PipelineConfig`] as a dispatch axis: there is
+/// currently only one Rust-side encoder implementation (CUDA / nvcc), so a 2-D
+/// `(Encoding, Backend) → &'static dyn QuantumEncoder` table would have nothing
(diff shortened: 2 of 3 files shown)


rich7420:
closing this one for now after re-evaluating the plan.

the original motivation was a 2-D `(encoding, backend)` dispatch table to route between cuda and amd encoders in rust. but the amd path landed entirely in python (`qumat_qdp/backends/amd.py`), so there's no rust-side amd encoder for the enum to dispatch to. that makes the enum decorative — no caller actually needs it right now.

revisiting if/when:
- a rust-side amd impl is proposed, or
- a future feasibility planner needs to introspect compute target as a typed value

branch stays on origin so it's easy to revive. no rush.

PR #1321, opened 2026-05-17
fix(qdp-core): link cudart locally and stub FFI when toolkit is absent 

### LINKED ISSUE #1318: [Bug] qdp-core does not own its libcudart link directive; build fails on driver-only systems
### Summary
On Linux systems with the NVIDIA driver installed but without the CUDA Toolkit
(no `nvcc`, no `libcudart.so` — a common setup, since PyTorch ships its own
bundled cudart inside the wheel), `make test_rust` fails with confusing
linker errors like:
 
```
rust-lld: error: undefined symbol: cudaMemGetInfo
rust-lld: error: undefined symbol: cudaHostAlloc
rust-lld: error: undefined symbol: cudaEventCreateWithFlags
... (12 such symbols)
```
 
### Root cause
`qdp/qdp-core/src/gpu/cuda_ffi.rs` declares CUDA Runtime API symbols in an
unconditional `unsafe extern "C" { ... }` block, but `qdp-core` itself does
not emit any `cargo:rustc-link-lib=cudart` directive. The only such directive
lives in `qdp/qdp-kernels/build.rs` and is emitted only when `nvcc` is found
on `PATH`.
 
Build-script `cargo:rustc-cfg` flags are scoped to the crate that emits them.
So when `qdp-kernels`' build script can't find `nvcc`, it sets a
`qdp_no_cuda` cfg **for itself** (correctly stubbing out kernel launchers)
but `qdp-core` has no way to learn about this and keeps declaring extern
references to cudart symbols that nothing links.
 
`qdp/DEVELOPMENT.md` documents a "no-CUDA build" sanity check using
`cargo build --workspace --lib --release` and `cargo check --workspace --tests`
(emphasis on `check`, not `test`) — so the linker hole is invisible in CI but
trips real users who run the test target.
 
### Reproduction
On a Linux box with NVIDIA driver but no CUDA Toolkit:
```
make test_rust
```
 
### Expected
Either the tests link and run (toolkit present), or the build skips them with
a clear message ("CUDA Toolkit required for Rust integration tests").
 
### Actual
Confusing 50+ line linker error pointing at 12 undefined cudart symbols.

andrewmusselman:
Closes #1318 .
 
`qdp-core/build.rs` (extended; it already existed for protoc) now:
- probes for `nvcc` with the same logic as `qdp-kernels/build.rs`,
- emits `cargo:rustc-link-search=native=$CUDA_PATH/lib64` and
  `cargo:rustc-link-lib=cudart` when found,
- emits `cargo:rustc-cfg=qdp_no_cuda` when not found, with a clear
  `cargo:warning` pointing at the toolkit install,
- respects `QDP_NO_CUDA=1` for explicit forcing (matching `qdp-kernels`).
`qdp-core/src/gpu/cuda_ffi.rs` wraps the `extern "C"` block in
`#[cfg(not(qdp_no_cuda))]` and adds matching `#[cfg(qdp_no_cuda)] mod
no_cuda_stubs { ... }` with `pub(crate) unsafe fn` stubs for all 14
declared functions. Each stub returns `999` — the same sentinel
`qdp-kernels` uses for its kernel-launcher stubs — so existing caller
error paths (`if ret != 0 { return Err(...) }`) surface a clean runtime
error if anyone calls a CUDA function on a no-toolkit build, instead of
failing at link time. The whole stub module is wrapped in a single
`#[allow(non_snake_case)]` since the originals are camelCase to match
the real CUDA Runtime API.
 
Cross-crate behaviour after this PR:
| environment              | qdp-kernels       | qdp-core         | result                  |
|--------------------------|-------------------|------------------|-------------------------|
| toolkit installed        | links cudart      | links cudart     | normal GPU build        |
| driver only / macOS / CI | stub launchers    | stub Runtime API | links; runtime err 999  |
| `QDP_NO_CUDA=1`          | stub launchers    | stub Runtime API | links; runtime err 999  |
 
Verified on Linux + CUDA 12.4: `cargo build --workspace --tests
--exclude qdp-python` succeeds both with and without `QDP_NO_CUDA=1`;
`make test_rust` runs full integration tests on the GPU; 12 lint
warnings introduced by the stubs are silenced.

### DIFF
--- a/qdp/qdp-core/build.rs
+++ b/qdp/qdp-core/build.rs
@@ -64,6 +64,9 @@ fn configure_cuda_linkage() {
     println!("cargo::rustc-check-cfg=cfg(qdp_no_cuda)");
     println!("cargo:rerun-if-env-changed=QDP_NO_CUDA");
     println!("cargo:rerun-if-env-changed=CUDA_PATH");
+    // The whole CUDA-vs-stub decision hinges on finding nvcc on PATH, so a PATH
+    // change (e.g. installing the toolkit) must re-trigger this script.
+    println!("cargo:rerun-if-env-changed=PATH");
 
     let force_no_cuda = env::var("QDP_NO_CUDA")
         .map(|v| v == "1" || v.eq_ignore_ascii_case("true") || v.eq_ignore_ascii_case("yes"))
@@ -73,7 +76,8 @@ fn configure_cuda_linkage() {
         && Command::new("nvcc")
             .arg("--version")
             .output()
-            .is_ok();
+            .map(|o| o.status.success())
+            .unwrap_or(false);
 
     if !has_cuda {
         println!("cargo:rustc-cfg=qdp_no_cuda");
--- a/qdp/qdp-core/build.rs
+++ b/qdp/qdp-core/build.rs
@@ -14,10 +14,18 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::env;
+use std::process::Command;
+
 fn main() {
+    compile_protos();
+    configure_cuda_linkage();
+}
+
+fn compile_protos() {
     // Use vendored protoc to avoid missing protoc in CI/dev environments
     unsafe {
-        std::env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
+        env::set_var("PROTOC", protoc_bin_vendored::protoc_bin_path().unwrap());
     }
 
     let mut config = prost_build::Config::new();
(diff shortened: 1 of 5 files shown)


ryankert01:
Like the high-level idea:

- With CUDA toolkit installed: normal GPU-enabled build.
- Without CUDA toolkit: build still links successfully, but CUDA calls return a controlled runtime error.
- With QDP_NO_CUDA=1: same forced no-CUDA behavior.

qdp/qdp-core/build.rs
```diff
@@ -34,3 +42,54 @@ fn main() {
...
+/// Toolkit. Previously the only `-lcudart` directive lived in `qdp-kernels`'
+/// build script, and the `qdp_no_cuda` cfg it sets does not propagate
+/// cross-crate. The result was confusing linker errors on systems that have
+/// the NVIDIA driver but not the toolkit (e.g. PyTorch-only setups, where
+/// PyTorch ships its own bundled cudart inside the wheel).
+///
+/// This function:
+///   * emits `cargo:rustc-link-lib=cudart` and the appropriate
+///     `cargo:rustc-link-search` path when nvcc is found, and
+///   * emits `cargo:rustc-cfg=qdp_no_cuda` when it is not, gating the
```
ryankert01, line 60: `output().is_ok()` is true even if nvcc exits non-zero — a half-installed nvcc would set `has_cuda = true` and fall through to a link error. Suggest `.map(|o| o.status.success()).unwrap_or(false)`. Same idiom in `qdp-kernels/build.rs:177`; fix both so they can't disagree.

qdp/qdp-core/build.rs
```diff
@@ -34,3 +42,54 @@ fn main() {
 
     println!("cargo:rerun-if-changed=proto/tensor.proto");
 }
+
+/// Detect the CUDA Runtime toolkit and emit the appropriate link directives.
+///
+/// `qdp-core` declares CUDA Runtime API extern symbols in `src/gpu/cuda_ffi.rs`
+/// (cudaHostAlloc, cudaMemGetInfo, cudaEventCreateWithFlags, ...). Those symbols
+/// must be resolved at link time, which requires `libcudart` from the CUDA
```
ryankert01, line 50: Missing `cargo:rerun-if-env-changed=PATH`. The whole decision hinges on nvcc-on-PATH, so installing CUDA after a stub build won't re-trigger the script until `cargo clean`. Same gap in qdp-kernels.

ryankert01 (commented):
LGTM. Two small things worth doing. Also, needs to fix the pre-commit.

ryankert01 (approved):

Issue #1319, opened 2026-05-16
[Bug] make test_rust is gated on nvidia-smi but actually needs nvcc


andrewmusselman:
### Summary
The `Makefile`'s `HAS_NVIDIA` check uses `nvidia-smi -L`, which succeeds with
just the NVIDIA driver. `test_rust` then runs `cargo llvm-cov test`, which
links against `libcudart` from the CUDA **Toolkit** — a separate install. On
driver-only systems the gate lets the target run anyway and produces the
linker errors described in Issue 1 (or, after Issue 1 is fixed, succeeds at
link but with stub CUDA functions that all return error 999 — still not a
useful test run).
 
### Fix
See PR #2. Adds a distinct `HAS_NVCC` probe and uses it to gate `test_rust`,
with a clear message that points users at the toolkit install when only the
driver is present.
 
---
 
### Summary
`torch.cuda.is_available()` returns `True` whenever a usable driver and at
least one GPU are present, but does not verify that the GPU's compute
capability is in the current PyTorch wheel's compiled arch list. Recent
PyTorch wheels target sm_70+ only. On a Pascal-or-older GPU (e.g. GTX 1060,
sm_61) the loader and several tests pick `cuda:0` and then every kernel
launch fails with `cudaErrorNoKernelImageForDevice`.
 
Currently fails: 12 tests in `testing/qdp_python/test_fallback.py` and
`testing/qdp_python/test_torch_ref.py`.
 
PyTorch itself emits this warning at import:
> NVIDIA GeForce GTX 1060 with Max-Q Design with CUDA capability sm_61 is
> not compatible with the current PyTorch installation. The current PyTorch
> install supports CUDA capabilities sm_70 sm_75 sm_80 sm_86 sm_90 sm_100
> sm_120.
 
…then proceeds to fail at first kernel launch because nothing reads it.
 
### Affected code
- `qdp/qdp-python/qumat_qdp/loader.py:481` — picks `cuda:N` based on
  `torch.cuda.is_available()` alone.
- `testing/qdp_python/test_torch_ref.py::TestDevicePlacement::test_gpu_output` —
  `@skipif(not torch.cuda.is_available())` lets incompatible GPUs through.
- `testing/qdp_python/test_torch_ref.py::TestCrossValidation::test_encoding_matches_rust` —
  same.

PR #1322, opened 2026-05-17
Makefile: detect the CUDA Toolkit, not just the driver, for test_rust

andrewmusselman:
Fixes Issue #1319 .
 
Adds a `HAS_NVCC` probe alongside the existing `HAS_NVIDIA` and gates
`test_rust` on the new variable. When the driver is present but the
toolkit is not, prints a message explaining the distinction and pointing
at the install path. `test_python` continues to use `HAS_NVIDIA` since
— after PR #1 — `maturin develop` can build the Rust extension against
the stub CUDA Runtime symbols.
 
Verified on Linux with toolkit absent: `make test_rust` prints the new
skip message instead of attempting the build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,8 +16,16 @@
 
 .PHONY: test_rust test_python tests pre-commit setup-test-python install-llvm-cov setup-benchmark
 
-# Detect NVIDIA GPU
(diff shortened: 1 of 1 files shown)


ryankert01 (approved):

ryankert01:
lg

Issue #1320, opened 2026-05-16
[Bug] PyTorch device selection doesn't check GPU compute capability; incompatible GPUs error instead of skip


andrewmusselman:
### Summary
`torch.cuda.is_available()` returns `True` whenever a usable driver and at
least one GPU are present, but does not verify that the GPU's compute
capability is in the current PyTorch wheel's compiled arch list. Recent
PyTorch wheels target sm_70+ only. On a Pascal-or-older GPU (e.g. GTX 1060,
sm_61) the loader and several tests pick `cuda:0` and then every kernel
launch fails with `cudaErrorNoKernelImageForDevice`.
 
Currently fails: 12 tests in `testing/qdp_python/test_fallback.py` and
`testing/qdp_python/test_torch_ref.py`.
 
PyTorch itself emits this warning at import:
> NVIDIA GeForce GTX 1060 with Max-Q Design with CUDA capability sm_61 is
> not compatible with the current PyTorch installation. The current PyTorch
> install supports CUDA capabilities sm_70 sm_75 sm_80 sm_86 sm_90 sm_100
> sm_120.
 
…then proceeds to fail at first kernel launch because nothing reads it.
 
### Affected code
- `qdp/qdp-python/qumat_qdp/loader.py:481` — picks `cuda:N` based on
  `torch.cuda.is_available()` alone.
- `testing/qdp_python/test_torch_ref.py::TestDevicePlacement::test_gpu_output` —
  `@skipif(not torch.cuda.is_available())` lets incompatible GPUs through.
- `testing/qdp_python/test_torch_ref.py::TestCrossValidation::test_encoding_matches_rust` —
  same.

PR #1323, opened 2026-05-17
qumat_qdp: intersect GPU compute capability with PyTorch's arch list

andrewmusselman:
Fixes Issue #1320 .
 
Adds `_select_torch_device(torch, device_id)` in `qumat_qdp.loader`. It:
- returns `"cpu"` when CUDA isn't available,
- raises `ValueError` on out-of-range `device_id` (preserves prior
  contract),
- checks `torch.cuda.get_device_capability(device_id)` against
  `torch.cuda.get_arch_list()` and falls back to `"cpu"` with a clear
  `warnings.warn` when the device's `sm_NN` isn't in the list,
- otherwise returns `f"cuda:{device_id}"`.
Both `qumat_qdp.loader.QuantumDataLoader._create_pytorch_iterator` and
`qumat_qdp.api.QdpBenchmark._run_throughput_pytorch` use the new helper
(the latter previously duplicated the same incomplete selection logic).
 
`testing/qdp_python/test_torch_ref.py` gets a mirror helper
`_torch_cuda_usable()` used by the two `@skipif`s that previously only
checked `is_available()`.
 
After this change, on an incompatible GPU:
- 8 pytorch-backend loader tests + 4 benchmark tests silently fall back
  to CPU and pass (each emits one `UserWarning`),
- the 5 explicit GPU tests skip with `"CUDA not available or GPU
  compute capability not supported by this PyTorch build"`.
Verified on Linux + GTX 1060 (sm_61) with PyTorch wheel targeting
sm_70+: tests that previously errored with
`cudaErrorNoKernelImageForDevice` now pass or skip cleanly.

### DIFF
--- a/qdp/qdp-python/qumat_qdp/api.py
+++ b/qdp/qdp-python/qumat_qdp/api.py
@@ -189,17 +189,10 @@ class QdpBenchmark:
     def _run_throughput_pytorch(self) -> ThroughputResult:
         import torch
 
+        from qumat_qdp.loader import _select_torch_device
         from qumat_qdp.torch_ref import encode
 
-        if torch.cuda.is_available():
-            if self._device_id < 0 or self._device_id >= torch.cuda.device_count():
-                raise ValueError(
-                    f"Invalid CUDA device_id {self._device_id}; "
-                    f"{torch.cuda.device_count()} device(s) available."
-                )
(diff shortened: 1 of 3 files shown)


ryankert01 (approved):
Make sense! Let copilot double check for me.

PR #1324, opened 2026-05-18
fix(qdp): guard batch size multiplication against overflow

rich7420:
## Summary

- `Preprocessor::validate_batch` and `AmplitudeEncoder::encode_batch_f32` computed `num_samples * sample_size` without overflow checks. The angle encoder paths already use `checked_mul` (angle.rs:87, :143, :327, :405, :528) — these two sites were missed.
- On 64-bit hosts the product can wrap, letting an undersized buffer pass the length check and causing out-of-bounds reads downstream in the encoder.
- This PR mirrors the angle.rs pattern: `checked_mul` returning `MahoutError::InvalidInput` on overflow.

## Changes

- `qdp/qdp-core/src/preprocessing.rs`: guard `num_samples * sample_size` in `validate_batch`.
- `qdp/qdp-core/src/gpu/encodings/amplitude.rs`: guard the same multiplication in `encode_batch_f32` (the f64 path already routes through `Preprocessor::validate_batch`).
- Added a unit test in `preprocessing.rs` covering the overflow path.

## Test plan

- [x] `cargo test --manifest-path qdp/qdp-core/Cargo.toml --lib` — 77 passed, 0 failed
- [x] pre-commit (clippy, rustfmt, license headers) — all pass
- [x] New test `validate_batch_rejects_size_overflow` exercises the `checked_mul` failure branch

### DIFF
--- a/qdp/qdp-core/src/gpu/encodings/amplitude.rs
+++ b/qdp/qdp-core/src/gpu/encodings/amplitude.rs
@@ -482,12 +482,18 @@ impl QuantumEncoder for AmplitudeEncoder {
                 sample_size, state_len, num_qubits
             )));
         }
-        if batch_data.len() != num_samples * sample_size {
+        let expected_len = num_samples.checked_mul(sample_size).ok_or_else(|| {
+            MahoutError::InvalidInput(format!(
+                "Batch size overflow: num_samples {} * sample_size {}",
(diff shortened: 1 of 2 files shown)


ryankert01 (approved):

ryankert01:
Make sense, will let copilot double check for me.

PR #1325, opened 2026-05-18
[QDP] Add CUDA kernel tests for basis encoding

tongyu0924:
### Changes

- [x] Test

### Why

`basis.cu` had no test coverage.

### How

Add `tests/basis_encode.rs` covering:
- Single-sample encode: first, middle, last index (f64 + f32)
- Batch encode: multiple samples with different indices (f64 + f32)
- Error rejection: out-of-range index, zero `state_len`, zero `num_samples`
- Non-Linux dummy stub

## Checklist

### DIFF
--- a/qdp/qdp-kernels/tests/basis_encode.rs
+++ b/qdp/qdp-kernels/tests/basis_encode.rs
@@ -350,10 +350,10 @@ fn test_basis_encode_batch_f32_basic() {
         }
     };
 
-    let num_samples = 4usize;
-    let state_len = 4usize;
-    let num_qubits = 2u32;
-    let basis_indices: Vec<usize> = vec![0, 1, 2, 3];
+    let num_samples = 5usize;
+    let state_len = 8usize;
+    let num_qubits = 3u32;
+    let basis_indices: Vec<usize> = vec![0, 7, 3, 7, 0];
 
     let indices_d = device.htod_copy(basis_indices.clone()).unwrap();
     let mut state_d = device
--- a/qdp/qdp-kernels/tests/basis_encode.rs
+++ b/qdp/qdp-kernels/tests/basis_encode.rs
@@ -57,9 +57,9 @@ fn test_basis_encode_first_index() {
     let state_h = device.dtoh_sync_copy(&state_d).unwrap();
     assert!((state_h[0].x - 1.0).abs() < EPSILON, "state[0].re should be 1.0");
     assert!(state_h[0].y.abs() < EPSILON, "state[0].im should be 0");
-    for i in 1..state_len {
-        assert!(state_h[i].x.abs() < EPSILON, "state[{i}].re should be 0");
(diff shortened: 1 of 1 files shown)


qdp/qdp-kernels/tests/basis_encode.rs
```diff
@@ -0,0 +1,428 @@
...
+        Err(_) => {
+            println!("SKIP: No CUDA device available");
+            return;
+        }
+    };
+
+    let num_samples = 4usize;
+    let state_len = 4usize;
+    let num_qubits = 2u32;
+    let basis_indices: Vec<usize> = vec![0, 1, 2, 3];
```
ryankert01: Can we have multiple test cases and hard testcase that test the limit and edge case of the code?

400Ping:
Please fix pre-commit

400Ping (changes requested):
Thanks for adding kernel-level test coverage for basis encoding. I think this PR should be tightened in a few places:

1. Please reduce the duplicated test setup/assertion logic.
   Most tests repeat CUDA device creation, allocation, launch, copy-back, and basis-state assertions. A small helper such as `cuda_device_or_skip()`, `assert_basis_state_f64()`, `assert_basis_state_f32()`, and table-driven cases for first/middle/last would make this much easier to maintain.

2. Please add or clarify the batch-kernel edge cases.
   The batch kernels rely on `state_len == 2^num_qubits` because they compute:
   - `sample_idx = global_idx >> num_qubits`
   - `element_idx = global_idx & (state_len - 1)`

   Right now the launcher only rejects zero `num_samples` / `state_len`. If the intended contract is that callers must validate `state_len`, `num_qubits`, and basis-index bounds before calling the launcher, please document that in the test or PR description. Otherwise, please add validation/tests for:
   - batch out-of-range basis index
   - `state_len` / `num_qubits` mismatch
   - f32 batch zero `state_len`

3. Please assert the imaginary part in the f32 batch test too.
   `test_basis_encode_batch_f32_basic` currently checks only `actual.x`, while the f64 batch test checks both real and imaginary values.

rich7420:
@tongyu0924 thanks for the patch!

nits: the state buffers are created with `alloc_zeros`, so the "all other elements should be 0" assertions pass even if the kernel never writes those slots — they're zero to begin with. Since basis encoding's contract is to overwrite the *entire* state vector, pre-filling the buffer with a sentinel (e.g. `1.0`) before launch would actually verify the kernel zeroes them. Matters most for the batch kernels, where a grid-stride indexing bug could silently skip elements.

viiccwen:
@tongyu0924 any updates?

tongyu0924:
> @tongyu0924 any updates?

Sorry, I've been pretty busy lately and haven't had much time to work on this. Please feel free to let someone else take over if they'd like to continue with it.

Issue #1338, opened 2026-05-25
[Roadmap] GSOC2026 Rust code quality: QDP pipeline


0lai0:
# GSoC 2026 — QDP Pipeline Cleanup

**Scope:** `qdp-core`, `qdp-python`, `qdp-kernels`

**Pipeline today:** read data (synthetic / file / streaming Parquet) → CPU prefetch → GPU `encode_batch` → yield `QuantumTensor` to Python (`QuantumDataLoader`)

**Prerequisite:** #1310 (single-sample f32 on `QuantumEncoder` trait) — confirm merged before starting

---

## Track A — Parquet / reader f32

> Make file/Parquet sources honor `dtype=f32` end-to-end (no forced f64 widening)

| ID | Issue | What to do |
|----|-------|------------|
| A1 | [Refactor] Add FloatElem trait and DataReader<T> #1339 | Generalize `DataReader::read_batch` from `Vec<f64>` to `Vec<T>` in `reader.rs` |
| A2 | [Feature] ParquetReader: f32/f64 columns + Arrow cast #1340 | Accept f32 Parquet columns; Arrow `cast` when dtypes differ |
| A3 | [Feature] Pipeline file load respects PipelineConfig.dtype #1341 | Wire A2 into `pipeline_runner` so `dtype=f32` reaches f32 kernels |
| A4 | [Testing] Parquet f32 tests and benchmark #1342 | Fidelity tests + optional throughput benchmark |

- [x] #1339
- [x] #1340
- [x] #1341
- [x] #1342

---

## Track B — GPU memory estimate

> Early OOM detection: fail at loader build, not after 90 min of running

| ID | Issue | What to do |
|----|-------|------------|
| B1 | [Feature] Add estimate_memory() for pipeline GPU footprint #1429 | Pure function estimating CPU + GPU memory from config params |
| B2 | [Feature] Fail fast when estimated memory exceeds VRAM #1430 | Call B1 at loader construction; return actionable error |

- [x] #1429
- [ ] #1430

---

## Track C — Preprocessor Python API

> Expose `Preprocessor` helpers that already exist in Rust to Python users

| ID | Issue | What to do |
|----|-------|------------|
| C1 | [Feature] PyO3 bindings for Preprocessor #1431 | Bind `validate_input`, `calculate_l2_norm`, `validate_batch`, `calculate_batch_l2_norms` via PyO3 |
[...]

PR #1343, opened 2026-05-29
refactor(qdp-core): generic DataReader<T> with FloatElem

### LINKED ISSUE #1339: [Refactor] Add FloatElem trait and DataReader<T>
### What

Generalize `DataReader` from `Vec<f64>` to `Vec<T>` by introducing a `FloatElem` trait, so that `dtype=f32` data from file sources can stay as `f32` through the entire read path.

### Why

`DataReader::read_batch` returns `Vec<f64>` only (`qdp-core/src/reader.rs`; the comment at line 100 marks this as structural debt). Until this is fixed, `dtype=f32` file sources are silently cast to f64, preventing a true f32 pipeline.

### How

- Add `FloatElem` marker trait with `f32` / `f64` impls (or reuse existing `Dtype` mapping) in `qdp-core/src/types.rs` or a new `qdp-core/src/float_elem.rs`
- Change `DataReader` → `DataReader<T: FloatElem>`, `read_batch -> Result<(Vec<T>, usize, usize)>`
- Update `StreamingDataReader<T>` the same way
- Fix call sites that no longer compile (minimal — non-Parquet readers may still produce `f64` for now)
- No behavior change on existing `f64` paths; trait docs explain the f32 end-to-end goal

**Out of scope:** Parquet schema/cast logic, pipeline_runner wiring
**Acceptance criteria:**
- [ ] `cargo test -p qdp-core` passes on Linux
- [ ] Trait docs explain f32 end-to-end goal
- [ ] No behavior change on existing f64 paths

0lai0:
### Related Issues
Closes #1339 
Part of #1338 
### Changes

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

DataReader::read_batch was hard-coded to Vec<f64>, so dtype=f32 file sources were silently promoted to f64 before encoding. That blocks a true f32 read path into encode_batch_f32. This PR parameterizes the reader traits so f32 can be added per format without another breaking API change.

### How

- Introduce FloatElem as a sealed-by-usage marker trait (Copy + Send + Sync + 'static) with impls only for f32 and f64.
- Parameterize DataReader / StreamingDataReader with T: FloatElem, defaulting to f64 so existing impl DataReader for … and call sites (io::read_*_batch, tests, pipeline_runner) compile with no behavior change.
- Keep null handling and Arrow helpers on f64 for now (handle_float64_nulls).

Follow-up: impl DataReader<f32> for Parquet (and other formats as needed), plus pipeline wiring to consume Vec<f32> without casting.

## Checklist

### DIFF
--- a/qdp/qdp-core/src/reader.rs
+++ b/qdp/qdp-core/src/reader.rs
@@ -49,6 +49,15 @@ use arrow::array::{Array, Float64Array};
 
 use crate::error::Result;
 
+/// Scalar element type for [`DataReader`] output (`f32` or `f64` only).
+///
+/// Keeps f32 file data as `Vec<f32>` end-to-end once readers implement
+/// `DataReader<f32>`; today most readers use the default `T = f64`.
+pub trait FloatElem: Copy + Send + Sync + 'static {}
+
+impl FloatElem for f32 {}
+impl FloatElem for f64 {}
+
 /// Policy for handling null values in Float64 arrays.
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum NullHandling {
@@ -96,19 +105,16 @@ pub fn handle_float64_nulls(
 ///
 /// This interface enables zero-copy streaming where possible and maintains
 /// memory efficiency for large datasets.
-//
-// Structural debt: this trait hard-codes `Vec<f64>` in `read_batch`, so
-// `float32_pipeline=true` still forces every file-backed source to materialise
-// as f64 before casting. A future refactor should parameterise over the element
-// type (`T: FloatElem`) so f32 pipelines stay f32 end-to-end.
-pub trait DataReader {
+///
+/// Parameterised by [`FloatElem`] (`T` defaults to `f64` for existing readers).
+pub trait DataReader<T: FloatElem = f64> {
     /// Read all data from the source.
     ///
     /// Returns a tuple of:
-    /// - `Vec<f64>`: Flattened batch data (all samples concatenated)
+    /// - `Vec<T>`: Flattened batch data (all samples concatenated)
     /// - `usize`: Number of samples
     /// - `usize`: Sample size (elements per sample)
-    fn read_batch(&mut self) -> Result<(Vec<f64>, usize, usize)>;
+    fn read_batch(&mut self) -> Result<(Vec<T>, usize, usize)>;
 
     /// Get the sample size if known before reading.
     ///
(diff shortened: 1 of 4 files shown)


rich7420 (approved):
@0lai0 thanks for the patch

qdp/qdp-core/src/reader.rs
```diff
@@ -49,6 +49,15 @@ use arrow::array::{Array, Float64Array};
 
 use crate::error::Result;
 
+/// Scalar element type for [`DataReader`] output (`f32` or `f64` only).
+///
+/// Keeps f32 file data as `Vec<f32>` end-to-end once readers implement
```
rich7420, line 54: Nit: the PR description calls this "sealed-by-usage", but as written `FloatElem` is implementable by any downstream type satisfying `Copy + Send + Sync + 'static` (e.g. `u32`, or `half::f16` via a feature flag). 

Either soften the wording (e.g. `/// Currently implemented for f32 and f64.`) or actually seal it with a private supertrait:

```rust
mod sealed { pub trait Sealed {} impl Sealed for f32 {} impl Sealed for f64 {} }
pub trait FloatElem: sealed::Sealed + Copy + Send + Sync + 'static {}
```

qdp/qdp-core/tests/reader.rs
```diff
@@ -0,0 +1,115 @@
...
+        Ok(n)
+    }
+
+    fn total_rows(&self) -> usize {
+        self.total_rows
+    }
+}
+
+#[test]
+fn data_reader_default_elem_type_is_f64() {
```
rich7420, line 74: Nit: this test asserts behavior of `BatchReader<T>`'s float-literal inference, not the trait's `T = f64` default — `T` is bound here by the struct field, not the trait. The test would still pass if you changed the trait default to something else. To exercise the trait default, parameterize on the trait instead, e.g.:

```rust
fn read_default<R: DataReader>(r: &mut R) -> (Vec<f64>, usize, usize) {
    r.read_batch().unwrap()
}
```

Either rename the test or use a helper like the above.

rich7420 (commented):
some nits:

rich7420:
once you fix that , this pr will be merged

qdp/qdp-core/src/reader.rs
0lai0: Thanks @rich7420 for review. I'll update in next commit.

guan404ming (approved):

PR #1393, opened 2026-06-09
[Feature][QDP]ParquetReader generic over f32/f64 with Arrow cast

### LINKED ISSUE #1340: [Feature] ParquetReader: f32/f64 columns + Arrow cast
### What

Make `ParquetReader` generic over `FloatElem` so it can read both `Float32` and `Float64` Parquet columns, with zero-copy for same-dtype and an Arrow cast for cross-dtype reads.

### Why

`ParquetReader::new` rejects non-Float64 columns at schema check (`readers/parquet.rs` ~line 91), making f32 Parquet files completely unusable today. This blocks any f32 file source pipeline.

### How

- Introduce `ParquetReader<T: FloatElem>` accepting `List<Float32>` and `List<Float64>` (and FixedSizeList variants)
- Same dtype → zero-copy path; different dtype → `arrow::compute::cast` with safe options (document NaN/Inf/overflow behavior)
- Apply the same policy to `ParquetStreamingReader`
- New test file `qdp-core/tests/parquet_f32.rs` covering: f32 column read as f32, f64 column cast to f32, unsupported column type returns `InvalidInput` with dtype in message

**Out of scope:** pipeline_runner wiring

**Acceptance criteria:**
- [ ] `ParquetReader::<f32>` succeeds on f32 column
- [ ] f64 column read as f32: documented cast rules applied
- [ ] Unsupported column type returns `InvalidInput` error with column dtype in message
- [ ] New tests in `qdp-core/tests/parquet_f32.rs`

0lai0:
### Related Issues

Closes #1340 
Part of #1338

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

`ParquetReader::new` previously rejected any non-`Float64` column at schema validation, making f32 Parquet files unusable. This blocked f32 file-source pipelines that need to keep data as `Vec<f32>` end-to-end.

Note: higher-level helpers such as `read_parquet_batch()` still return `Vec<f64>` today. Wiring f32 through the full pipeline is tracked separately under #1338.

### How

`src/reader.rs`
- Add `Default` supertrait bound to `FloatElem` (`f32` / `f64` already satisfy it)
- Add `handle_float32_nulls`: mirrors `handle_float64_nulls`, with a zero-copy fast path when no nulls are present

`src/readers/parquet.rs`
- Introduce internal `pub(crate)` trait `ArrowPrimitive` — provides `extend_from_arrow_array` / `collect_from_arrow_array` per concrete type:
  - **same dtype** → `extend_from_slice` directly from the Arrow buffer (zero-copy)
  - **cross dtype** → `arrow::compute::cast`, then `extend_from_slice`
    - `f32 → f64`: exact, NaN preserved
    - `f64 → f32`: overflow becomes `±Inf`, NaN preserved
- `ParquetReader<T: FloatElem = f64>` and `ParquetStreamingReader<T: FloatElem = f64>`
- Schema validation now accepts:
  - `List<Float32>`
  - `List<Float64>`
  - `FixedSizeList<Float32>`
  - `FixedSizeList<Float64>`
- Default type parameter `T = f64` keeps most existing call sites unchanged (type inference from `Vec<f64>` return types still works)

`src/lib.rs`
- Re-export `handle_float32_nulls` for symmetry with `handle_float64_nulls`

`src/remote.rs`
- Add explicit `ParquetReader::<f64>::new(...)` in a test helper after genericisation (compiler could not infer `T` from `data.len()` alone)

## Checklist

### DIFF
--- a/qdp/qdp-core/src/reader.rs
+++ b/qdp/qdp-core/src/reader.rs
@@ -45,20 +45,39 @@
 //! }
 //! ```
 
-use arrow::array::{Array, Float32Array, Float64Array};
+use arrow::array::{Array, Float32Array, Float64Array, PrimitiveArray};
+use arrow::datatypes::{ArrowPrimitiveType, Float32Type, Float64Type};
 
-use crate::error::Result;
+use crate::error::{MahoutError, Result};
+
+/// Maps a Rust float primitive to its Arrow array type.
+///
+/// `pub(crate)` seals `FloatElem`: external callers cannot implement `ArrowPrimitive`
+/// and therefore cannot implement `FloatElem` for new types.
+pub(crate) trait ArrowPrimitive {
+    type ArrowType: ArrowPrimitiveType<Native = Self>;
+}
+
+impl ArrowPrimitive for f32 {
+    type ArrowType = Float32Type;
+}
+
+impl ArrowPrimitive for f64 {
+    type ArrowType = Float64Type;
+}
 
 /// Scalar element type for [`DataReader`] output (`f32` or `f64` only).
 ///
-/// Keeps f32 file data as `Vec<f32>` end-to-end once readers implement
-/// `DataReader<f32>`; today most readers use the default `T = f64`.
-pub trait FloatElem: Copy + Default + Send + Sync + 'static {}
+/// Sealed by the `pub(crate) ArrowPrimitive` supertrait — no external implementations
+/// are possible. Keeps f32 file data as `Vec<f32>` end-to-end; today most readers
+/// use the default `T = f64`.
+#[allow(private_bounds)]
+pub trait FloatElem: ArrowPrimitive + Copy + Default + Send + Sync + 'static {}
 
 impl FloatElem for f32 {}
 impl FloatElem for f64 {}
 
-/// Policy for handling null values in Float64 arrays.
+/// Policy for handling null values in float arrays.
 #[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
 pub enum NullHandling {
     /// Replace nulls with 0.0 (backward-compatible default).
@@ -68,6 +87,37 @@ pub enum NullHandling {
     Reject,
 }
 
+/// Append values from a primitive array into `output`, applying the given null policy.
+///
+/// When there are no nulls the fast path copies the underlying buffer directly.
+pub(crate) fn handle_primitive_nulls<P: ArrowPrimitiveType>(
+    output: &mut Vec<P::Native>,
+    array: &PrimitiveArray<P>,
+    null_handling: NullHandling,
+) -> Result<()>
+where
+    P::Native: Default,
+{
+    if array.null_count() == 0 {
+        output.extend_from_slice(array.values());
+    } else {
+        match null_handling {
+            NullHandling::FillZero => {
+                output.extend(array.iter().map(|opt| opt.unwrap_or_default()));
+            }
+            NullHandling::Reject => {
+                return Err(MahoutError::InvalidInput(format!(
+                    "Null value encountered in {:?} array. \
+                     Use NullHandling::FillZero to replace nulls with 0.0, \
+                     or clean the data at the source.",
+                    P::DATA_TYPE,
+                )));
+            }
+        }
+    }
+    Ok(())
+}
+
 /// Append values from a `Float64Array` into `output`, applying the given null policy.
 ///
 /// When there are no nulls the fast path copies the underlying buffer directly.
@@ -75,25 +125,8 @@ pub fn handle_float64_nulls(
     output: &mut Vec<f64>,
     float_array: &Float64Array,
     null_handling: NullHandling,
-) -> crate::error::Result<()> {
-    if float_array.null_count() == 0 {
-        output.extend_from_slice(float_array.values());
-    } else {
-        match null_handling {
-            NullHandling::FillZero => {
-                output.extend(float_array.iter().map(|opt| opt.unwrap_or(0.0)));
-            }
-            NullHandling::Reject => {
-                return Err(crate::error::MahoutError::InvalidInput(
-                    "Null value encountered in Float64Array. \
-                     Use NullHandling::FillZero to replace nulls with 0.0, \
-                     or clean the data at the source."
-                        .to_string(),
-                ));
-            }
-        }
-    }
-    Ok(())
+) -> Result<()> {
+    handle_primitive_nulls::<Float64Type>(output, float_array, null_handling)
 }
 
 /// Append values from a `Float32Array` into `output`, applying the given null policy.
@@ -103,25 +136,8 @@ pub fn handle_float32_nulls(
     output: &mut Vec<f32>,
     float_array: &Float32Array,
     null_handling: NullHandling,
-) -> crate::error::Result<()> {
-    if float_array.null_count() == 0 {
-        output.extend_from_slice(float_array.values());
-    } else {
-        match null_handling {
-            NullHandling::FillZero => {
-                output.extend(float_array.iter().map(|opt| opt.unwrap_or(0.0)));
-            }
-            NullHandling::Reject => {
-                return Err(crate::error::MahoutError::InvalidInput(
-                    "Null value encountered in Float32Array. \
-                     Use NullHandling::FillZero to replace nulls with 0.0, \
-                     or clean the data at the source."
-                        .to_string(),
-                ));
-            }
-        }
-    }
-    Ok(())
+) -> Result<()> {
+    handle_primitive_nulls::<Float32Type>(output, float_array, null_handling)
 }
 
 /// Generic data reader interface for batch quantum data.
--- a/qdp/qdp-core/src/reader.rs
+++ b/qdp/qdp-core/src/reader.rs
@@ -45,7 +45,7 @@
 //! }
 //! ```
 
-use arrow::array::{Array, Float64Array};
+use arrow::array::{Array, Float32Array, Float64Array};
 
 use crate::error::Result;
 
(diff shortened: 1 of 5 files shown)


qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -17,24 +17,180 @@
...
+                return Err(MahoutError::InvalidInput(format!(
+                    "Expected Float32 or Float64 values, got {other:?}"
+                )));
+            }
+        }
+        Ok(())
+    }
+}
+
+impl ArrowPrimitive for f32 {
```
rich7420: These two `ArrowPrimitive` impls are basically mirror images (same-dtype downcast + cross-dtype cast, identical wiring). Could collapse into one generic helper + an associated type, e.g.:
```rust
trait ArrowPrimitive: FloatElem { type Arrow: ArrowPrimitiveType<Native = Self>; /* default method */ }
impl ArrowPrimitive for f64 { type Arrow = Float64Type; }
impl ArrowPrimitive for f32 { type Arrow = Float32Type; }
```
with a single `extend_floats::<P>(...)`. Cuts ~60 lines and keeps the zero-copy fast path. Not blocking, just nice-to-have.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -17,24 +17,180 @@
...
+            DataType::Float64 => {
+                let arr = array
+                    .as_any()
+                    .downcast_ref::<Float64Array>()
+                    .ok_or_else(|| MahoutError::Io("Float64 downcast failed".to_string()))?;
+                handle_float64_nulls(output, arr, null_handling)?;
+            }
+            DataType::Float32 => {
+                let casted = compute::cast(array, &DataType::Float64)
+                    .map_err(|e| MahoutError::Io(format!("Arrow cast f32→f64: {e}")))?;
```
rich7420: Cast failures map to `MahoutError::Io`, but a dtype conversion isn't really I/O. `InvalidInput` fits better and matches the `other =>` arm right below. Same for the f64→f32 one on L112.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -17,24 +17,180 @@
...
+    ) -> Result<()> {
+        match array.data_type() {
+            DataType::Float32 => {
+                let arr = array
+                    .as_any()
+                    .downcast_ref::<Float32Array>()
+                    .ok_or_else(|| MahoutError::Io("Float32 downcast failed".to_string()))?;
+                handle_float32_nulls(output, arr, null_handling)?;
+            }
+            DataType::Float64 => {
```
rich7420: Heads up: `ParquetReader::<f32>` will happily take an f64 file and silently downcast — out-of-range → ±Inf with no warning/error. That's an easy way to get garbage into encoding without noticing. Is a strict mode (or at least a `log::warn`) worth it? Also wondering if this should tie into the existing `Precision`/`Dtype` instead of an ad-hoc cast. Genuine question, fine to punt to #1338.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -180,8 +305,11 @@ impl DataReader for ParquetReader {
                             all_data.reserve(current_size * self.total_rows);
                         }
 
-                        handle_float64_nulls(&mut all_data, float_array, self.null_handling)?;
-
+                        T::extend_from_arrow_array(
```
rich7420: This calls `extend_from_arrow_array` once per row, so on the cross-dtype path `compute::cast` runs (and allocates) per row. The FixedSizeList arm already casts the whole `values()` buffer once — could do the same here: validate sizes in the loop, then extend once from `list_array.values()`. Big win for wide/large files.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -439,24 +541,16 @@ impl StreamingDataReader for ParquetStreamingReader {
                                 continue;
                             }
 
-                            let mut batch_values = Vec::new();
+                            let mut batch_values: Vec<T> = Vec::new();
                             let mut current_sample_size = None;
                             for i in 0..list_array.len() {
```
rich7420: Same per-row cast pattern as the non-streaming path (L292) — one `compute::cast` alloc per row on cross-dtype reads. Cast the batch `values()` once instead.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -353,13 +455,13 @@ impl ParquetStreamingReader {
...
-impl DataReader for ParquetStreamingReader {
-    fn read_batch(&mut self) -> Result<(Vec<f64>, usize, usize)> {
+impl<T: FloatElem + ArrowPrimitive> DataReader<T> for ParquetStreamingReader<T> {
+    fn read_batch(&mut self) -> Result<(Vec<T>, usize, usize)> {
         let mut all_data = Vec::new();
         let mut num_samples = 0;
 
         loop {
-            let mut buffer = vec![0.0; 1024 * 1024]; // 1M elements buffer
+            let mut buffer = vec![T::default(); 1024 * 1024];
```
rich7420: This 1M-element buffer gets allocated + zero-filled on every loop iteration (pre-existing, but now `T::default()`). It's fully overwritten up to `written`, so it can be hoisted out of the loop and reused. Minor.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -17,24 +17,180 @@
...
+        }
+        Ok(())
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Shared schema validator
+// ---------------------------------------------------------------------------
+
+fn validate_float_list_schema(field: &arrow::datatypes::Field) -> Result<()> {
```
rich7420, line 42: The List and FixedSizeList arms run the identical `matches!(.., Float32 | Float64)` check. A tiny `fn is_supported_float(dt: &DataType) -> bool` reused here + in the scalar validator below would keep the accepted set in one place. nit.

rich7420 (commented):
@0lai0 thanks for the patch!

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -17,24 +17,180 @@
...
+pub struct ParquetReader<T: FloatElem = f64> {
     reader: Option<parquet::arrow::arrow_reader::ParquetRecordBatchReader>,
     sample_size: Option<usize>,
     total_rows: usize,
     null_handling: NullHandling,
+    _phantom: PhantomData<T>,
 }
 
-impl ParquetReader {
+#[allow(private_bounds)]
```
rich7420: `#[allow(private_bounds)]` on a public `new` because `ArrowPrimitive` is `pub(crate)` — downstream users can't name the bound and only see it via a compile error. Since it's effectively a sealed f32/f64 trait, putting the bound on the public `FloatElem` (or a sealed public trait) would drop the `allow`. Minor design nit.

qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -644,13 +713,17 @@ mod tests {
...
-        let result = ParquetReader::new(file.path(), None, NullHandling::FillZero);
+        let result = ParquetReader::<f64>::new(file.path(), None, NullHandling::FillZero);
         assert!(result.is_err());
         let err_msg = match result {
             Err(e) => e.to_string(),
             Ok(_) => panic!(),
         };
-        assert!(err_msg.contains("Expected List<Float64> or FixedSizeList<Float64> column"));
+        assert!(
+            err_msg.contains("Expected List")
```
rich7420: This assertion got pretty loose 
cuz the `|| contains("Float32/Float64")` substring shows up in almost every error message in this file, so the test would pass even if the wrong validation branch fired. Maybe assert on the actual `Int32`/`List<` dtype in the message instead?

qdp/qdp-core/src/reader.rs
```diff
@@ -96,6 +96,34 @@ pub fn handle_float64_nulls(
     Ok(())
 }
 
+/// Append values from a `Float32Array` into `output`, applying the given null policy.
+///
+/// When there are no nulls the fast path copies the underlying buffer directly.
+pub fn handle_float32_nulls(
```
rich7420, line 135: `handle_float32_nulls` is a verbatim copy of `handle_float64_nulls` (only the type + the `32`/`64` literal differ). Could be one generic `handle_primitive_nulls<T: ArrowPrimitiveType>` with `T::Native: Default` (use `unwrap_or_default()`), keeping these two as thin wrappers for the public API. 

This one is optional.

qdp/qdp-core/src/readers/parquet.rs
0lai0: Thank @rich7420 for review. I'll update in next commit.

rich7420:
@0lai0 thanks for your hard work and one follow-up after we merge 🙏

Heads up: a **null outer list row** makes the whole read fail. Reading a `List<Float64>` of `[[1,2], null, [3,4]]` returns:

```
InvalidInput("Inconsistent sample sizes: expected 2, got 0")
```

`FillZero` doesn't help — a null row's `value_length(i)` is 0, which trips the consistency check (and if it's row 0, `sample_size` locks to 0).

- It's **pre-existing** (old per-row code did the same), **not introduced by this PR**
- But real-world nullable List columns hit it, and there's no null-outer-row test
- Related: `written / self.sample_size.unwrap_or(1)` — `unwrap_or(1)` only guards `None`, not `Some(0)`. Unreachable in practice (the check errors first), but a cheap `Some(0)` guard wouldn't hurt

rich7420 (approved):
LGTM

PR #1402, opened 2026-06-13
[Bug][QDP] Fix List<T> readers failing on nullable outer rows

### LINKED ISSUE #1401: [Bug] List<T> reader fails with "Inconsistent sample sizes" on nullable outer rows
## What

When reading a nullable `List<T>` column, a `null` **outer** row makes the entire read fail with an `InvalidInput` consistency error, instead of being treated as a missing sample.

The List sample-size validation calls Arrow's `ListArray::value_length(i)` without an `is_null(i)` guard. For a `null` outer row, `value_length(i)` returns `0`; the reader treats this as a length-0 list, which trips the `sample_size` consistency check. After a non-null row it fails with `expected N, got 0`; if the `null` row is row `0`, it seeds `sample_size` to `0` and breaks every subsequent row. `FillZero` (`NullHandling::FillZero`) does not help — the failure is at the *list-length* level, before any value-filling runs.

Reproduce by reading a `List<Float64>` column with the values `[[1, 2], null, [3, 4]]`:

```
InvalidInput("Inconsistent sample sizes: expected 2, got 0")
```

## Why

- **Real-world impact.** Nullable `List` columns with null rows are common in practice, so this is hit by ordinary data, not just edge cases.
- **No coverage.** There is currently no test for a null outer row.
- **Pre-existing, but worth fixing.** This is not introduced by #1393 — the previous per-row code behaved the same way. The PR is fine to merge; this is a follow-up.
- **Affects multiple readers.** Three readers share the same gap (see How), so the bug surfaces across both batch and streaming paths and the IPC path.

## How

The fix is to add an `is_null(i)` guard to the List sample-size validation so a `null` row is not counted as a length-0 list (including the row-0 seed). Affected code, all in `qdp/qdp-core/src/readers/`:

- **`ParquetReader`** — `parquet.rs:294–306`. Loop over all rows calls `value_length(i)` at line 295 with no `is_null(i)` check; a null row 0 locks `sample_size` to `0`.
- **`ParquetStreamingReader`** — `parquet.rs`. Row-0 seed at line 564 (no null check) and validation loop at lines 568–573, same gap.
- **`ArrowIPCReader`** — `arrow_ipc.rs:154`. Uses `list_array.value(i).len()` with no `is_null(i)` guard; same problem.

The intended semantics of a null row (skip vs. fill per null policy) should be decided first, since it determines how the guard behaves.

### Tasks

- [ ] Add the `is_null(i)` guard to all three readers (`ParquetReader`, `ParquetStreamingReader`, `ArrowIPCReader`).
- [ ] Add a regression test for a nullable `List` column with a `null` outer row, covering both null-in-the-middle and null-at-row-0.
- [ ] (Optional) Harden `written / self.sample_size.unwrap_or(1)` at `parquet.rs:490` against `Some(0)`: `unwrap_or(1)` guards `None` but not `Some(0)`. Unreachable today (the consistency check errors first), but a cheap guard keeps the division safe against a future `sample_size == 0` path.

---

Follow-up to #1393.

0lai0:
### Related Issues

Closes #1401 
Part of #1338 
### Changes

- [x] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

Reading a nullable `List<T>` column with a null outer row (e.g. `[[1, 2], null, [3, 4]]`) fails with:
`InvalidInput("Inconsistent sample sizes: expected 2, got 0")`
Root cause: sample-size validation calls `ListArray::value_length(i)` without checking `is_null(i)`. For a null outer row, Arrow returns 0, which the reader treats as a length-0 list.

If a non-null row was seen first → validation fails with `expected N, got 0`
If the null row is row 0 → sample_size is seeded to 0 and corrupts all subsequent rows
This is pre-existing (not introduced by #1393) and affects three readers:
- ParquetReader
- ParquetStreamingReader
- ArrowIPCReader

`NullHandling::FillZero` does not help because the failure happens at list-length validation, before value filling runs.

### How

 **Null outer row semantics**

| Policy | Behavior |
|--------|----------|
| `Reject` | Return `InvalidInput` with a clear message |
| `FillZero` | Fill `sample_size` zeros once `sample_size` is known |
| `sample_size` unknown | Skip the null row (no zeros, not counted in `num_samples`) — e.g. leading all-null batch before any non-null row establishes size |

**Code changes**

1. **`ParquetReader`** (`parquet.rs`)
   - Add `is_null(i)` guard in sample-size validation loop
   - Split fast path (no null outer rows) vs slow path (per-row null handling per `NullHandling`)

2. **`ParquetStreamingReader`** (`parquet.rs`)
   - Seed `sample_size` from first non-null row instead of row 0
   - Skip all-null batches when size is unknown
   - Same null-handling policy as batch reader
   - Defensive `.max(1)` on sample-size divisor in `read_batch`

3. **`ArrowIPCReader`** (`arrow_ipc.rs`)
   - Two-phase List handling
   - Phase 1: validate size from non-null rows only
   - Phase 2: collect data with the same null policy
## Checklist

### DIFF
--- a/qdp/qdp-core/src/readers/arrow_ipc.rs
+++ b/qdp/qdp-core/src/readers/arrow_ipc.rs
@@ -150,16 +150,12 @@ impl DataReader for ArrowIPCReader {
                             MahoutError::Io("Failed to downcast to ListArray".to_string())
                         })?;
 
+                    // Phase 1: find sample_size from non-null rows and validate consistency.
                     for i in 0..list_array.len() {
-                        let value_array = list_array.value(i);
-                        let float_array = value_array
-                            .as_any()
-                            .downcast_ref::<Float64Array>()
-                            .ok_or_else(|| {
-                                MahoutError::Io("List values must be Float64".to_string())
-                            })?;
-
-                        let current_size = float_array.len();
+                        if list_array.is_null(i) {
+                            continue;
+                        }
+                        let current_size = list_array.value_length(i) as usize;
 
                         if let Some(expected) = sample_size {
                             if current_size != expected {
@@ -180,10 +176,53 @@ impl DataReader for ArrowIPCReader {
                                 })?;
                             all_data.reserve(new_capacity);
                         }
+                    }
 
+                    // Phase 2: collect data, handling null outer rows per NullHandling policy.
+                    if list_array.null_count() == 0 {
+                        let values = list_array.values();
+                        let float_array = values
+                            .as_any()
+                            .downcast_ref::<Float64Array>()
+                            .ok_or_else(|| MahoutError::Io("Values must be Float64".to_string()))?;
                         handle_float64_nulls(&mut all_data, float_array, self.null_handling)?;
-
-                        num_samples += 1;
+                        num_samples += list_array.len();
+                    } else {
+                        for i in 0..list_array.len() {
+                            if list_array.is_null(i) {
+                                match self.null_handling {
+                                    NullHandling::Reject => {
+                                        return Err(MahoutError::InvalidInput(
+                                            "Null outer row in List column. Use \
+                                             NullHandling::FillZero to replace with zeros, \
+                                             or clean the data at the source."
+                                                .to_string(),
+                                        ));
+                                    }
+                                    NullHandling::FillZero => {
+                                        if let Some(ss) = sample_size {
+                                            all_data.extend(std::iter::repeat_n(0.0_f64, ss));
+                                            num_samples += 1;
+                                        }
+                                        // sample_size unknown: skip this null row without counting it.
+                                    }
+                                }
+                            } else {
+                                let value_array = list_array.value(i);
+                                let float_array = value_array
+                                    .as_any()
+                                    .downcast_ref::<Float64Array>()
+                                    .ok_or_else(|| {
+                                        MahoutError::Io("List values must be Float64".to_string())
+                                    })?;
+                                handle_float64_nulls(
+                                    &mut all_data,
+                                    float_array,
+                                    self.null_handling,
+                                )?;
+                                num_samples += 1;
+                            }
+                        }
                     }
                 }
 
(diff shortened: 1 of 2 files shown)


qdp/qdp-core/src/readers/parquet.rs
```diff
@@ -561,10 +600,25 @@ impl<T: FloatElem> StreamingDataReader<T> for ParquetStreamingReader<T> {
...
+                            // Null outer rows return value_length 0 and must be skipped.
+                            let first_non_null =
+                                (0..list_array.len()).find(|&i| !list_array.is_null(i));
+                            let current_sample_size = match first_non_null {
+                                Some(i) => list_array.value_length(i) as usize,
+                                None => match self.sample_size {
+                                    // All rows null but sample_size known from an earlier batch.
+                                    Some(ss) => ss,
+                                    // All rows null and sample_size unknown: skip batch.
+                                    None => continue,
```
rich7420: **nit:** In `Reject` mode, when the first batch is entirely null and `sample_size` is still unknown, `None => continue` returns before `null_handling` is consulted — so this batch is silently skipped instead of rejected. The batch `ParquetReader` errors on the same input, so the two readers diverge here. It's also the one `Reject` path without test coverage. Suggest either erroring on `Reject` here too, or documenting the difference explicitly.

rich7420:
@0lai0 thanks for the patch!

qdp/qdp-core/src/readers/arrow_ipc.rs
```diff
@@ -150,16 +150,12 @@ impl DataReader for ArrowIPCReader {
...
-                        let value_array = list_array.value(i);
-                        let float_array = value_array
-                            .as_any()
-                            .downcast_ref::<Float64Array>()
-                            .ok_or_else(|| {
-                                MahoutError::Io("List values must be Float64".to_string())
-                            })?;
-
-                        let current_size = float_array.len();
+                        if list_array.is_null(i) {
```
ryankert01, line 155: I want to ask about the context of a possible null might exists in a stream of data? We often assume the data user sent are correct. We need a boundry of what we should check and what we shouldn't check for user. (to strike a balance between performance and early stop when handling error data) 

this pr already do well actually~

ryankert01 (approved):
lgtm!

qdp/qdp-core/src/readers/arrow_ipc.rs
0lai0: Thanks @ryankert01. 
Good point! Arrow's List field defaults to nullable=true in most export tools (pandas, polars, Parquet writers), so null outer rows can appear even without user error. 
The NullHandling policy lets the caller decide strictness. Reject for tight pipelines, FillZero for tolerant ones.

guan404ming (approved):

PR #1407, opened 2026-06-18
[Feature][QDP] Pipeline file load respects PipelineConfig.dtype

### LINKED ISSUE #1341: [Feature] Pipeline file load respects PipelineConfig.dtype
### What

Wire `PipelineConfig.dtype` through `read_file_by_extension` and the producer layer so that f32-configured pipelines actually reach the f32 encoder kernel when loading from file sources.

### Why

`read_file_by_extension` always returns `Vec<f64>` (`pipeline_runner.rs`). `InMemoryProducer` and `StreamingProducer` populate `BatchData::F64` even when `config.dtype == Float32`, so the f32 encoder is never reached from file sources. The user's dtype choice is silently ignored.

### How

- `read_file_by_extension` (and streaming path) selects the `f32` vs `f64` reader based on `PipelineConfig.dtype`
- `InMemoryProducer` / `StreamingProducer` use the matching `BatchData` variant
- Python API (`QuantumDataLoader.source_file`) unchanged at the surface
- Integration test: f32 Parquet + `dtype=f32` config → `encode_batch_f32_for_pipeline` path is exercised

**Acceptance criteria:**
- [ ] Integration test: f32 Parquet + `dtype=f32` config → `encode_batch_f32_for_pipeline` path is used
- [ ] Existing f64 file tests still pass
- [ ] No silent f64 kernel when user requested f32 file pipeline

0lai0:
### Related Issues

Closes #1341 
Part of #1338

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

`PipelineConfig.dtype` is silently ignored when loading from files. `read_file_by_extension` always returned `Vec<f64>`, so `BatchData::F64` was always produced and `encode_batch_f32_for_pipeline` was never reached from a file source. `ParquetReader<T> / ParquetStreamingReader<T> `already had f32/f64 support from #1393 (issue #1340) — this PR just wires the dtype through.

### How

- InMemoryProducer and StreamingProducer are now generic over `T: FloatElem (default f64)`. A pub(crate) ToBatchData trait handles wrapping/unwrapping BatchData inside produce().
- read_file_by_extension takes dtype: Precision and returns BatchData. Parquet dispatches to ParquetReader::<f32|f64> directly. Other formats (Arrow IPC, NumPy, PyTorch, TensorFlow) read as f64 and cast with a narrowing warning — native f32 readers for those formats are a follow-up.
- new_from_file_streaming's reader setup was extracted into build_streaming_producer::<T> to avoid duplicating ~40 lines.

## Checklist

### DIFF
--- a/qdp/qdp-core/src/pipeline_runner.rs
+++ b/qdp/qdp-core/src/pipeline_runner.rs
@@ -428,6 +428,26 @@ fn path_extension_lower(path: &Path) -> Option<String> {
         .map(|s| s.to_lowercase())
 }
 
+/// Largest integer exactly representable by an f32 (its mantissa is 24 bits): `2^24`.
+/// Basis indices above this would silently change value when narrowed to f32.
+const MAX_EXACT_F32_INT: f64 = (1u64 << 24) as f64;
+
+/// Reject an explicit f32 request for a basis file whose indices exceed f32's exact
+/// integer range. Basis values are integer state indices, not floats, so narrowing
+/// `16_777_217` to `16_777_216` would encode the wrong state. Rather than silently
+/// corrupt (or silently widen back to f64), surface the conflict to the caller.
+fn reject_basis_f32_out_of_range(data: &BatchData) -> Result<()> {
+    if let BatchData::F64(v) = data
+        && let Some(&bad) = v.iter().find(|&&x| x > MAX_EXACT_F32_INT)
+    {
+        return Err(MahoutError::InvalidInput(format!(
+            "basis index {bad:.0} exceeds f32's exact integer range ({MAX_EXACT_F32_INT:.0}); \
+             narrowing to f32 would encode the wrong state. Use dtype='float64' for this basis file."
+        )));
+    }
+    Ok(())
+}
+
 /// f64→f32 narrowing cast: values outside f32 range silently become ±Inf.
 fn cast_f64_to_batch_data(
     data: Vec<f64>,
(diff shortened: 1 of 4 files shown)


0lai0:
cc @rich7420

ryankert01 (changes requested):
overall lg. The dtype wiring looks like the right direction overall, but `basis` should not go through a lossy f32 representation before index validation.

I also left a smaller inline note about the Python loader defaulting file pipelines to f32. If that default is intentional, it probably needs to be explicit in the public API/docs.

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -389,28 +428,67 @@ fn path_extension_lower(path: &Path) -> Option<String> {
...
+) -> Result<(BatchData, usize, usize)> {
+    use crate::reader::DataReader;
     let ext_lower = path_extension_lower(path);
     let ext = ext_lower.as_deref();
     match ext {
         Some("parquet") => {
-            use crate::reader::DataReader;
-            let mut reader = crate::readers::ParquetReader::new(path, None, null_handling)?;
-            reader.read_batch()
+            if matches!(dtype, Precision::Float32) {
```
ryankert01: One concern here: this f32 path also covers `basis`, where the values are indices rather than ordinary float features. A valid index like `16_777_217` for 25 qubits is not exactly representable as `f32`, so it becomes `16_777_216`; the basis f32 encoder then validates the rounded value and encodes the wrong state. Can we keep basis file input in f64/integer form, or reject f32 basis input once indices exceed the exact f32 range?

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -389,28 +428,67 @@ fn path_extension_lower(path: &Path) -> Option<String> {
...
-/// Unsupported or missing extension returns Err with message listing supported formats.
+/// f64→f32 narrowing cast: values outside f32 range silently become ±Inf.
+fn cast_f64_to_batch_data(
+    data: Vec<f64>,
+    n: usize,
+    s: usize,
+    dtype: Precision,
+    fmt: &str,
+) -> (BatchData, usize, usize) {
+    if matches!(dtype, Precision::Float32) {
```
ryankert01, line 459: This cast now affects `QuantumDataLoader.source_file()` too, because the Python loader factories build file pipelines with `Dtype::Float32` and there is no public loader dtype option. That makes f64 file input narrow by default for f32-capable encodings. Is that the behavior we want? If so, I think it should be called out in the loader docs, and ideally exposed as an explicit loader setting.

qdp/qdp-core/src/pipeline_runner.rs
rich7420: +1

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -1365,4 +1474,190 @@ mod tests {
...
+
+        #[test]
+        fn test_read_file_by_extension_f32_parquet_returns_f32_batch_data() {
+            let path = temp_parquet_path("f32");
+            write_f32_parquet(&path);
+            let result = read_file_by_extension(&path, NullHandling::FillZero, Precision::Float32);
+            let _ = fs::remove_file(&path);
+            let (batch_data, num_samples, sample_size) = result.unwrap();
+            assert!(
+                matches!(batch_data, BatchData::F32(_)),
```
rich7420: nit (non-blocking): these new tests only check the `BatchData` *variant*, not the values — so a batch that comes back as `F32` but with zeroed/garbled contents would still pass green. Might be worth reading one element back (e.g. `assert` the first value ≈ `0.25`) so it's a real check, like `test_synthetic_producer_f32_amplitude` does. Same gap means the `cast_f64_to_batch_data` path (arrow/npy/pt/pb → f32) and the f64-parquet→f32 cross-cast don't have coverage yet.

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -492,26 +628,42 @@ impl PipelineIterator {
...
-            sample_size,
-            batch_size: config.batch_size,
-            num_qubits: config.num_qubits as usize,
-            batches_yielded: 0,
-            batch_limit,
-        };
         let prefetch_depth = config.prefetch_depth;
-        let (rx, recycle_tx, _producer_handle) = spawn_producer(producer, prefetch_depth)?;
+        let (rx, recycle_tx, _producer_handle) = match batch_data {
+            BatchData::F32(data) => spawn_producer(
```
rich7420: nit (totally optional): these two arms are byte-for-byte identical apart from the `<f32>`/`<f64>` type param, and the streaming path already has `build_streaming_producer<T>` — a little `build_inmemory_producer<T>` helper here would kill the copy-paste and mirror that pattern. Just caught my eye while reading.

rich7420:
btw, @0lai0  thanks for your patch!

qdp/qdp-core/src/pipeline_runner.rs
0lai0: Thanks @ryankert01 and @rich7420 for review. Sure, I'll keep basis file input in f64 and reject an explicit dtype=f32 request in next commit.

qdp/qdp-core/src/pipeline_runner.rs
0lai0: Great catch. I've updated the implementation so that file loads now default to f64 for lossless precision, while Float32 is now opt-in via a new public .dtype() loader method, where parse_dtype(None) defaults to Float64. Thanks for review.

qdp/qdp-core/src/pipeline_runner.rs
0lai0: Tests now assert the actual values (first element ≈ 0.25), not just the variant. Thanks for review.

qdp/qdp-core/src/pipeline_runner.rs
0lai0: I updated `extracted::build_inmemory_producer<T>` so that the two `from_file` arms are now consolidated into a single generic call, aligning it with `build_streaming_producer<T>`.

ryankert01 (approved):

rich7420 (approved):

rich7420:
@ryankert01 thanks for the review , @0lai0 thanks for the patch!

PR #1413, opened 2026-06-26
ci(qdp): build the extension once and reuse it across the test matrix

ryankert01:
### Changes

- [x] CI/CD pipeline

### Why

The Python Testing workflow rebuilt the QDP extension independently in every
matrix leg (3.10 / 3.11 / 3.12), compiling the same Rust code three times. The
legs run in parallel, so this duplicated CI compute (billed minutes) without
buying any wall-clock speedup.

### How

- Add a `build` job that compiles the CPU / no-CUDA stub wheels for all
  supported interpreters **once** and uploads them as an artifact.
- Rewire the `test` matrix to `needs: build`: it downloads the artifact and
  installs the wheel matching each leg's interpreter instead of recompiling.
- `QDP_NO_CUDA=1` makes the stub build explicit, matching the prior per-leg
  behaviour (which fell back to stubs because `nvcc` is absent on the runner —
  see `qdp/qdp-core/build.rs` and `qdp/qdp-kernels/build.rs`).

Net effect: the heavy non-pyo3 crates compile a single time instead of once per
leg. The `rust-check` gate, the test environment setup, and the `pytest`
invocation are otherwise unchanged. The CUDA wheel build in `wheel-build.yml`
(used for distribution) is intentionally left as-is.

Job graph: `rust-check → build → test (matrix)`.

## Checklist

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -57,8 +57,44 @@ jobs:
           QDP_NO_CUDA: "1"
         run: cargo check --workspace --tests
 
-  test:
+  # Build the QDP extension once (CPU / no-CUDA stub build, all supported
+  # Python versions) and hand the wheels to the test matrix below, so the
+  # extension is compiled a single time instead of once per matrix leg.
+  build:
     needs: rust-check
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


ryankert01:
Not effective

PR #1415, opened 2026-06-26
refactor(qdp): rename validate_tensor to validate_tensor_cpu and drop redundant check

### LINKED ISSUE #907: refactor: validate_tensor()
Good catch, got it!

 In `validate_tensor` has a part of it redundant as it rechecks `is_pytorch_tensor(tensor)`, but it can also catch other hardware tensor like MPS, XLA, HPU, etc. if any. 
We can refactor it and remove `is_pytorch_tensor(tensor)` as a followup. and the function name probably should be renamed to `validate_tensor_cpu` or something.

_Originally posted by @ryankert01 in https://github.com/apache/mahout/pull/881#discussion_r2708210622_

ryankert01:
### Related Issues

Closes #907

### Changes

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

Follow-up to the review thread on #881. `validate_tensor()` re-checked
`is_pytorch_tensor(tensor)` even though its **only** caller already
guarantees the object is a `torch.Tensor`:

- `validate_tensor` has a single call site: `QdpEngine::encode_from_pytorch`
  (`engine.rs:209`).
- `encode_from_pytorch` is only reached from the `encode()` dispatcher
  (`engine.rs:125`), and that branch is already gated by
  `is_pytorch_tensor(data)?` at `engine.rs:119`.

So the inner check was dead weight on the real execution path. The
function's actual job is validating *device placement* — the
`device.type != "cpu"` guard, which is what already rejects non-CPU
backends (CUDA/MPS/XLA/HPU). The old name `validate_tensor` didn't convey
that, hence the rename suggested in the issue.

### How

- Removed the redundant `is_pytorch_tensor` check (and its now-dead
  `"Object is not a PyTorch Tensor"` error arm) from the helper.
- Renamed `validate_tensor` → `validate_tensor_cpu`, with a doc comment
  stating the caller-side contract (caller has already confirmed it is a
  `torch.Tensor`) and that it rejects non-CPU backends.
- Updated the import and call site in `engine.rs`.

No behavior change on any supported path. Verified with
`cargo check` (`--cfg qdp_no_cuda`) and `cargo fmt --check`.

## Checklist

- [ ] Added or updated unit tests for all changes — N/A, behavior-preserving rename; existing tests still cover the CPU-tensor path
- [x] Added or updated documentation for all changes — doc comment on `validate_tensor_cpu`

### DIFF
--- a/qdp/qdp-python/src/engine.rs
+++ b/qdp/qdp-python/src/engine.rs
@@ -16,7 +16,7 @@
 
 use crate::pytorch::{
     extract_cuda_tensor_info, get_torch_cuda_stream_ptr, is_cuda_tensor, is_pytorch_tensor,
-    validate_cuda_tensor_for_encoding, validate_shape, validate_tensor,
+    validate_cuda_tensor_for_encoding, validate_shape, validate_tensor_cpu,
 };
 use crate::tensor::QuantumTensor;
 use numpy::{PyReadonlyArray1, PyReadonlyArray2, PyUntypedArrayMethods};
@@ -206,7 +206,7 @@ impl QdpEngine {
         }
 
         // CPU tensor path
-        validate_tensor(data)?;
+        validate_tensor_cpu(data)?;
         // PERF: Avoid Tensor -> Python list -> Vec deep copies.
         //
         // For CPU tensors, `tensor.detach().numpy()` returns a NumPy view that shares the same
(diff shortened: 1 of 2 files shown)


viiccwen (approved):
LGTM!

viiccwen:
cc @400Ping

400Ping (approved):

PR #1416, opened 2026-06-26
[MAHOUT-1414][QDP] Expose native cuda_available() and gate GPU tests on it

### LINKED ISSUE #1414: [QDP] Unify GPU-availability detection in tests; stub CUDA path should raise, not abort
### Context

#1321 lets `qdp-core` / `_qdp` build and import **without the CUDA toolkit** (it links stub CUDA Runtime symbols when `nvcc` is absent). This is the right capability, but it breaks a long-standing assumption baked into the test suite: **"`_qdp` is importable" ⟺ "a GPU is usable."** That assumption was true before (no toolkit → link failure → `_qdp` unimportable), so a single import check was enough to gate GPU tests.

After #1321, on a GPU-less runner `_qdp` imports fine (stubs), so the `@pytest.mark.gpu` tests ran against the stub engine and **aborted the pytest workers** (`Fatal Python error: Aborted`). #1321 patched the immediate breakage in `testing/conftest.py` (skip `@pytest.mark.gpu` when `torch.cuda.is_available()` is False) and marked one loader test, which is enough to make CI green — but the underlying detection strategy is fragmented and has a real gap. This issue tracks doing it properly.

### Problems

1. **Fragmented "needs GPU/QDP" detection — no single source of truth.** The suite currently expresses this three different ways:
   - `@pytest.mark.gpu` (~95 sites), now auto-skipped by conftest
   - inline `if not torch.cuda.is_available(): pytest.skip(...)` (~65 sites, e.g. `testing/qdp/test_bindings.py`)
   - import-only guards `requires_qdp` / `_qdp_available()` / `_loader_available()` (~110 sites) that still conflate "importable" with "GPU present"

2. **`torch.cuda.is_available()` is a proxy, not the truth.** It reports torch's view of CUDA, not whether `_qdp` itself has a real runtime. Gap on #1321's *headline* scenario — a box with a GPU + PyTorch but **no toolkit**: `_qdp` is a stub build, yet `torch.cuda.is_available()` is `True`, so the GPU tests would run against the stub and abort again. `_qdp` exposes no "stub vs real CUDA" signal today.

3. **The stub path aborts instead of raising.** #1321's description promises that calling a CUDA function on a no-toolkit build "surfaces a clean runtime error instead of failing at link time," but in practice the path SIGABRTs the process. `test_synthetic_loader_batch_count` even has a `try/except RuntimeError` that *expects* the clean behaviour. If the stub raised cleanly, the failures would be ordinary errors (no worker death) and far less test-side gating would be needed.

### Proposed design

- Expose `_qdp`'s real-CUDA state to Python — e.g. a build-time flag / `_qdp.cuda_available()` distinguishing a stub build (and ideally probing for a device).
- Add `qumat_qdp.is_cuda_available()` mirroring the existing `is_triton_amd_available()`, as the **single** source of truth.
- Route the conftest auto-skip and the import-only guards through it; collapse the ~65 redundant inline `torch.cuda.is_available()` checks.
- Make the stub CUDA path raise a clean `RuntimeError` rather than abort, fulfilling #1321's stated contract.

### Not doing this in #1321 on purpose

#1321 is a contained cudart-linkage fix and is green. The above touches the Rust extension and 100+ test sites, so it belongs in its own PR.

_Refs: #1321; conftest change in `bcdd84264`; the 10 tests that crashed were the `@pytest.mark.gpu` benchmark/bindings/high_fidelity tests plus `test_synthetic_loader_batch_count`._

ryankert01:
Closes #1414.

## Why

#1321 lets `_qdp` build and import **without the CUDA toolkit** (stub CUDA Runtime symbols). That breaks the long-standing assumption that "the `_qdp` extension imports" ⟺ "a GPU is usable", which the test suite relied on to gate GPU tests. #1321 patched the immediate breakage by skipping `@pytest.mark.gpu` tests on `torch.cuda.is_available()`, but that is a **proxy**: on this feature's headline scenario — a GPU host with PyTorch but no toolkit — `_qdp` is a stub build, yet `torch.cuda.is_available()` is still `True`, so the GPU tests would run against the stub and abort.

This PR makes the engine report the truth.

## What

- **`qdp-core`**: add `cuda_runtime_available()` — queries `cudaGetDeviceCount` and returns `true` only when a device is actually present. In a `qdp_no_cuda` (stub) build it resolves to the existing 999-sentinel stub, so it returns `false` with no `cfg` branching or dead code. Re-exported from the crate root.
- **`_qdp`**: expose it as `_qdp.cuda_available()`.
- **`qumat_qdp`**: add `is_cuda_available()` (mirrors the existing `is_triton_amd_available()`) as the single Python source of truth.
- **`testing/conftest.py`**: gate the `@pytest.mark.gpu` auto-skip on the native signal, falling back to `torch` only if the helper is absent.
- **`test_fallback.py`**: coverage that also runs on a stub build, guarding that querying availability returns a `bool` without aborting.

## Test plan

Verified locally on 4× NVIDIA GPUs (CUDA 12.8), real build:

- `_qdp.cuda_available()` / `qumat_qdp.is_cuda_available()` → `True` with a GPU, `False` under `CUDA_VISIBLE_DEVICES=""`.
- `testing/qdp` + `testing/qdp_python` with CUDA hidden: **130 passed, 156 skipped, 0 failed** (no aborts).
- The 10 tests that crashed in #1321's CI run: **pass on GPU**, **skip when hidden**.
- `qdp-core` builds in both real and `QDP_NO_CUDA=1` stub modes; full Rust suite, `clippy --all-features -D warnings`, `ruff`, and `ty` all clean.

## Out of scope (left in #1414)

Making the stub path raise a clean `RuntimeError` instead of aborting for direct (non-test) callers, and collapsing the ~65 redundant inline `torch.cuda.is_available()` checks now that the conftest gate is accurate.

### DIFF
--- a/qdp/qdp-core/src/gpu/cuda_ffi.rs
+++ b/qdp/qdp-core/src/gpu/cuda_ffi.rs
@@ -69,6 +69,10 @@ unsafe extern "C" {
 
     pub(crate) fn cudaMemGetInfo(free: *mut usize, total: *mut usize) -> i32;
 
+    /// Number of CUDA-capable devices. Used to detect whether a GPU is
+    /// actually present at runtime (see `cuda_runtime_available`).
+    pub(crate) fn cudaGetDeviceCount(count: *mut i32) -> i32;
+
     pub(crate) fn cudaMemcpyAsync(
         dst: *mut c_void,
         src: *const c_void,
@@ -153,6 +157,10 @@ mod no_cuda_stubs {
         QDP_CUDA_UNAVAILABLE
     }
 
+    pub(crate) unsafe fn cudaGetDeviceCount(_count: *mut i32) -> i32 {
+        QDP_CUDA_UNAVAILABLE
+    }
+
     pub(crate) unsafe fn cudaMemcpyAsync(
         _dst: *mut c_void,
         _src: *const c_void,
@@ -216,3 +224,19 @@ mod no_cuda_stubs {
 
 #[cfg(qdp_no_cuda)]
 pub(crate) use no_cuda_stubs::*;
+
+/// Returns `true` if a usable CUDA device is present at runtime.
+///
+/// This is deliberately distinct from "the extension is importable": a
+/// no-toolkit (`qdp_no_cuda`) build links CUDA stubs that return a non-zero
+/// sentinel, and a host with the toolkit but no GPU reports zero devices. Both
+/// yield `false` here, so callers (and the Python test suite) can gate GPU work
+/// on an accurate signal rather than mere importability.
+pub fn cuda_runtime_available() -> bool {
+    let mut count: i32 = 0;
+    // SAFETY: `cudaGetDeviceCount` writes a single `i32` through `count` and
+    // accesses no other memory. In a `qdp_no_cuda` build this resolves to the
+    // stub above, which returns the unavailable sentinel without dereferencing.
+    let rc = unsafe { cudaGetDeviceCount(&mut count) };
+    rc == CUDA_SUCCESS && count > 0
+}
--- a/qdp/qdp-core/src/lib.rs
+++ b/qdp/qdp-core/src/lib.rs
@@ -36,6 +36,7 @@ pub mod types;
 mod profiling;
 
 pub use error::{MahoutError, Result, cuda_error_to_string};
+pub use gpu::cuda_ffi::cuda_runtime_available;
 pub use gpu::memory::Precision;
 pub use reader::{FloatElem, NullHandling, handle_float32_nulls, handle_float64_nulls};
 pub use types::{Dtype, Encoding};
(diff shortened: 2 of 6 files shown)


rich7420:
@ryankert01 thanks for the patch, Nice one 
locally tested
the no-cfg-branch stub resolve is clean, and gating on the native signal instead of torch is the right fix. Verified both crates build in stub mode.

Two tiny nits, both optional:

1. Three names for one concept (`cuda_runtime_available` → `cuda_available` → `is_cuda_available`). Each follows local convention so it's fine, just a little to track.(it's not that important)

2. `is_cuda_available()` does a fresh FFI `cudaGetDeviceCount` every call, and conftest calls it per gpu test. Cheap so not a perf issue, but the result can't change within a run — an `lru_cache` (or `OnceLock<bool>`) would make it a one-time query.

rich7420 (approved):

viiccwen (approved):
LGTM!

PR #1422, opened 2026-06-28
[Testing][QDP] f32 Parquet fidelity tests and benchmark

### LINKED ISSUE #1342: [Testing] Parquet f32 tests and benchmark
### What

Add fidelity tests and an optional benchmark script that verify correctness and measure throughput for the end-to-end f32 Parquet pipeline enabled by A1–A3.

### Why

Without dedicated tests, regressions in the f32 reader or dispatch path may go undetected. Without a benchmark, the expected ~25–35% throughput improvement over the f64 path (when the Parquet column is native f32) remains unvalidated.

### How

- Fidelity test: compare f64-kernel output vs f32-reader + f32-kernel output (threshold ≥ 0.99999 where applicable) in `qdp-core/tests/`
- Optional: `qdp-python/benchmark/benchmark_parquet_f32.py` reporting vec/s for f32 vs f64, with a usage comment or README note
- Tests run in CI Linux CUDA job (or CPU smoke if no GPU available in CI)

**Acceptance criteria:**
- [ ] Tests run in CI Linux CUDA job (or CPU smoke if no GPU in CI)
- [ ] Benchmark script has usage comment or README note

0lai0:
### Related Issues

Closes #1342 
Part of #1338 
### Changes

- [ ] Bug fix
- [ ] New feature
- [ ] Refactoring
- [ ] Documentation
- [x] Test
- [x] CI/CD pipeline
- [ ] Other

### Why

The f32 Parquet pipeline (A1–A3) lacked dedicated end-to-end fidelity coverage and throughput validation. Without tests, regressions in the f32 reader or dispatch path could go undetected, and without a benchmark, the expected 25–35% throughput improvement over the f64 path for native f32 columns remained unmeasured.

### How

- `qdp-core/tests/parquet_f32_fidelity.rs`: End-to-end f32-vs-f64 fidelity test for the Parquet pipeline.
  - CPU smoke(asserts in CI): f32 and f64 readers return the same logical values up to f64→f32 rounding.
  - GPU fidelity (Linux + CUDA, self-skip on no-CUDA stub via kernel probe): same deterministic data written as native `List<Float32>` / `List<Float64>`, read through matching readers, encoded with f32/f64 kernels, compared via `fidelity_cross_precision`. Thresholds match `gpu_fidelity.rs` (1e-3 at 8–16 qubits; amplitude f32 norm precision — the issue's "≥0.99999 where applicable" does not apply here).
- `qdp-core/tests/common/mod.rs`: Shared `List<Float32>` / `List<Float64>` Parquet writers for fidelity fixtures.
- `qdp-python/benchmark/benchmark_parquet_f32.py` + `README.md`: Optional local/GPU throughput benchmark (vec/s + f32/f64 speedup). Default run generates separate native f32 and f64 files so the f32 path exercises zero-copy native read; `--parquet PATH` benchmarks a single user file at both dtypes. Not wired into CI.
- `.github/workflows/python-testing.yml`: Run `parquet_f32` (existing #1340 reader acceptance) and `parquet_f32_fidelity` in the `rust-check` job under `QDP_NO_CUDA=1`. CPU smoke asserts; GPU cases skip without a GPU. Scoped to these two binaries — other `gpu_*.rs` tests lack probe-skip and would fail on the stub.

## Checklist

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -59,9 +59,10 @@ jobs:
         run: cargo check --workspace --tests
 
       # CPU smoke for the f32 Parquet pipeline (issue #1342). No GPU in CI, so
-      # the GPU fidelity cases self-skip (kernel probe fails on the stub) and the
-      # reader-level assertions still run. Scoped to these two binaries to avoid
-      # building/running the GPU-only test binaries here.
+      # GPU fidelity cases self-skip when the probe finds no functional GPU
+      # (stub build: kernel launch returns Err; no driver: cudarc panics and
+      # catch_unwind treats it as skip). Reader-level assertions still run.
+      # Scoped to these two binaries — other gpu_*.rs tests lack probe-skip.
       - name: Cargo test (f32 Parquet CPU smoke)
         working-directory: qdp
         env:
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -41,9 +41,10 @@ on:
 
 jobs:
   rust-check:
-    # Fast gate: type-check both with and without CUDA stubs so duplicate
-    # stub definitions or cfg mismatches fail in ~30s instead of surfacing
-    # during the slower maturin build below.
+    # Early gate: type-check both with and without CUDA stubs so duplicate
+    # stub definitions or cfg mismatches fail quickly instead of surfacing
+    # during the slower maturin build below. Also runs qdp-core f32 Parquet
+    # CPU smoke tests (issue #1342).
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6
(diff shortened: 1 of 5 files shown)


0lai0:
cc @rich7420

rich7420:
@0lai0 thanks for the patch!
LGTM overall 
pulled it down, tests pass and clippy's clean, GPU cases self-skip nicely. Threshold matching gpu_fidelity.rs (1e-3, not 0.99999) is the right call.

Two small nits, neither blocking:

1. `write_list_parquet_f32`/`write_list_parquet_f64` are near-identical (only the builder type differs). Not worth forcing a generic over Arrow builders, but figured I'd flag the dupe.

2. In the benchmark, `--parquet` vec/s divides metadata `num_rows` by the time to iterate `total_batches` batches. Fine for the generated case where they line up, but a bit off for an arbitrary user file where `batches*batch_size != num_rows`.

rich7420 (approved):

PR #1454, opened 2026-07-23
[QDP][feature]Add estimate_memory() for pipeline CPU/GPU footprint

### LINKED ISSUE #1429: [Feature] Add estimate_memory() for pipeline GPU footprint
### What

Add a pure `estimate_memory()` function that, given an encoding configuration, returns the predicted CPU prefetch-pool and GPU state-buffer memory the pipeline will allocate — so callers can know the footprint before the first batch runs. No allocation, no CUDA calls, no side effects.

### Why

There is no upfront memory check anywhere in the pipeline today. Users only discover they are over budget when CUDA throws OOM mid-run — sometimes after 90+ minutes of encoding (see #1262). Because the failure surfaces deep inside `encode_batch`, the error gives no hint about which knob (batch size, qubit count, prefetch depth) caused it. A cheap, deterministic estimate computed from config alone lets us fail fast (B2) and gives users an actionable number to tune against.

### How

- Add a pure function in `qdp-core` (e.g. `qdp-core/src/estimate.rs` — note `qdp-core/src/gpu/memory.rs` already exists, so avoid a bare `memory.rs` at the crate root to prevent confusion):

```rust
pub fn estimate_memory(
    encoding: Encoding,
    num_qubits: u32,
    batch_size: usize,
    dtype: Dtype,          // existing enum in qdp-core/src/types.rs (Float32/Float64)
    prefetch_depth: usize,
) -> MemoryEstimate
```

- Introduce a `MemoryEstimate` struct exposing at least `cpu_prefetch_bytes`, `gpu_state_bytes`, and a `total()` helper.

**Memory model.** These formulas are the source of truth for this issue:

```
cpu_prefetch_bytes = prefetch_depth × batch_size × sample_size × bytes_per_elem
gpu_state_bytes    = 2 × batch_size × state_len × complex_bytes
total()            = cpu_prefetch_bytes + gpu_state_bytes
```

where:

| term | definition |
|---|---|
| `bytes_per_elem` | from `Dtype::bytes()` — 4 for `Float32`, 8 for `Float64`. Do not hardcode. |
| `complex_bytes` | `2 × bytes_per_elem` (8 for f32 → complex64, 16 for f64 → complex128) |
| `sample_size` | input vector length; for amplitude, `2^num_qubits` |
| `state_len` | output state vector length; for amplitude, `2^num_qubits` |

- Derive `sample_size` / `state_len` from `encoding` + `num_qubits` so the function stays correct across encodings.

**Modeling notes — please preserve these semantics and document them in rustdoc:**

1. **The leading `2×` on `gpu_state_bytes` is a double-buffering allowance (state vector + recycle buffer), not the complex real/imaginary factor** — the real/imag pair is already inside `complex_bytes`. The current batch path (`GpuStateVector::new_batch`) allocates a single `batch_size × state_len × complex_bytes` buffer, so the `2×` is a deliberate conservative upper bound.
2. **`cpu_prefetch_bytes` covers exactly `prefetch_depth` batches**, matching the formula. The true instantaneous peak can reach `prefetch_depth + 2` batches (one being produced, one being consumed, alongside a full channel). Document this in rustdoc but **do not** fold the `+2` into the returned value — the returned number stays aligned with the formula above.
3. **`prefetch_depth` is a parameter here, not something this function computes.** The pipeline auto-computes it at runtime (`pipeline_runner.rs:153`; it yields 16 at 16 qubits and 1 at 20 qubits). `estimate_memory` must not replicate that logic — callers pass the resolved value.
4. **Known exclusion:** the host-side streaming chunk buffer (Parquet reader) is *not* included in this estimate. Document this explicitly in rustdoc so B2 does not silently under-count for streaming sources. Folding it in requires reader config beyond this signature — tracked as a follow-up.

**Out of scope:** VRAM querying and fail-fast wiring (B2), Python bindings, the streaming chunk buffer term (note 4), non-amplitude encodings beyond the size formula already needed here.

### Acceptance criteria

- [ ] `estimate_memory` is pure: no allocation, no CUDA dependency
- [ ] Unit tests for the two reference cases below assert the **exact byte values**:

| case | encoding | qubits | dtype | batch_size | prefetch_depth | `cpu_prefetch_bytes` | `gpu_state_bytes` | `total()` |
|---|---|---|---|---|---|---|---|---|
| A | amplitude | 16 | f32 | 64 | 16 | 268,435,456 (256 MiB) | 67,108,864 (64 MiB) | 335,544,320 (320 MiB) |
| B | amplitude | 20 | f32 | 64 | 1 | 268,435,456 (256 MiB) | 1,073,741,824 (1 GiB) | 1,342,177,280 (1.25 GiB) |

  The `prefetch_depth` values match what the pipeline auto-computes at those qubit counts, but the tests should pass them explicitly rather than depending on the auto-compute path.

- [ ] Unit test covering the f64 vs f32 `bytes_per_elem` difference via `Dtype::bytes()` — case A with `dtype = Float64` must double both terms: `cpu = 536,870,912` (512 MiB), `gpu = 134,217,728` (128 MiB), `total = 671,088,640` (640 MiB)
- [ ] `MemoryEstimate` documented in rustdoc with the formula and a worked example
- [ ] Rustdoc states the three documented caveats: the `2×` is double-buffering (not the complex factor), the `+2` instantaneous peak is excluded, and the streaming chunk buffer is excluded
- [ ] `cargo test -p qdp-core` passes on Linux

0lai0:
### Related Issues

Closes #1429

### Changes

- [ ] Bug fix
- [x] New feature
- [ ] Refactoring
- [ ] Documentation
- [ ] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

Pipelines only hit memory overflow mid-run, after extended processing. This adds a config-only estimator so callers can check the budget and fail fast before the first batch.

### How

Adds `qdp_core::estimate_memory()` — a pure, allocation-free function (no CUDA, no device queries) predicting a config's CPU prefetch-pool and GPU state-buffer footprint:
- `cpu_prefetch_bytes = prefetch_depth * batch_size * sample_len * bpe`
- `gpu_state_bytes    = 2 * batch_size * state_len * complex_bytes`

Notes:
- Returns `Result` — all arithmetic is checked; overflow / oversized `num_qubits`
  → `InvalidInput`, never panics.
- f32 → f64 fallback mirrors `PipelineConfig::normalize`.
- Streaming Parquet chunk buffer excluded (follow-up / B2).

Files: `estimate.rs` (new), `lib.rs` (register + re-export), `tests/estimate.rs` (new).

## Checklist

### DIFF
--- a/qdp/qdp-core/src/estimate.rs
+++ b/qdp/qdp-core/src/estimate.rs
@@ -1,0 +1,182 @@
+//
+// Licensed to the Apache Software Foundation (ASF) under one or more
+// contributor license agreements.  See the NOTICE file distributed with
+// this work for additional information regarding copyright ownership.
+// The ASF licenses this file to You under the Apache License, Version 2.0
+// (the "License"); you may not use this file except in compliance with
+// the License.  You may obtain a copy of the License at
+//
+//    http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Pure, allocation-free memory estimation for pipeline configurations.
+//!
+//! [`estimate_memory`] answers "how much memory will this config need?" from the config alone —
+//! no allocation, no CUDA calls, no device queries — so callers can fail fast before the first
+//! batch instead of discovering an overflow mid-run.
+
+use crate::error::{MahoutError, Result};
+use crate::types::{Dtype, Encoding};
+
+/// Predicted memory footprint of a pipeline configuration, in bytes.
+///
+/// Produced by [`estimate_memory`]. Both fields are upper-bound estimates derived from config
+/// arithmetic only; nothing here reflects an actual allocation. See [`estimate_memory`] for the
+/// formulas behind each field and a worked example.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub struct MemoryEstimate {
+    /// Host-side prefetch pool: `prefetch_depth` batches of raw (unencoded) input.
+    pub cpu_prefetch_bytes: u64,
+    /// Device-side state-vector buffer, including a double-buffering allowance.
+    pub gpu_state_bytes: u64,
+}
+
+impl MemoryEstimate {
+    /// Combined host + device footprint in bytes.
+    ///
+    /// Saturates instead of overflowing: values produced by [`estimate_memory`] are checked to sum
+    /// without overflow, but the fields are public and may be set directly.
+    #[must_use]
+    pub const fn total(&self) -> u64 {
+        self.cpu_prefetch_bytes.saturating_add(self.gpu_state_bytes)
+    }
+}
+
+/// Estimate the CPU prefetch-pool and GPU state-buffer footprint of a pipeline configuration.
+///
+/// Pure function: performs no allocation, touches no device, and never panics — every product is
+/// computed with checked arithmetic and an overflow becomes [`MahoutError::InvalidInput`] naming
+/// the parameter responsible.
+///
+/// # Formulas
+///
+/// Let `n = num_qubits`, `bpe = dtype.bytes()` (4 for f32, 8 for f64), and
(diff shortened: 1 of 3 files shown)


0lai0:
cc @rich7420

rich7420:
LGTM.

rich7420 (approved):

PR #1462, opened 2026-08-05
[QDP] StreamingProducer: use VecDeque for O(1) buffer advance

### LINKED ISSUE #1436: [Refactor] StreamingProducer: use VecDeque for O(1) buffer advance
### What

Change the streaming producer's internal buffer from a `Vec` advanced with `Vec::drain(..cursor)` to a `VecDeque`, so advancing past consumed samples is O(1) instead of O(n).

### Why

`StreamingProducer` currently advances its buffer with `self.buffer.drain(..self.buffer_cursor)` (`qdp-core/src/pipeline_runner.rs:368`), which memmoves the remaining elements down on every batch — O(n) per advance, paid on the streaming hot path. A `VecDeque` makes front-popping O(1) and avoids the repeated memmove, with no change to the data produced.

### How

- Replace the backing `Vec` buffer in the streaming producer (or `StreamingAdapter`, if E2 has landed) with a `VecDeque<T>`.
- Swap the `Vec::drain(..cursor)` front-removal for `VecDeque` front operations (`drain(..cursor)` / `pop_front`) so the advance is O(1) amortized.
- Ensure capacity stays bounded across many batches (no unbounded growth or reallocation churn).
- Keep output identical to the current streaming path.

**Out of scope:** Producer unification (E2) if taken separately, any change to streaming correctness or output.

**Acceptance criteria:**

- [ ] Test: `VecDeque` capacity stays constant over 100 consecutive batches (no reallocation growth)
- [ ] No correctness change on the streaming coverage in `tests/parquet_f32.rs` and `tests/reader.rs`
- [ ] Streaming hot-path advance is O(1) (no per-batch O(n) memmove)

0lai0:
### Related Issues

Closes #1436

### Changes

- [ ] Bug fix
- [ ] New feature
- [x] Refactoring
- [ ] Documentation
- [x] Test
- [ ] CI/CD pipeline
- [ ] Other

### Why

`StreamingProducer` tracked consumed elements with a `buffer_cursor` and, once the cursor passed the halfway mark (`BUFFER_COMPACT_DENOM`), reclaimed the prefix with `Vec::drain(..cursor)` — an O(n) memmove of the retained tail, on the streaming hot path.

A `VecDeque` advances its head instead, so discarding a consumed prefix never shifts the data that is still live: O(1) amortized buffer advance, no periodic compaction pass, and no change in output.

### How

- `buffer: Vec<T>` + `buffer_cursor: usize` → `buffer: VecDeque<T>`; removed the now-unused `BUFFER_COMPACT_DENOM` compaction heuristic
- `produce()` copies the batch out of `as_slices()` (stitching both halves when a batch straddles the ring's wrap boundary), then `drain(..take)`s the consumed prefix. Copying via the slices keeps the batch copy on `extend_from_slice`'s bulk path — `Drain` is not `TrustedLen`, so `extend(drain)` would copy element by element. Recycled batch buffers are still reused, so the batch copy itself stays allocation-free
- Refills use `extend(&scratch[..written])` rather than `.iter().copied()`: `VecDeque` specializes `Extend<&T> for T: Copy` into a bulk `copy_slice`
- The build path pre-reserves the steady state (`batch_size * sample_size + initial_cap`, the peak live length) so the ring never reallocates mid-run. `VecDeque::from(Vec)` reuses the existing allocation, so this only tops the capacity up.
## Checklist


viiccwen (commented):
Thx for the correctness and capacity tests. Since the main change is hot-path performance, could you also provide a small reproducible before/after benchmark?

It would be helpful to compare the current `Vec` + cursor implementation against this `VecDeque` implementation.

rich7420:
LGTM.

0lai0:
Thanks @viiccwen and @rich7420 
Added beanch mark `make -C qdp bench_streaming_buffer`
(before = frozen pre-change producer; after = shipped `StreamingProducer`)

Buffer-only (refill chunk 65536), Xeon w3-2435, `--release`:

| batch:chunk | before | after | speedup |
|---|---|---|---|
| 1:16 | 782 ns | 504 ns | 1.55x |
| 1:4 | 3370 ns | 2480 ns | 1.36x |
| 1:1 | 9668 ns | 9716 ns | 1.00x |
| 4:1 | 54176 ns | 54469 ns | 0.99x |

E2E via Parquet: ~1.01–1.02x (decode dominates). Win only when batch ≪ chunk; at 1:1+ both do the same work. 
Happy to reword the PR around #1436's O(1) advance / constant-capacity criteria rather than throughput

qdp/Makefile
viiccwen, line 1: we don't need to expose any benchmark experiment in command.

qdp/docs/benchmarks/streaming-buffer-vecdeque.md
viiccwen, line 1: we don't need this.

viiccwen (changes requested):
left comments.

qdp/docs/benchmarks/streaming-buffer-vecdeque.md
0lai0: Sure, I’ll take care of that.

qdp/Makefile
0lai0: Sure, PR updated
Thanks for review

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -1796,6 +1812,117 @@ mod tests {
...
+            const BATCH_SIZE: usize = 4;
+            const SAMPLE_LEN: usize = 4; // 2 qubits, amplitude encoding
+            const BATCHES: usize = 100;
+            // Elements consumed per batch.
+            const STRIDE: usize = BATCH_SIZE * SAMPLE_LEN;
+            // One sample more than a batch: produce() refills on nearly every batch and always
+            // leaves a remainder behind, so the head walks around the ring instead of resetting.
+            const READ_CHUNK: usize = STRIDE + SAMPLE_LEN;
+            // Each sample written by write_f32_parquet_n, repeated across the batch.
+            const SAMPLE: [f32; SAMPLE_LEN] = [0.25, 0.5, 0.75, 1.0];
```
viiccwen, line 1841: Every sample in this fixture has the same values, the test cannot detect samples being reordered, duplicated, or skipped across the wrap boundary. Reversing the `front` and `back` slices still passes this test.

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -1796,6 +1812,117 @@ mod tests {
...
+                let mut producer = StreamingProducer::<f32> {
+                    reader,
+                    buffer: VecDeque::with_capacity(STRIDE + READ_CHUNK),
+                    read_chunk_scratch: vec![0.0_f32; READ_CHUNK],
+                    sample_size: SAMPLE_LEN,
+                    batch_size: BATCH_SIZE,
+                    num_qubits: 2,
+                    batches_yielded: 0,
+                    batch_limit: usize::MAX,
+                };
```
viiccwen, line 1864: IMO, This test constructs the deque with the expected steady-state capacity directly, so it does not verify the `reserve` logic in `build_streaming_producer`, right? The production reservation could be removed or broken while this test still passes.

Could the test create the producer through the prod initialization path and then verify that capacity remains stable for 100 batches?

qdp/qdp-core/src/pipeline_runner.rs
```diff
@@ -631,10 +636,17 @@ where
...
     buffer.truncate(written);
     let read_chunk_scratch = vec![T::default(); initial_cap];
+    // Pre-reserve the streaming steady state so the ring never reallocates mid-run. Each batch
+    // drains up to `required` elements from the front while a refill appends up to one
+    // `initial_cap` chunk to the back, so live length peaks at `required + initial_cap`.
+    // `VecDeque::from` reuses the Vec's existing `initial_cap` capacity (O(1), no realloc), so
+    // this only tops it up by `required` — making front-advance realloc-free from batch 0.
+    let required = config.batch_size * sample_size;
+    let mut buffer = VecDeque::from(buffer);
+    buffer.reserve((required + initial_cap).saturating_sub(buffer.len()));
```
viiccwen, line 646: This eagerly reserves space for a full batch plus one refill chunk, even when the file only contains a partial batch. For large samples, this can substantially increase memory usage; for example, 20-qubit amplitude data with `batch_size=64` and `f64` grows from roughly 8 MiB to 520 MiB.

Could we avoid reserving the theoretical peak before EOF is known, and use checked arithmetic for the capacity calculation?

viiccwen (changes requested):
left comments.

PR #1466, opened 2026-08-13
[QDP][feature] reject pipeline configs that exceed free VRAM

### LINKED ISSUE #1430: [Feature] Fail fast when estimated memory exceeds VRAM
### What

Call `estimate_memory()` (B1) at loader/iterator construction, compare it against free GPU memory, and reject configurations that won't fit — returning a clear, actionable error in under a second instead of OOM-ing after a long run.

### Why

This is the user-facing half of the OOM problem in #1262. Today an oversized config is accepted silently and only fails once the GPU state buffer is actually allocated, deep into encoding. Validating at build time turns a confusing late crash into an immediate, descriptive error that tells the user exactly which parameters to reduce.

### How

- In `PipelineIterator::new_synthetic` / `new_from_file` / `new_from_file_streaming` (`qdp-core/src/pipeline_runner.rs`) and the Python loader build path (`qdp-python/src/loader.rs`), call `estimate_memory()` with the resolved `PipelineConfig`.
- Query free GPU memory via the existing CUDA helpers (the same layer `encode_batch` uses).
- If `estimate.gpu_state_bytes` exceeds free VRAM (optionally minus a configurable safety margin), return a structured error whose message names the offending values and suggests remedies: `reduce batch_size, num_qubits, or prefetch_depth`.
- Degrade gracefully when no GPU is present — the CPU-only / smoke path must not hard-fail on the VRAM check.
- Keep the check cheap enough to run unconditionally at construction (<1 s).
- Optional (nice-to-have): expose `estimate_memory` to Python as a preview hook for `plan_encoding`.

**Out of scope:** Changing the memory model itself (owned by B1), automatic config downsizing, multi-GPU accounting.

0lai0:
### Related Issues
Closes #1430
### Changes
- [ ] Bug fix
- [x] New feature
- [x] Refactoring
- [x] Documentation
- [x] Test
- [x] CI/CD pipeline
- [ ] Other

### Why

An oversized QDP pipeline currently runs until `encode_batch` allocates GPU buffers before it fails — slow, and the error does not say what to change. This adds a fail-fast check against free VRAM so a configuration that cannot fit is rejected in milliseconds, before any input file is read, with a message naming the values to lower.

### How

- After `normalize()`, the three `PipelineIterator` constructors  (`new_synthetic`, `new_from_file`, `new_from_file_streaming`) compare `estimate_memory().gpu_state_bytes` to free device memory. The check is skipped when the CUDA runtime reports no usable device — stub / no-toolkit builds, and also an empty `CUDA_VISIBLE_DEVICES`.

- **Refactor (needed to test the rule on CI, which has no GPU):**
`ensure_device_memory_available` now queries `cudaMemGetInfo` and delegates the comparison to `ensure_fits_in_free_memory`. Tests inject `(free, total)` through `ensure_config_fits_device_with`. Production path, error type, and message are unchanged; both helpers are `pub(crate)` / private.

- The budget is **two** concurrent batch state buffers. That is deliberate: a `for qt in loader:` loop keeps the previous DLPack tensor alive while the next batch is allocated. A config that fits one buffer but not two is rejected. There is no override.
**This budget is a floor, not the true peak:** the encoders also upload the input batch to the device (`htod_sync_copy`) and hold it alongside the state buffers — for amplitude, half a buffer, so the real steady-state peak is nearer 2.5. A configuration sitting just under free memory can therefore still OOM mid-run; this narrows that window rather than closing it. Modelling it properly means changing the memory model, which #1430 puts out of scope.

- Precision is the wider of the loader dtype and the engine precision. File loaders budget basis as float64 (indices are read as f64); synthetic basis is budgeted at the requested dtype.

- The comparison is against free memory *sampled* at that moment — nothing is reserved. Another process or a second loader can take it before the first batch allocates, so passing the check is a fast sanity check, not a guarantee.

- On the Python surface the check runs when **iteration starts** (`for qt in loader:`), not when `QuantumDataLoader(...)` returns. The rejection is a `RuntimeError` wrapping the Rust `MemoryAllocation` message (the existing conversion for every backend error). A dedicated exception type is left as follow-up.

- `qumat_qdp.estimate_memory()` exposes the same arithmetic without opening a device, so a caller can size `qubits` / `batch_size` before iterating.

- CI covers the accept/reject rule with injected free-memory figures (no GPU required). Integration tests still exercise the live-GPU path and self-skip on stub builds.

Out of scope, per #1430: changing the underlying memory model (see the 2-buffer note above), automatic config downsizing, and multi-GPU accounting — the guard queries the current CUDA device and does not call `cudaSetDevice`. Also not covered: host-side streaming chunk limits, and the benchmark helpers (`run_throughput_pipeline` / `run_latency_pipeline`), which build their own engine and bypass `PipelineIterator`.

### Verification

Run on a host with CUDA 12.5 and a live GPU:

- `make pre-commit` — clean.
- `make test_rust` — 379 passed, 0 failed, 0 ignored.
- `make test_python` — 888 passed, 1 skipped (pre-existing, unrelated), 0 failed.
- `QDP_NO_CUDA=1 cargo check --workspace --tests` — clean; both new CI steps pass on the stub build (15 and 47 tests).
- `cd website && npm run typecheck && npm run build` — clean; the two broken-link warnings are pre-existing (verified by building `upstream/main`).

## Checklist

### DIFF
--- a/.github/workflows/python-testing.yml
+++ b/.github/workflows/python-testing.yml
@@ -62,12 +62,25 @@ jobs:
       # GPU fidelity cases self-skip when the probe finds no functional GPU
       # (stub build: kernel launch returns Err; no driver: cudarc panics and
       # catch_unwind treats it as skip). Reader-level assertions still run.
-      # Scoped to these two binaries — other gpu_*.rs tests lack probe-skip.
+      # pipeline_vram_guard (issue #1430) uses the same catch_unwind probe via
+      # common::qdp_engine_probed, so it skips rather than fails here.
+      # Scoped to these binaries — other gpu_*.rs tests lack probe-skip.
       - name: Cargo test (f32 Parquet CPU smoke)
         working-directory: qdp
         env:
           QDP_NO_CUDA: "1"
-        run: cargo test -p qdp-core --test parquet_f32 --test parquet_f32_fidelity
+        run: cargo test -p qdp-core --test parquet_f32 --test parquet_f32_fidelity --test pipeline_vram_guard
+
+      # The pipeline_vram_guard cases above all self-skip without a device, so the
+      # memory-guard logic (issue #1430) is covered here instead: these unit tests
+      # are pure and need no GPU. Scoped to the module rather than a bare `--lib`
+      # because other qdp-core unit tests (e.g. encoding::amplitude's stream
+      # end-to-end case) do require a device and fail on the stub build.
+      - name: Cargo test (pipeline_runner CPU unit tests)
+        working-directory: qdp
+        env:
+          QDP_NO_CUDA: "1"
+        run: cargo test -p qdp-core --lib pipeline_runner::tests
 
   test:
     needs: rust-check
--- a/docs/qdp/python-api.md
+++ b/docs/qdp/python-api.md
@@ -21,6 +21,7 @@ from qumat_qdp import (
     BACKEND,
     Backend,
     LatencyResult,
+    MemoryEstimate,
     NativeQuantumTensor,
     QdpBenchmark,
     QdpEngine,
@@ -30,6 +31,7 @@ from qumat_qdp import (
     RustQdpEngine,
     ThroughputResult,
     TritonAmdEngine,
+    estimate_memory,
     force_backend,
     is_triton_amd_available,
     run_throughput_pipeline_py,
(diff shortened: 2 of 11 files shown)


ryankert01 (commented):
There's generally two problems need to think about when doing this:
1. Whether the guard rejects configs that actually runs
2. Whether it still OOM anyways. (guard too loose for some)

Because I have think of this before, I personally want to know the design choice on these two problems.

ps. this looks generally ok, while we need to be careful of how the guard works and avoid introduce weird mem bound bugs which makes you think is OOM.

Issue #1477, opened 2026-09-09
[Feature][QDP] One encode path across Python, Rust and CUDA (tracking)


ryankert01:
### TL;DR

Adding one encoding to QDP today means writing it three times: a CUDA host launcher plus `extern "C"` declaration plus no-CUDA stub, an 8-method Rust encoder (f32/f64 × single/batch × host/device), and a hand-matched PyO3 branch. #1268 added *one* f32 variant of angle encoding and touched 9 files across 2 languages for 1,135 lines.

I refactored this so an encoding is one device-only `.cu` file, one ~80-line Rust descriptor, and one PyTorch reference function. The data-movement engine (dual-stream pinned upload, prefetch, buffer pool, DLPack) is untouched, the public Python API is unchanged, and the repo's standard benchmarks show no measurable throughput or latency change.

### What

Six commits on a branch, each independently green (Rust tests, clippy, no-CUDA build, Python suites including a new parity grid). I'll open them as PRs against this issue:

- [ ] **Parity grid + baseline tool** — one test over encoding × precision × shape × input location against `torch_ref.py` (which now covers phase); `benchmark/baseline.py` captures and compares throughput/latency with a threshold.
- [ ] **Runtime kernel loading** — `build.rs` compiles each `.cu` to a device-only fatbin and embeds it; `qdp_kernels::registry` loads through the CUDA driver API on first use and exposes one generic `launch` + `kernel_args!`. A test proves every embedded symbol resolves.
- [ ] **Encoder collapse** — a `Kernel` trait with a single `launch` on device-resident data, `Input` (host slice or device pointer on a stream), and `QdpEngine::encode(input, shape, num_qubits, encoding)` replacing 24 named methods (old names kept as one-line shorthands in `compat.rs`). Bindings become `PyInput::from_py` + one call.
- [ ] **Host-path fixes** found while benchmarking (amplitude CPU finite pass, basis device cast).
- [ ] **Remove launchers** — 26 host launchers, 26 declarations, 26 stubs, the static nvcc compile and the `cc` dependency are gone.
- [ ] **Docs** — DEVELOPMENT.md architecture section, a "which files do I touch" table, the three-file recipe, `make check-kernels`, `make parity`.

| Surface | Before | After |
|---|---|---|
| CUDA host launchers / `extern "C"` decls / no-CUDA stubs | 26 / 26 / 26 | 0 |
| `qdp-kernels/src/lib.rs` | 726 lines | 57 lines |
[...]

PR #1478, opened 2026-09-09
[QDP] Add cross-backend parity grid and benchmark baseline tool (1/5)

ryankert01:
### Related Issues
Part of #1477 (1/5).

### Changes
- [x] Test
- [x] Other: benchmark tooling

### Why
Before touching the encode path, the branch needs a specification and a performance gate that exist independently of the Rust and CUDA layers. Today the only cross-backend check is the PyTorch reference tests, and phase has no reference at all.

### How
- `qumat_qdp.torch_ref.phase_encode`: the pure-PyTorch reference now covers every encoding the engine supports.
- `testing/qdp/test_parity.py`: one parametrised test over encoding × engine precision × shape × input location (NumPy, torch CPU, CUDA f64, CUDA f32), each cell checked against `torch_ref`. Known gaps (encodings without an f32 device kernel) are skipped explicitly.
- `qdp/qdp-python/benchmark/baseline.py`: capture throughput and latency for every encoding and dtype into JSON, and compare a later build against it with a regression threshold. Numbers are machine-specific and are not committed.

No engine code changes. `make parity` (added in 5/5) runs the grid.

## Checklist

---


ryankert01:

ryankert01:
Replaced by #1483 (GitHub Stack #1488).

PR #1479, opened 2026-09-09
[QDP] Load CUDA kernels at runtime from embedded fatbins (2/5)

ryankert01:
### Related Issues
Part of #1477 (2/5). Builds on #1478: this PR's diff against `main` includes that PR's commits; review the last commit only, or diff against `qdp-one-path/1-parity-grid`.

### Changes
- [x] Refactoring
- [x] CI/CD pipeline: `qdp-kernels/build.rs` changes how kernels are built

### Why
Every CUDA kernel today needs a hand-written host launcher, a matching `extern "C"` declaration in Rust, and a stub for builds without CUDA. The declaration is matched by eye; a mismatch is undefined behaviour, not a compile error. That is the first of the three places every encoder is expanded by hand.

### How
- `build.rs` compiles each `.cu` to a device-only fatbin (cubins for each supported arch plus PTX) and embeds it, plus the list of `extern "C" __global__` symbols and Rust constants mirrored from `kernel_config.h`. A kernel must be declared as `extern "C" __global__ void NAME(` on one line (optionally with `__launch_bounds__`); anything else fails the build with an explanation.
- `qdp_kernels::registry` loads a module through the CUDA driver API on first use, caches module and function handles **per primary context**, and exposes one generic `launch` plus a `kernel_args!` macro. Handles are reloaded transparently if the driver reports them stale, which happens when the last `CudaDevice` for a GPU is dropped and a new one created. Unknown symbols are reported by the driver.
- Kernels are marked `extern "C"` so symbols are unmangled. The hand-written launchers, declarations and stubs **stay in this PR** so `qdp-core` keeps building; 4/5 removes them.
- `tests/registry.rs` proves every embedded symbol resolves and that the generic launcher passes arguments with the correct ABI.

Risk: the build-script change. CI without `nvcc` embeds no kernels and every lookup returns `KernelError::Unavailable`; the `QDP_NO_CUDA=1` build is exercised.

Launch overhead is unchanged: `<<<>>>` compiles down to the same `cuLaunchKernel`. Verified by nsys in the issue.

## Checklist

---


ryankert01:

ryankert01:
Replaced by #1484 (GitHub Stack #1488).

PR #1483, opened 2026-09-09
[QDP] Add cross-backend parity grid and benchmark baseline tool

ryankert01:
### Related Issues
Part of #1477 (1/5).

### Changes
- [x] Test
- [x] Other: benchmark tooling

### Why
Before touching the encode path, the branch needs a specification and a performance gate that exist independently of the Rust and CUDA layers. Today the only cross-backend check is the PyTorch reference tests, and phase has no reference at all.

### How
- `qumat_qdp.torch_ref.phase_encode`: the pure-PyTorch reference now covers every encoding the engine supports.
- `testing/qdp/test_parity.py`: one parametrised test over encoding × engine precision × shape × input location (NumPy, torch CPU, CUDA f64, CUDA f32), each cell checked against `torch_ref`. Known gaps (encodings without an f32 device kernel) are skipped explicitly.
- `qdp/qdp-python/benchmark/baseline.py`: capture throughput and latency for every encoding and dtype into JSON, and compare a later build against it with a regression threshold. Numbers are machine-specific and are not committed.

No engine code changes. `make parity` (added in 5/5) runs the grid.

## Checklist

---
GitHub Stack #1488, series #1483 → #1484 → #1485 → #1486 → #1487. Merge in order; GitHub retargets the next PR's base when one merges.


qdp/qdp-python/benchmark/baseline.py
```diff
@@ -0,0 +1,178 @@
...
+        )
+        return out.stdout.strip()
+    except Exception:  # informational only
+        return "unknown"
+
+
+def measure(args: argparse.Namespace) -> dict[str, Any]:
+    cells: dict[str, dict[str, float]] = {}
+    for encoding in ENCODINGS:
+        for dtype in DTYPES:
```
viiccwen: For IQP, IQP-Z, and phase, an `f32` request is normalized to `f64` by the Rust pipeline because these encodings do not support the f32 path. This still records those measurements as `iqp/f32`, `iqp-z/f32`, and `phase/f32`, so the baseline reports f32 coverage while measuring f64. Could unsupported cells be skipped, or could the effective dtype be recorded instead? Otherwise, a future real f32 implementation would be compared against an f64 baseline.

qdp/qdp-python/qumat_qdp/torch_ref.py
```diff
@@ -30,7 +30,14 @@
...
 import torch.nn.functional as F
 
-__all__ = ["amplitude_encode", "angle_encode", "basis_encode", "encode", "iqp_encode"]
+__all__ = [
+    "amplitude_encode",
+    "angle_encode",
+    "basis_encode",
+    "encode",
+    "iqp_encode",
+    "phase_encode",
```
viiccwen, line 39: Adding phase to this dispatcher with Rust backend will be fine. However, pytorch backend still treats phase like amplitude, and make it error.

viiccwen (commented):
`phase` falls to the amplitude default branch in pytorch backend, maybe use todo / issue ticket to record it.

https://github.com/apache/mahout/blob/c176559841e99ec547142ae3f43238d8d0956129/qdp/qdp-python/qumat_qdp/api.py#L290-L297

https://github.com/apache/mahout/blob/c176559841e99ec547142ae3f43238d8d0956129/qdp/qdp-python/qumat_qdp/api.py#L299-L316

qdp/qdp-python/benchmark/baseline.py
ryankert01: Good catch, thanks. Fixed in 1dcf7d9: `Encoding::supports_f32` is now exported as `_qdp.encoding_supports_f32`, and `baseline.py` skips `f32` cells for encodings without a native f32 path (currently iqp, iqp-z, phase). Skipped cells are printed and listed under `"skipped"` in the JSON so the omission is visible, and `compare` ignores them. A test checks the exported flag against the parity grid's f32 list, so a real f32 kernel landing later will flip both together.

qdp/qdp-python/qumat_qdp/torch_ref.py
ryankert01: Fixed in 1dcf7d9 rather than deferred: the PyTorch backend now treats `phase` like `angle`/`iqp-z` (one angle per qubit), so it no longer falls into the amplitude branch. `test_pytorch_throughput_every_encoding` runs the PyTorch backend for all six encodings.

